PR #8, opened 2022-04-09
Start implementing MemHost

graydon:


graydon:
This gets far enough to be able to run the test_host_objects invoke function from a unit test using the mock MemHost.

leighmcculloch (approved):

PR #10, opened 2022-04-12
Make as_u63 public

leighmcculloch:
Make `is_u63` and `as_u63` public.

Also, fixed a bug in `TryFrom<i64> for Val`.

Also, I noticed the logic for checking the validity of a u63 existed in two places, in `TryFrom<i64> for Val` and in `from_u63`, so I made it so it only lives in the TryFrom. I chose the TryFrom because it needs to do the check or error, and so it is simple for from_u63 to wrap that and abort.

We should probably normalize on the From/TryFrom doing the work and the as_ and from_ functions simply wrapping them for the same reason and consistency.


graydon (approved):
LGTM

PR #12, opened 2022-04-12
Organize all test components into testing mod

leighmcculloch:
### What
Move the mock host and its implementations, and the swap function, into a testing module. Change the visibility of those pieces to public.

### Why
The visibility needs changing to be usable from tests. Most of the host module is intended to be private, so making one part of it public is a little odd. It is easier to see what is test functionality if it is separated into another module.


stellar_contract_sdk/src/testing/host/mod.rs
```diff
@@ -8,19 +8,9 @@
 
 pub mod fs;
 pub mod mem;
-use core::fmt::Debug;
 
 use crate::{Object, Val};
 
-impl Debug for Val {
```
leighmcculloch, line 15: For some reason I couldn't leave this here, it kept complaining about a conflict. I saw the status one lives alongside the type, so I figured that was okay and did the same here.

stellar_contract_sdk/src/testing/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(not(target_family = "wasm"))]
```
graydon, line 1: These 3 cfg attributes aren't necessary -- the whole testing module is not attached to the crate in that config already.

graydon: Also this mod declaration is redundant / likely going to break stuff: you're mounting the same file `src/testing/host/mod.rs` at two different paths in the crate: under `crate::host` and `crate::testing::host`. This will produce duplicate definitions of everything inside of it, as you observed with the definition of `impl Debug for Val`.

I'll push a reorg to this branch that I think will get you what you're trying to do here.

PR #36, opened 2022-04-18
Make invalid symbols error

leighmcculloch:
### What
Make symbols that are invalid – longer than 10 chars – error during compiling or at runtime.

### Why
It's too easy to make symbols that collide if you define two that have the same first 10 characters.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -41,8 +41,10 @@ pub use val::Val;
 pub use vec::Vec;
 
 #[inline(always)]
-pub fn require(b: bool) {
-    b.or_abort();
+pub const fn require(b: bool) {
+    if !b {
+        panic!();
+    }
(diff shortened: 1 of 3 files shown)


sdk/src/lib.rs
```diff
@@ -41,8 +41,10 @@ pub use val::Val;
 pub use vec::Vec;
 
 #[inline(always)]
-pub fn require(b: bool) {
-    b.or_abort();
+pub const fn require(b: bool) {
+    if !b {
+        panic!();
```
leighmcculloch, line 46: It seems like there's no value in calling `trap` directly via `or_abort`. Calling it directly can't be done from const functions because the wasm unreachable function is not const. Calling panic seems to do the same thing.

graydon (approved):

PR #37, opened 2022-04-18
Improve debug output of Val, Object, Symbol

leighmcculloch:
### What
Implement `Debug` for more cases of `Val`, and for `Object`, `Symbol`, and `Status`.

### Why
To make `log_value` output more useful info for some types.


tests/log/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+mod test {
+    use super::log;
+    use sdk::{Symbol, Val};
+    use stellar_contract_sdk as sdk;
+
+    #[test]
+    fn test_log() {
+        log(Val::from_i64(7));
+        log(Val::from_i64(-7));
+        log(Symbol::from_str("asdf").into());
```
leighmcculloch, line 21: This test will output the following with this PR and the fix from #34:
```
MemHost::log_value: Val { u63: 7 }
MemHost::log_value: Val { object: Object { type: 4, idx: 0, 0: 0 } }
MemHost::log_value: Val { symbol: Symbol { str: "asdf" } }
```
The second value is the `I64` object, which still doesn't render wonderfully, but this is all a step in the right direction.

graydon (approved):

PR #50, opened 2022-04-20
makefile: add fmt

leighmcculloch:


Makefile
```diff
@@ -8,6 +8,8 @@ all:
 		ls -l "$$i"; \
 	done
 
+fmt:
+	rustfmt --check $$(find . -type f -name '*.rs' -print)
```
leighmcculloch, line 12: I realize this won't expand filenames with whitespace in the names, but the chance of us naming .rs files like that is low, and rustfmt is much faster if you give it a lot of files at once, than it is if you execute it for each file individually.

Makefile
```diff
@@ -8,6 +8,8 @@ all:
 		ls -l "$$i"; \
 	done
 
+fmt:
+	rustfmt $$(find . -type f -name '*.rs' -print)
```
leighmcculloch, line 12: I realize this won't expand filenames with whitespace in the names, but the chance of us naming .rs files like that is low, and rustfmt is much faster if you give it a lot of files at once, than it is if you execute it for each file individually.

PR #68, opened 2022-04-30
Update strkey

leighmcculloch:
### What

Update the strkey package.

### Why

@tomerweller did some work on it.

### Known limitations

N/A

### DIFF
--- a/examples/liqpool/src/lib.rs
+++ b/examples/liqpool/src/lib.rs
@@ -328,7 +328,7 @@ mod test {
     use stellar_contract_sdk::testing::mem::{MemHost, MemLedgerKey, MemLedgerVal, MemObj};
     use stellar_contract_sdk::testing::{swap_mock_host, with_mock_host};
     use stellar_contract_sdk::Object;
-    use stellar_strkey::PublicKeyEd25519;
+    use stellar_strkey::PublicKey as StrkeyPublicKey;
     use stellar_xdr::{AccountId, AlphaNum4, Asset, AssetCode4, PublicKey, Uint256};
     extern crate alloc;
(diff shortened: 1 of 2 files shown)


examples/liqpool/src/lib.rs
```diff
@@ -340,35 +340,35 @@ mod test {
         let og_host = swap_mock_host(host);
 
         let addr_p = AccountId(PublicKey::PublicKeyTypeEd25519(Uint256(
-            PublicKeyEd25519::from_string(
+            StrkeyPublicKey::from_string(
```
leighmcculloch, line 343: @tomerweller The Strkey and XDR type will often be used together. I had suggested we drop the "Ed25519" from the strkey type, but maybe we should have kept the Ed25519 as it would parallel the XDR type. Wdyt?

examples/liqpool/src/lib.rs
```diff
@@ -328,7 +328,7 @@ mod test {
     use stellar_contract_sdk::testing::mem::{MemHost, MemLedgerKey, MemLedgerVal, MemObj};
     use stellar_contract_sdk::testing::{swap_mock_host, with_mock_host};
     use stellar_contract_sdk::Object;
-    use stellar_strkey::PublicKeyEd25519;
+    use stellar_strkey::PublicKey as StrkeyPublicKey;
```
leighmcculloch, line 331: @tomerweller Since the strkey package's PublicKey will always collide with the xdr packages PublicKey, I wonder if we should just name it `StrkeyPublicKey` in the strkey crate 🤔. It's more likely to result in us naming it the same thing everywhere. Wdyt?

PR #75, opened 2022-05-10
Make Vec fns mutate self

leighmcculloch:
### What

Make Vec's fns that mutate the vector if that is conceptually what they do.

### Why

Vec is immutable but that constraint is unnecessary and prevents us from doing some things in the future we might want to do, like support mutable Vec's on the host side. This will also have some nice side effects that Vec's will work and look more like Rust stdlib Vec's, and it will be clearer whenever copies are being made via .clone(). 

Related conversation here: https://discord.com/channels/897514728459468821/973360018051186798

### Known limitations

N/A

### DIFF
--- a/sdk/src/vec.rs
+++ b/sdk/src/vec.rs
@@ -59,19 +59,17 @@ impl<T: EnvValType + RawValType> Vec<T> {
     // values of T? T values may be objects containing an Env?
 
     #[inline(always)]
-    pub fn put(&self, i: u32, v: T) -> Vec<T> {
+    pub fn put(&mut self, i: u32, v: T) {
         let env = self.env();
         let vec = env.vec_put(self.0.as_ref().val, i.into(), v.into_raw_val(env));
-        let vec: Object = vec.in_env(env).try_into().or_abort();
-        unsafe { Self::unchecked_new(vec) }
+        self.0 = vec.in_env(env).try_into().or_abort();
     }
 
     #[inline(always)]
-    pub fn del(&self, i: u32) -> Vec<T> {
+    pub fn del(&mut self, i: u32) {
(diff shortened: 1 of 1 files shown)


sdk/src/vec.rs
```diff
@@ -82,35 +80,31 @@ impl<T: EnvValType + RawValType> Vec<T> {
...
     }
 
     #[inline(always)]
-    pub fn pop(&self) -> Vec<T> {
+    pub fn pop(&mut self) {
         let env = self.env();
         let vec = env.vec_pop(self.0.as_ref().val);
-        let vec: Object = vec.in_env(env).try_into().or_abort();
-        unsafe { Self::unchecked_new(vec) }
+        self.0 = vec.in_env(env).try_into().or_abort();
```
leighmcculloch, line 93: We could reduce the instructions that get written if the env crate allowed the sdk to do:
```suggestion
        self.0.0.val = vec;
```
@graydon What do you think about us exposing an unsafe set_val-like access?

graydon: seems fine to me. I don't even really know that I'd say it should be unsafe. all the "unsafe" adornments I've been putting on this code so far is like "users stay away, SDK-only", but I don't know that the user is going to be able to mess things up too badly if they swap one val for another on a given EnvVal.

leighmcculloch: With that in mind I think I'll make the EnvVal pub within EnvObj. But that really raises the question: if the EnvVal is public, what is EnvObj achieving....

leighmcculloch: Addressing this another way: https://github.com/stellar/rs-stellar-contract-env/pull/39

PR #128, opened 2022-06-14
Add example using contract data

jonjove:
### What

Add an example that uses contract data.

### Why

This example is used in a stellar-core test to confirm that footprints are working correctly.

### Known limitations

This is based off of a local version of rs-stellar-contract-env, so I needed to comment out some code in bigint.rs to make it compile.

### DIFF
--- a/examples/contract_data/src/lib.rs
+++ b/examples/contract_data/src/lib.rs
@@ -1,12 +1,12 @@
 #![no_std]
-use stellar_contract_sdk::{Env, Symbol};
+use stellar_contract_sdk::{contractfn, Env, Symbol};
 
-#[no_mangle]
+#[contractfn]
 pub fn put(e: Env, key: Symbol, val: Symbol) {
     e.put_contract_data(key, val)
 }
 
-#[no_mangle]
+#[contractfn]
 pub fn del(e: Env, key: Symbol) {
     e.del_contract_data(key)
 }
--- a/examples/contract_data/src/lib.rs
+++ b/examples/contract_data/src/lib.rs
@@ -1,0 +1,12 @@
+#![no_std]
+use stellar_contract_sdk::{Env, Symbol};
+
+#[no_mangle]
+pub fn put(e: Env, key: Symbol, val: Symbol) {
+    e.put_contract_data(key, val)
+}
+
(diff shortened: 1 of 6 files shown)


sdk/src/bigint.rs
```diff
@@ -21,12 +21,12 @@ impl TryFrom<EnvVal<RawVal>> for BigInt {
 impl TryFrom<EnvObj> for BigInt {
     type Error = ();
 
-    fn try_from(obj: EnvObj) -> Result<Self, Self::Error> {
-        if obj.as_tagged().is_obj_type(ScObjectType::Bigint) {
+    fn try_from(_obj: EnvObj) -> Result<Self, Self::Error> {
```
leighmcculloch, line 24: If you merge the main branch into this PR this should be unnecessary.

sdk/src/lib.rs
```diff
@@ -16,7 +16,7 @@ pub use env::RawVal;
 pub use env::Status;
 pub use env::Symbol;
 pub use env::TryFromVal;
-use env::*;
+pub use env::*;
```
leighmcculloch, line 19: Instead of pubing the env to expose the `Env` trait, if you merge the main branch into this PR you should find that the `Env` type in the SDK is a struct that you can add methods to, and then we should add some methods there for accessing contract data and unwrapping/wrapping types that are convertible to and from RawVal.

sdk/src/bigint.rs
jonjove: Fixed.

sdk/src/lib.rs
jonjove: Fixed, thanks.

examples/contract_data/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+#![no_std]
+use stellar_contract_sdk::{Env, Symbol};
+
+#[no_mangle]
```
leighmcculloch, line 4:
```suggestion
#[contractfn]
```

jonjove: Fixed.

leighmcculloch (approved):

PR #133, opened 2022-06-16
Add example of contract calling another contract that generates error

jayz22:
### What

Needed for https://github.com/stellar/rs-stellar-contract-env/pull/230#event-7006705774
Generates two example contracts:
1. Has a single vec function that generates an out-of-bound error 
2. Calls contract 1.
The wasm code are embedded in the contract-env test to observe that errors are propagated properly.  

### Why

### Known limitations

[TODO or N/A]


jayz22:
@leighmcculloch Can this be merged? They are examples used for generating a test on the contract-env. Would hope to get this in soon so the test binary file and the code don't go out of sync again.

PR #148, opened 2022-06-23
Implement Binary and FixedLengthBinary

jonjove:
### What

Implements wrapper classes for `Binary` and `FixedLengthBinary`.

### Why

These are generally useful, but specifically are needed for CAP-54.

### Known limitations

Tests don't currently pass because binary host functions are not implemented in stellar-contract-env-host (cc @jayz22).

### DIFF
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -44,7 +44,7 @@ impl TryFrom<EnvObj> for Binary {
 
     #[inline(always)]
     fn try_from(obj: EnvObj) -> Result<Self, Self::Error> {
-        if obj.as_tagged().is_obj_type(ScObjectType::Vec) {
+        if obj.as_tagged().is_obj_type(ScObjectType::Binary) {
             Ok(unsafe { Binary::unchecked_new(obj) })
         } else {
             Err(())
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -242,3 +242,42 @@ impl<const N: u32> From<FixedLengthBinary<N>> for Binary {
         v.0
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_bin() {
+        let env = Env::default();
+
+        let mut bin = Binary::new(&env);
+        assert_eq!(bin.len(), 0);
+        bin.push(10);
+        assert_eq!(bin.len(), 1);
+        bin.push(20);
+        assert_eq!(bin.len(), 2);
+        bin.push(30);
+        assert_eq!(bin.len(), 3);
+
+        let bin_ref = &bin;
+        assert_eq!(bin_ref.len(), 3);
+
+        let mut bin_copy = bin.clone();
+        assert!(bin == bin_copy);
+        assert_eq!(bin_copy.len(), 3);
+        bin_copy.push(40);
+        assert_eq!(bin_copy.len(), 4);
+        assert!(bin != bin_copy);
+
+        assert_eq!(bin.len(), 3);
+        assert_eq!(bin_ref.len(), 3);
+
+        bin_copy.pop();
+        assert!(bin == bin_copy);
+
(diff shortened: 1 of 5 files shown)


jonjove:
Resolves #147

sdk/src/binary.rs
```diff
@@ -0,0 +1,283 @@
...
+        obj.try_into()
+    }
+}
+
+impl TryFrom<EnvObj> for Binary {
+    type Error = ();
+
+    #[inline(always)]
+    fn try_from(obj: EnvObj) -> Result<Self, Self::Error> {
+        if obj.as_tagged().is_obj_type(ScObjectType::Vec) {
```
leighmcculloch, line 47: A binary is its own type.
```suggestion
        if obj.as_tagged().is_obj_type(ScObjectType::Binary) {
```

sdk/src/binary.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn try_from(bin: Binary) -> Result<Self, Self::Error> {
+        if bin.len() == N {
+            Ok(Self(bin))
+        } else {
+            Err(())
+        }
+    }
+}
+
+impl<const N: u32> AsRef<Binary> for FixedLengthBinary<N> {
```
leighmcculloch, line 211: I think it'd also be worth impl `Deref<Binary>`, although importantly not `DerefMut`. So that all the read-only fns of Binary are available on `FixedLengthBinary`.

sdk/src/binary.rs
```diff
@@ -0,0 +1,283 @@
...
+        self.0 = self
+            .env()
+            .binary_append(self.0.to_tagged(), other.0.to_tagged())
+            .in_env(self.env());
+    }
+}
+
+#[derive(Clone)]
+#[repr(transparent)]
+pub struct FixedLengthBinary<const N: u32>(Binary);
```
leighmcculloch, line 176: We should impl the `put` fn and pass through to the underlying `Binary`.

leighmcculloch, line 176: Suggest we use the term ArrayBinary. It is shorter, and consistent with the idea that an array is fixed length, and consistent with the ArrayVec crate (which we don't use, just mentioning it as an example).
```suggestion
pub struct ArrayBinary<const N: u32>(Binary);
```

leighmcculloch (commented):
Couple asks inline. I think the underlying impl of Binary will change significantly in the near future, but the outer API looks good.

If you don't need it, I would avoid implementing slice for the moment until we've implemented it on Vec.

jayz22:
> Tests don't currently pass because binary host functions are not implemented in stellar-contract-env-host (cc @jayz22).

I updated my branch https://github.com/stellar/rs-stellar-contract-env/pull/131 with most of the binary functions. If you want, can give it a try. Or I can try land the particular ones you need first in order to pass your tests. (That PR will be pretty huge.)

sdk/src/binary.rs
jonjove: Copy-paste error, thanks.

sdk/src/binary.rs
jonjove: Done.

sdk/src/binary.rs
jonjove: This is an [anti-pattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html). I introduced new traits `FixedLengthBinary` and `VariableLengthBinary` which include the functions relevant to those to categories. `Binary` implements `FixedLengthBinary` and `VariableLengthBinary`. `ArrayBinary<N>` implements `FixedLengthBinary` as a pass-through to `Binary`, except that it returns the static length `N` instead of passing-through for `len`.

Now you can operate on either of them generically by using the appropriate trait bounds.

sdk/src/binary.rs
jonjove: Done, https://github.com/stellar/rs-stellar-contract-sdk/pull/148#discussion_r907531873.

PR #160, opened 2022-07-05
Update to latest host env

jayz22:
### What

- Update the host env crate. 
- Fix the `bigint` incompatibility. 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/sdk/src/bigint.rs
+++ b/sdk/src/bigint.rs
@@ -201,7 +201,7 @@ impl Shl<i32> for BigInt {
     type Output = BigInt;
     fn shl(self, rhs: i32) -> Self::Output {
         let env = self.env();
-        let b = env.bigint_shl(self.0.to_tagged(), rhs.into());
+        let b = env.bigint_shl(self.0.to_tagged(), BigInt::from_i32(env, rhs).0.to_tagged());
         Self::try_from_val(env, b).unwrap()
     }
 }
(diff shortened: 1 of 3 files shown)


sdk/src/bigint.rs
```diff
@@ -201,7 +201,7 @@ impl Shl<i32> for BigInt {
     type Output = BigInt;
     fn shl(self, rhs: i32) -> Self::Output {
         let env = self.env();
-        let b = env.bigint_shl(self.0.to_tagged(), rhs.into());
+        let b = env.bigint_shl(self.0.to_tagged(), BigInt::from_i32(env, rhs).0.to_tagged());
```
jonjove, line 204: My understanding was that, per https://github.com/stellar/rs-stellar-contract-env/pull/159, the goal was to get interface uniformity? If so, shouldn't this function take a `BigInt` instead of changing the conversions here? Same thing for `shr`.

jayz22: Yeah that makes sense. I will make the change.
I wasn't sure if from the contract developer perspective it makes more sense to work with `i32` here so I kept the type unchanged. 
Perhaps down the line we want something like: 

```        shl_impl! { $t, u8 }
        shl_impl! { $t, u16 }
        shl_impl! { $t, u32 }
        shl_impl! { $t, u64 }
        shl_impl! { $t, u128 }
        shl_impl! { $t, usize }
        shl_impl! { $t, i8 }
        shl_impl! { $t, i16 }
        shl_impl! { $t, i32 }
        shl_impl! { $t, i64 }
        shl_impl! { $t, i128 }
        shl_impl! { $t, isize }
```
to interop the `BigInt` with different normal integer types, where underneath code wraps them into `BigInt`.

jonjove (approved):

PR #162, opened 2022-07-05
Require TryFrom::Error to impl Debug

### LINKED ISSUE #161: Reduce Usage of map_err

There is an issue around not being able to just `unwrap` certain errors, requiring a `map_err` first. See https://github.com/stellar/rs-stellar-contract-sdk/pull/149#discussion_r905469854.

`unwrap` or `expect` "just work".

None.

leighmcculloch:
As example to https://github.com/stellar/rs-stellar-contract-sdk/pull/149/files#r914120998

Close #161

### DIFF
--- a/sdk/src/env.rs
+++ b/sdk/src/env.rs
@@ -86,7 +86,7 @@ impl Env {
 
     pub fn get_invoking_contract(&self) -> ArrayBinary<32> {
         let rv = internal::Env::get_invoking_contract(self).to_raw();
-        let bin = Binary::try_from_val(self, rv).map_err(|_| ()).unwrap();
+        let bin = Binary::try_from_val(self, rv).unwrap();
         bin.try_into().unwrap()
     }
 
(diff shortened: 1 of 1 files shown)


jonjove:
Related to #161

jonjove (approved):

leighmcculloch:
Note that in the future, once associated type bounds (https://github.com/rust-lang/rust/issues/52662) are stabilized, we'd probably be able to omit the need for the constraint on every function in the SDK. We'd be able to change the definition of `IntoTryFromVal` so that it bounds the `TryFrom::Error` to `Debug` with the following change. Likely a similar change in the env crate would help us out up and down the stack.

```diff
- pub trait IntoTryFromVal: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal> {
+ pub trait IntoTryFromVal: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal, Error: Debug> {
```

PR #171, opened 2022-07-06
Generate Symbol at compile time for User-Defined Type in Contract Spec

leighmcculloch:
### What

Generate Symbol at compiler time for User-Defined Type in Contract Spec.

### Why

Symbols can be created at compile time because the fns for doing so are const fns, but the compiler will only evaluate them at compile time if the const fn is called in a const context. The Symbols for struct fields were being created at runtime because they weren't explicitly in a const context. This change causes them to be evaluated at compile time.

This change reduces the size of `example_udt.wasm` from `1809` to `1522` bytes.

Thanks to @jonjove who noticed this change in size when he started using the contract spec macros, and who noticed that the function being included is probably related to Symbol generation.

### Known limitations

[TODO or N/A]

### DIFF
--- a/macros/src/derive_type.rs
+++ b/macros/src/derive_type.rs
@@ -43,7 +43,7 @@ pub fn derive_type_struct(ident: &Ident, data: &DataStruct, spec: bool) -> Token
                     }
                 }),
             };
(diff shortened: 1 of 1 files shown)


leighmcculloch:
@jonjove FYI that the size difference of the resulting WASM file between using a `Vec` and using a `Map` now that this is fixed is nominal:
- `Map`: `1522`
- `Vec`: `1518`
So it seems using a Map now is not a concern in terms of generated code that ends up in the WASM.

PR #184, opened 2022-07-07
Vec slice, iterators, macro

leighmcculloch:
### What

Add Vec slice, iterators, and macro.

### Why

To improve the ergonomics of using the Vec type.

### Known limitations

[TODO or N/A]

### DIFF
--- a/sdk/src/vec.rs
+++ b/sdk/src/vec.rs
@@ -277,11 +277,29 @@ where
     type Item = T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.0.is_empty() {
+        let len = self.0.len();
+        if len == 0 {
             None
         } else {
-            let item = self.0.get(0);
+            let item = self.0.front();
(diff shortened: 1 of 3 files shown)


sdk/src/vec.rs
```diff
@@ -32,6 +47,17 @@ impl<T: IntoTryFromVal> Ord for Vec<T> {
...
 }
 
+#[cfg(any(feature = "std", test, feature = "testutils"))]
+impl<T> Debug for Vec<T>
+where
+    T: IntoTryFromVal + Debug + Clone,
+    T::Error: Debug,
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "Vec({:?})", self.iter().collect::<std::vec::Vec<_>>())
```
leighmcculloch, line 57: Todo: remove dependence on std, as if just superficial here and being used as a convenience.

PR #222, opened 2022-07-12
Union discriminant encode as Symbol instead of u32

### LINKED ISSUE #218: Contract Spec: Change union discriminant to be symbol instead of index

leighmcculloch:
### What

Union discriminant encode as Symbol instead of u32.

### Why

To make the encoded value more self-descriptive and resilient to changes such as inserting new enum values into earlier positions in the enum.

Close #218

### Known limitations

This increased the UDT example by about 50 bytes.


leighmcculloch:
TODO:
- [ ] Try adding derive Eq PartialEq to Symbol and reverting the if block to a match and seeing if that improves code size.

PR #226, opened 2022-07-12
Remove contractfn macro

### LINKED ISSUE #225: Contract Spec: Remove contractfn
contractfn is a hold over from initial experimentation and we should be using contractimpl instead everywhere, so that there is only one way to export fns and an impl from a contract. Also, contractfn might prevent the effectiveness of doing something like #221 in the future.

We should do this now so that when people start writing contracts we don't get stuck with needing to keep contractfn. Also, it's caused some confusion.

leighmcculloch:
### What
Remove contractfn macro.

### Why

It's causing some confusion because we don't hope people will continue to use it, but most of the examples use it.

Close #225

### Known limitations

[TODO or N/A]

### DIFF
--- a/examples/add_i32/src/lib.rs
+++ b/examples/add_i32/src/lib.rs
@@ -1,9 +1,13 @@
 #![no_std]
-use stellar_contract_sdk::contractfn;
+use stellar_contract_sdk::contractimpl;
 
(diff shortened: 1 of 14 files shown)


leighmcculloch:
cc @jonjove

PR #229, opened 2022-07-12
Add Vec,Map unchecked iter

### LINKED ISSUE #223: Type: Vec,Map: Add unchecked iterators

leighmcculloch:
### What

Add unchecked iter to Vec.

### Why

For use cases where the user doesn't need to check the types of every value and trusts the input.

Close #223

### Known limitations

[TODO or N/A]


sdk/src/map.rs
```diff
@@ -235,6 +237,28 @@ impl<K: IntoTryFromVal, V: IntoTryFromVal> Map<K, V> {
...
+
+    #[inline(always)]
+    pub fn into_iter_unchecked(self) -> UncheckedIter<MapIter<K, V>, (K, V), ConversionError>
+    where
+        K: IntoTryFromVal + Clone,
+        K::Error: Debug,
+        V: IntoTryFromVal + Clone,
+        V::Error: Debug,
+    {
+        self.into_iter().unchecked()
```
leighmcculloch, line 260: There's a tradeoff being made by using the generic UncheckedIter rather than a specific implementation for map. The UncheckedIter has slightly more code because the Map's results are resolved into a single which is then unwrapped by this iterator. If we implemented the Map its own iterator we could exit early on a failure on the key and maybe have one less branch. We can optimize this more probably down the line.

PR #233, opened 2022-07-13
Expose interface version as WASM global

leighmcculloch:
### What

Expose interface version as WASM global.

### Why

The host will read this out of WASM files to check that the WASM is compatible to execute.

Related https://github.com/stellar/rs-stellar-contract-env/pull/203

Close https://github.com/stellar/rs-stellar-contract-sdk/issues/142

### Known limitations

By using a WASM global we're not putting this in a place where we can efficiently build a larger set of meta. We may want to consider putting this into some XDR into a custom section instead. I'm still looking into this.

cc @graydon

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -624,16 +624,16 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 [[package]]
 name = "stellar-contract-env-common"
 version = "0.0.0"
-source = "git+https://github.com/stellar/rs-stellar-contract-env?rev=2d052cb3#2d052cb32e10134c391452c184fc36c1763cd7cf"
+source = "git+https://github.com/stellar/rs-stellar-contract-env?rev=894a749d#894a749d170978a2c56e5225a18e5fcc16853b38"
 dependencies = [
  "static_assertions",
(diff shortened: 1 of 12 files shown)


leighmcculloch:
Commit 936e152 results in the following ending up in the wasm file.

The value of the version is inserted into the `data` section, and a global is created to reference it. I would have thought it would be represented just in the global itself, instead of being a memory reference into the data. I need to do more research, but a quick glance and related issues might suggest Rust does globals this way: https://github.com/rust-lang/rust/issues/65987.

```wat
  (memory (;0;) 17)
  (global (;0;) (mut i32) (i32.const 1048576))
  (global (;1;) i32 (i32.const 1048576))
  (global (;2;) i32 (i32.const 1048584))
  (global (;3;) i32 (i32.const 1048592))
  (export "memory" (memory 0))
  (export "add" (func 8))
  (export "EIV" (global 1))
  (export "__data_end" (global 2))
  (export "__heap_base" (global 3))
  (data (;0;) (i32.const 1048576) "\02"))
```

leighmcculloch:
@graydon I'm leaning towards XDRing this into a custom section. Please see https://github.com/stellar/stellar-xdr-next/pull/10.

graydon:
@leighmcculloch I have no strong preference between XDR-in-a-section and a global.

PR #234, opened 2022-07-13
Export env meta XDR in WASM custom section

### LINKED ISSUE #142: Contract Spec: Include host function, sdk, env versions in WASM
The change https://github.com/stellar/rs-stellar-contract-env/pull/120 renamed host functions. This is fine and we're still iterating and I'm not upset about it but it broke existing wasm blobs (eg. those embedded in tests) that expected the old host function names to exist.

This revealed / reveals a thing I think we should probably encode in contracts: the host function API version they were compiled against. Not because I expect we'll multi-version the API surface or anything -- I expect long-term we'll commit to a strictly backward-compatible API unless there are dangerous bugs -- but because I think @jonjove spent a while hunting the bug and it would be better to give concise diagnostics for such mismatches (new contract running on old host, or compatibility break for whatever reason) during contract instantiation rather than failing inscrutably during execution.

So I think we should have "the version of the host you were compiled against" as metadata pasted into yet another custom section in the wasm (or in the spec section if you can include it there, I don't care).

leighmcculloch:
### What

Expose env meta in WASM custom section.

### Why

So that the host can read it during contract invocation to determine if the contract is compatible with the current host.

Dependent on:
- https://github.com/stellar/rs-stellar-contract-env/pull/203
- https://github.com/stellar/rs-stellar-contract-env/pull/210 
- https://github.com/stellar/rs-stellar-contract-env/pull/212

Close:
- #142 

### Known limitations

[TODO or N/A]

### DIFF
--- a/examples/udt/src/lib.rs
+++ b/examples/udt/src/lib.rs
@@ -1,5 +1,7 @@
 #![no_std]
-use stellar_contract_sdk::{contractimpl, contracttype, ConversionError, IntoEnvVal};
+use stellar_contract_sdk::{contract, contractimpl, contracttype, ConversionError, IntoEnvVal};
+
+contract!();
 
 #[contracttype]
 pub enum UdtEnum {
--- a/sdk/src/env.rs
+++ b/sdk/src/env.rs
@@ -76,14 +76,6 @@ impl Env {
         }
     }
 
-    // TODO: push_test_frame is no longer a pub method, and this needs updating.
-    // #[cfg(feature = "testutils")]
-    // pub fn push_test_frame(&self, contract_id: ArrayBinary<32>) -> FrameGuard {
-    //     self.env_impl
-    //         .push_test_frame(contract_id.into_val(self))
-    //         .unwrap()
-    // }
-
     pub fn get_invoking_contract(&self) -> ArrayBinary<32> {
         let rv = internal::Env::get_invoking_contract(self).to_raw();
         let bin = Binary::try_from_val(self, rv).unwrap();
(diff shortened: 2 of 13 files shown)


sdk/src/env.rs
```diff
@@ -77,12 +76,13 @@ impl Env {
         }
     }
 
-    #[cfg(feature = "testutils")]
-    pub fn push_test_frame(&self, contract_id: ArrayBinary<32>) -> FrameGuard {
-        self.env_impl
-            .push_test_frame(contract_id.into_val(self))
-            .unwrap()
-    }
+    // TODO: push_test_frame is no longer a pub method, and this needs updating.
```
leighmcculloch, line 79: @jonjove After updating the stellar-contract-env-* dependencies this fn caused a compile error in the SDK. I figure you've already got a fix on its way so I've commented it with a TODO temporarily. You'll probably see this as a conflict in your fix, but hopefully it is not too difficult to resolve.

examples/udt/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![no_std]
-use stellar_contract_sdk::{contractimpl, contracttype, ConversionError, IntoEnvVal};
+use stellar_contract_sdk::{contract, contractimpl, contracttype, ConversionError, IntoEnvVal};
+
+contract!();
```
leighmcculloch, line 4: @jonjove We have a lot of options for how we evolve this but there's going to be a macro that we should find a way to require gets included. The macro will embed some metadata into contract. Probably we can embed other test related registrations, vtable generation, etc, in here possibly.

sdk/src/env.rs
jonjove: This function no longer exists in my version, so feel free to just delete it.

leighmcculloch: Done!

PR #245, opened 2022-07-16
Expose the "require" function, use in place of assert!() in example.

graydon:
This might not be _exactly_ the right thing to do; but I noticed one of the examples uses `assert!()` and that pulls in all the formatting machinery and makes the contract wasm 5kb when it is only 1kb when we use something simpler like `require` from env-common.

I'm open to other options here, but .. I think we should provide some help for people to avoid the space overhead of `assert!()`

### DIFF
--- a/examples/linear_memory/src/lib.rs
+++ b/examples/linear_memory/src/lib.rs
@@ -1,5 +1,5 @@
 #![no_std]
-use stellar_contract_sdk::{contractimpl, Binary, Env, FixedLengthBinary};
+use stellar_contract_sdk::{contractimpl, require, Binary, Env, FixedLengthBinary};
 
 pub struct Contract;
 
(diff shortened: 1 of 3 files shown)


examples/linear_memory/src/lib.rs
```diff
@@ -12,18 +12,18 @@ impl Contract {
...
 
     pub fn to_guest(e: Env, b: Binary, b_pos: u32, lm_pos: u32, len: u32) {
         let buf: [u8; 4] = [0; 4];
-        assert!(lm_pos + len > buf.len() as u32);
+        require(lm_pos + len > buf.len() as u32);
         let lm_pos: u32 = unsafe { buf.as_ptr().add(lm_pos as usize) as u32 };
         e.binary_copy_to_linear_memory(b.clone(), b_pos, lm_pos, len);
         for idx in lm_pos..buf.len() as u32 {
-            assert_eq!(buf[idx as usize], b.get(b_pos + idx));
+            require(buf[idx as usize] == b.get(b_pos + idx));
```
leighmcculloch, line 26: The problem isn't `assert!`, the problem is `assert_eq!`. If you simply change this line to as follows, the file size goes down too:
```
assert!(buf[idx as usize] == b.get(b_pos + idx));
```
`panic_immediate_abort` strips debug information from `assert!`. I'm surprised it doesn't do so for `assert_eq!` as well, maybe that's a bug!

leighmcculloch (commented):
I don't think this change is necessary, see below for why.

examples/linear_memory/src/lib.rs
leighmcculloch: So strange, looking at the code triggered by `assert_eq`, it isn't clear to me why there would be additional bloat with it: 
https://github.com/rust-lang/rust/blob/f565016eddc3cb812e647d54b06cfe74bdee2900/library/core/src/macros/mod.rs#L36-L63https://github.com/rust-lang/rust/blob/f565016eddc3cb812e647d54b06cfe74bdee2900/library/core/src/panicking.rs#L205-L231

leighmcculloch:
Closing as this was addressed in #246.

PR #246, opened 2022-07-16
Grab bag

graydon:
Bunch of small stuff I ran into while working on https://github.com/stellar/rs-stellar-contract-env/pull/228

Builds on top of https://github.com/stellar/rs-stellar-contract-sdk/pull/245 -- needs tweaking to whatever gets chosen there if we don't use `require`

### DIFF
--- a/examples/linear_memory/src/lib.rs
+++ b/examples/linear_memory/src/lib.rs
@@ -12,17 +12,17 @@ impl Contract {
         e.binary_new_from_linear_memory(buf.as_ptr() as u32, len)
     }
 
-    pub fn from_guest(e: Env, b: Binary, b_pos: u32, lm_pos: u32, len: u32) -> Binary {
+    pub fn from_guest(e: Env, b: Binary, b_pos: u32, buf_off: u32, len: u32) -> Binary {
         let buf: [u8; 4] = [0, 1, 2, 3];
-        assert!(lm_pos + len > buf.len() as u32);
-        let lm_pos: u32 = unsafe { buf.as_ptr().add(lm_pos as usize) as u32 };
+        assert!(buf_off + len <= buf.len() as u32);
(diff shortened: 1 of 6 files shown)


sdk/src/env.rs
```diff
@@ -14,6 +14,7 @@ pub mod internal {
 
 pub use crate::binary::{ArrayBinary, Binary, FixedLengthBinary};
 pub use internal::meta;
+pub use internal::require;
```
leighmcculloch, line 17: I think this is unnecessary per https://github.com/stellar/rs-stellar-contract-sdk/pull/245.

leighmcculloch (commented):
Lgtm, but I think we should leave require out:

sdk/src/env.rs
graydon: Ok, removed.

leighmcculloch (approved):
🎉

examples/linear_memory/src/lib.rs
```diff
@@ -10,20 +12,20 @@ impl Contract {
         e.binary_new_from_linear_memory(buf.as_ptr() as u32, len)
     }
 
-    pub fn from_guest(e: Env, b: Binary, b_pos: u32, lm_pos: u32, len: u32) -> Binary {
+    pub fn from_guest(e: Env, b: Binary, b_pos: u32, buf_off: u32, len: u32) -> Binary {
         let buf: [u8; 4] = [0, 1, 2, 3];
-        assert!(lm_pos + len > buf.len() as u32);
```
jayz22, line 15: Welp this is embarrassing. 
What I don't understand is how did this pass the [test](https://github.com/stellar/rs-stellar-contract-env/blob/main/stellar-contract-env-host/src/test/binary.rs#L168-L182)? They clearly did not panic or trap.

PR #250, opened 2022-07-18
Add conditional compilation to contractimpl macro

### LINKED ISSUE #244: Exported Contract Functions can Collide when Using Other Crates
In working on an example liquidity pool contract, I am currently depending on the [token contract](https://github.com/stellar/rs-stellar-token-contract) as a normal Cargo dependency

```toml
[dependencies]
# ...
stellar-token-contract = { path = "../rs-stellar-token-contract" }
```

Both contracts contain a `contractimpl` function `initialize`. When building native tests on Mac, these functions seem to suffer a collision at runtime. In other words, calling `liqpool::initialize` inadvertently calls `token::initialize` instead. Changing `liqpool::initialize` to `liqpool::init` resolves this issue.

Ideally, these should either (a) not collide at all or (b) it should be possible to change/disable the exported names when using a dependency.

jonjove:
### What

Makes `contractimpl` optionally take a feature to use for conditionally exporting functions. The old syntax was `#[contractimpl]` and still works (in this case, always export). The new syntax is `#[contractimpl("feature")]` where `feature` is the name of the feature you want to use to _enable_ exporting.

### Why

Resolves #244 

### Known limitations

No tests in macros crate.

### DIFF
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -109,16 +109,10 @@ pub fn derive_fn(
     } else {
         quote! {}
     };
-    let wrap_function = quote! {
-        pub fn #wrap_ident(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
-            <_ as stellar_contract_sdk::IntoVal<stellar_contract_sdk::Env, stellar_contract_sdk::RawVal>>::into_val(
-                #call(
-                    #env_call
-                    #(#wrap_calls),*
-                ),
-                &env
-            )
-        }
+    let export_name = if let Some(cfg_feature) = feature {
+        quote! { #[cfg_attr(feature = #cfg_feature, export_name = #wrap_export_name)] }
+    } else {
+        quote! { #[export_name = #wrap_export_name] }
     };
 
     // Generated code spec.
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):
Very nice. I think we can make this an attribute, so the argument isn't positional, but let's merge this and we can improve that later.

I'm also curious if we should make this a standard/common feature that we expect all contracts to use, rather than make it configurable. So I'd also be okay with us hardcoding a feature which keeps the proc-macro logic simpler.

macros/src/derive_fn.rs
```diff
@@ -121,19 +133,25 @@ pub fn derive_fn(
...
-        pub fn #wrap_ident(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
-            <_ as stellar_contract_sdk::IntoVal<stellar_contract_sdk::Env, stellar_contract_sdk::RawVal>>::into_val(
-                #call(
-                    #env_call
-                    #(#wrap_calls),*
-                ),
-                &env
-            )
+            #[cfg(feature = #cfg_feature)]
+            #[export_name = #wrap_export_name]
```
leighmcculloch, line 142: I think you could probably reuse much of this block and substitute in just the export part instead, and then wrap the export_name in a cfg_attr:
```suggestion
            #[cfg_attr(feature = #cfg_feature, export_name = #wrap_export_name)]
```

jonjove: I was not aware of `cfg_attr`, that's neat. I'll update this which should make the PR quite a bit simpler.

PR #252, opened 2022-07-19
Make contractimpl feature arg an attribute arg

leighmcculloch:
### What

Make contractimpl feature arg an attribute arg.

### Why

We plan to add additional arguments in https://github.com/stellar/rs-stellar-contract-sdk/issues/241, and having this be an argument with a name will make it easier to define the arguments without relying on their position in a list.

### Known limitations

We may prefer to use a crate like darling to simplify our parsing of tokens, and to generate better compile errors on things like typos or unsupported arguments. I didn't want to spend time on integrating that yet though, so I've added a TODO to revisit this.

### DIFF
--- a/examples/udt/Cargo.toml
+++ b/examples/udt/Cargo.toml
@@ -10,3 +10,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 stellar-contract-sdk = {path = "../../sdk"}
+
+[features]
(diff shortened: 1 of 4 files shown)


leighmcculloch:
cc @jonjove This will require a small tweak to your usage of `contractimpl`, but it should be minimal. See the udt example in this PR.

Issue #258, opened 2022-07-20
contractimpl macro doesn't fail when given positional argument


jonjove:
### What did you expect to see?

Related to https://github.com/stellar/rs-stellar-contract-sdk/pull/252, I expected `contractimpl` to only accept attribute arguments.

### What did you see instead?

`contractimpl` accepts positional argument but does nothing in that case.

PR #259, opened 2022-07-20
Improve contractimpl attribute field parsing

leighmcculloch:
### What

Improve contractimpl attribute field parsing.

### Why

### Known limitations

[TODO or N/A]


jonjove:
Should we leave #258 open until the actual issue described there is resolved? I agree with the approach, but I want to make sure we do eventually get this resolved fully.

leighmcculloch:
I think we should consider #258 fixed because the issue you experienced in the case that most others will experience it is addressed. Most developers who experience the issue will do so because they make a typo in a field name, not because they omit the field name. The case you experienced with the field name omitted was unique because the attribute used to behave that way. I doubt most people will omit the field name because it is a less subtle mistake to make. We could say that technically the issue is "won't fix" but I think the issue did highlight a meaningful problem to address. In your case the mistake was made because we had the attribute work in a way that is not common which we've also addressed.

leighmcculloch:
There's a fix for the specific issue of a literal in the meta list in https://github.com/TedDriggs/darling/pull/193.

PR #261, opened 2022-07-20
Change contractimpl to generate module per function

jonjove:
### What

Change `contractimpl` to generate a module per function, containing functions `call` and `call_slice`. Then generate function to register all the contract functions in a `TestContract`.

### Why

Makes the existence of wrapped functions generated by `contractimpl` invisible to developers.

### Known limitations

The design could use some improvement.

### DIFF
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -166,3 +166,24 @@ pub fn derive_fn(
         }
     }
 }
+
+#[allow(clippy::too_many_lines)]
+pub fn derive_add_functions<'a>(
+    ty: &Box<Type>,
+    methods: impl Iterator<Item = &'a syn::ImplItemMethod>,
+) -> TokenStream2 {
+    let (idents, wrap_idents): (Vec<_>, Vec<_>) = methods
+        .map(|m| {
+            let ident = format!("{}", m.sig.ident);
+            let wrap_ident = format_ident!("__{}", m.sig.ident);
+            (ident, wrap_ident)
+        })
+        .multiunzip();
+    quote! {
+        impl stellar_contract_sdk::AddFunctions for #ty {
+            fn add_functions(tc: &mut stellar_contract_sdk::TestContract) {
+                #(tc.add_function(#idents, &#wrap_idents::call_slice));*
+            }
+        }
+    }
+}
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -169,7 +169,7 @@ pub fn derive_fn(
 }
 
 #[allow(clippy::too_many_lines)]
-pub fn derive_add_functions<'a>(
+pub fn derive_contract_function_set<'a>(
     ty: &Box<Type>,
     methods: impl Iterator<Item = &'a syn::ImplItemMethod>,
     feature: &Option<String>,
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -147,7 +147,7 @@ pub fn derive_fn(
 
         pub mod #mod_ident {
             #export_name
-            pub fn call(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
+            pub fn call_raw(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
                 #use_trait;
                 <_ as stellar_contract_sdk::IntoVal<stellar_contract_sdk::Env, stellar_contract_sdk::RawVal>>::into_val(
                     #call(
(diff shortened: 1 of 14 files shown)


macros/src/lib.rs
```diff
@@ -40,22 +48,26 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
+            derive_fn(
+                &call,
+                ident,
+                &m.sig.inputs,
+                &m.sig.output,
+                &args.export_if,
+                &trait_ident,
+            )
         });
+    let add_functions = derive_add_functions(ty, get_methods(&imp), &args.tests_if);
```
leighmcculloch, line 66: This is a smart way to introduce stateless function registration. However, I suspect it will break contracts that have multiple `impl` blocks, since it will implement the `AddFunctions` trait multiple times. Do you know if that will work?

I suspect we need a different approach where we attempt to produce stateful storage, but if what you have in this PR works for at least some contracts, it is probably a good place for us to start and we can create a follow up issue for supporting multiple impl blocks.

leighmcculloch (approved):
I think we need to find a way to hide the `tests_if` feature switch, and I'm concerned about one case, but otherwise this looks good for now and I think we can return to both of these issues if you wouldn't mind creating follow up issues for them.

macros/src/derive_fn.rs
```diff
@@ -136,15 +145,50 @@ pub fn derive_fn(
...
+            (ident, wrap_ident)
+        })
+        .multiunzip();
+    let cfg = if let Some(cfg_feature) = feature {
+        quote! { #[cfg(feature = #cfg_feature)] }
+    } else {
+        quote! {}
+    };
+    quote! {
+        #cfg
```
leighmcculloch, line 187: I think it would be okay to hardcode `testutils` here. I think it will be helpful for the ecosystem if we encourage the testutils pattern, rather than make the macros highly customizable.

leighmcculloch: It also might be good to also couple it with `tests`, so that developers writing tests automatically get access to the additional capabilities if they don't otherwise need to add the testutils feature for their test run, i.e.:
```rust
#[cfg(or(test, feature = "testutils"))]
```

macros/src/derive_fn.rs
```diff
@@ -136,15 +145,50 @@ pub fn derive_fn(
...
+        impl stellar_contract_sdk::AddFunctions for #ty {
+            fn add_functions(tc: &mut stellar_contract_sdk::TestContract) {
```
leighmcculloch, line 189: Could we change this `AddFunctions` trait to `FunctionSet` and have it simply return a list of functions?

The reason I'm suggesting this is that right now this trait and function are bound tightly to `TestContract`, and have this almost-cyclic relationship with it because we give it a TC, and then it calls the TC. The developer calls `my_contract.add_functions(tc)`. I think there are some advantages to presenting it as `tc.add_functions(my_contract.function_set())`, where here we simply provide a function that gives the developer access to the functions that exist, and that can then be hooked up with the TC.

If there are any downsides to this feel free to disregards.

macros/src/derive_fn.rs
```diff
@@ -136,15 +145,50 @@ pub fn derive_fn(
...
-            <_ as stellar_contract_sdk::IntoVal<stellar_contract_sdk::Env, stellar_contract_sdk::RawVal>>::into_val(
-                #call(
-                    #env_call
-                    #(#wrap_calls),*
-                ),
-                &env
-            )
+        pub mod #mod_ident {
+            #export_name
+            pub fn call(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
```
leighmcculloch, line 150: Could we name these functions `call_raw` and `call_raw_slice`? The term `call` on a function a developer has written may be confusing that it doesn't simply and only call it with the parameters I've defined. The `_raw` eludes a little to the parameters being raw vals, and makes it easier to add a `call` that simply calls through to their contract function, which I think we may want to add as well.

leighmcculloch (commented):
I just had a couple thoughts, that I think are minor changes:

macros/src/derive_fn.rs
jonjove: I've made this change because I agree in principle that `call` not doing the obvious thing is surprising. But my general goal is that developers should never see these functions. This will already be true, for example, in rs-stellar-token-contract after this PR.

macros/src/derive_fn.rs
jonjove: I took a totally different approach. `TestContract` is gone. SDK exports a modified version of `ContractFunctionSet` that takes `Env` instead of `&EnvImpl` (to avoid exposing `EnvImpl` to clients). The macro generates an implementation of `ContractFunctionSet`, and then registration is very easy and smooth.

This can easily be extended to support multiple function sets by extending `InternalContractFunctionSet` to contain multiple `ContractFunctionSet`s and chaining them, because they return `Option<...>`.

macros/src/lib.rs
jonjove: Tracked in https://github.com/stellar/rs-stellar-contract-sdk/issues/268

macros/src/derive_fn.rs
jonjove: Opened https://github.com/stellar/rs-stellar-contract-sdk/issues/269

PR #273, opened 2022-07-21
Add iterator, debug support for Binary

jayz22:
### What

Resolves https://github.com/stellar/rs-stellar-contract-sdk/issues/220. 

### Why

### Known limitations

[TODO or N/A]


sdk/src/binary.rs
```diff
@@ -196,6 +201,44 @@ impl VariableLengthBinary for Binary {
...
+            let val = unsafe { <u32 as RawValConvertible>::unchecked_from_val(val) } as u8;
+            Some(val)
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.0.len() as usize;
+        (len, Some(len))
+    }
+}
```
leighmcculloch, line 240: Could you also implement `DoubleEndedIterator`, `ExactSizeIterator`, and `FusedIterator`? See `VecIter` for an example:
https://github.com/stellar/rs-stellar-contract-sdk/blob/7c3f909d519393cad5d11b54ace704e13e619b12/sdk/src/vec.rs#L416-L444

sdk/src/binary.rs
```diff
@@ -196,6 +201,44 @@ impl VariableLengthBinary for Binary {
...
+        for x in iter {
+            write!(f, ", {:?}", x)?;
+        }
+        write!(f, ")")?;
+        Ok(())
+    }
+}
+
+#[derive(Clone)]
+pub struct BinIter(Binary);
```
leighmcculloch, line 220: Could you move this to under the IntoIterator impl? It might be helpful if the file is structured with these types in the same order as vec.rs, just for general readability.

leighmcculloch (commented):
Could we add an iterator impl for ArrayBinary as well?

sdk/src/binary.rs
jayz22: Done.

sdk/src/binary.rs
jayz22: Done.

jayz22:
> Also could we add an iterator impl for ArrayBinary as well 

Done!

leighmcculloch (approved):

PR #277, opened 2022-07-22
Add ScVal conversions for SDK types

### LINKED ISSUE #274: Test Fwk: Add conversions from ScVal to SDK types using new ScVal conversions in env

leighmcculloch:
### What
Add ScVal conversions for SDK types.

### Why
So that user-defined types that contain SDK types can be converted into ScVal's for the external style testing and signature producing that the token contract does.

Close #274

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,6 +197,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dyn-fmt"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2a0836c9bd73a9d3ca55b0effc5b1eedf96dd13ef994389bcac6d4d33c46188"
+
+[[package]]
 name = "ed25519"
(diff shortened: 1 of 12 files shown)


leighmcculloch:
I added `ScVal` conversions by leveraging the existing EnvVal type and all of it's conversion traits, like `TryFromVal`. This required loosening it's absolute constraints in https://github.com/stellar/rs-stellar-contract-env/pull/247. This makes `EnvVal` not necessarily a `RawVal` or `TaggedVal`, it can contain any type. _All_ the things we care about inside a contract function are still true, there is no breaking changes, because all the functions and behaviors that existed are still true for `EnvVal` when it contains a `RawVal` or `TaggedVal`.

This PR crosses the boundary of complexity that I'm comfortable with usually. I'm not particularly happy with that, but the alternative was to duplicate some of the existing conversion system, and none of this complexity leaks into contracts, so I went with it.

I do have the feeling we will revisit some of this though.

PR #278, opened 2022-07-22
Update stellar-contract-env

leighmcculloch:


sdk/src/env.rs
```diff
@@ -270,6 +270,34 @@ impl internal::EnvBase for Env {
             env_impl: self.env_impl.deep_clone(),
         }
     }
+
+    fn binary_copy_from_slice(&self, _: Object, _: RawVal, _: &[u8]) -> Object {
```
leighmcculloch, line 274: @graydon I assume I need to impl this on the guest side.

leighmcculloch: Wait, do I even need an EnvBase impl on the guest side at all?

leighmcculloch: Nope, looks like it is required. @graydon Imo we should move these new functions into a new trait, something like `EnvBinary`, and then we can write other constraints on functions to say those functions and types only work if given an `Env + EnvBinary`. It feels wrong that we're polluting `unimplemented!()` around the place, and then trusting we'll just never call these functions.

PR #286, opened 2022-07-23
Add helpers for asserting that test code panics

graydon:
This adds some helpers to `Env` which help write tests that assert panics happen in narrow situations.

They are broadly wrappers around `catch_unwind` with a few caveats:

  - They capture some boilerplate for downcasting and inspecting the value carried in the panic
  - The status-seeking one actually tries to fish a matching status out of the debug log if the panic doesn't have it
  - They try to improve the odds of unwind-safety by deep-cloning the Env before getting under way
  - They remove and reinstall a panic hook so the test output isn't cluttered with spurious backtraces
 
Feel free to muck around with this, I just thought it might be useful for me to sketch.

### DIFF
--- a/sdk/src/env.rs
+++ b/sdk/src/env.rs
@@ -254,6 +254,89 @@ impl Env {
     pub fn invoke_contract(&mut self, hf: xdr::HostFunction, args: xdr::ScVec) -> xdr::ScVal {
         self.env_impl.invoke_function(hf, args).unwrap()
     }
+
+    #[cfg(not(target_family = "wasm"))]
+    fn clone_self_and_catch_panic<F, T>(&self, f: F) -> (Env, std::thread::Result<T>)
+    where
+        F: FnOnce(Env) -> T,
+    {
+        let hook = std::panic::take_hook();
+        std::panic::set_hook(Box::new(|_| ()));
+        let deep_clone = self.deep_clone();
+        let res = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| f(deep_clone.clone())));
+        std::panic::set_hook(hook);
+        (deep_clone, res)
+    }
+
+    #[cfg(not(target_family = "wasm"))]
+    pub fn assert_panic_with_string<F, T: Debug>(&self, s: &str, f: F)
+    where
+        F: FnOnce(Env) -> T,
+    {
+        match self.clone_self_and_catch_panic(f) {
+            (_, Ok(v)) => panic!("inner function expected to panic, but returned {:?}", v),
+            (_, Err(e)) => match e.downcast_ref::<String>() {
+                None => match e.downcast_ref::<&str>() {
+                    Some(ps) => assert_eq!(*ps, s),
+                    None => panic!(
+                        "inner function panicked with unknown type when \"{}\" expected",
+                        s
+                    ),
+                },
+                Some(ps) => assert_eq!(*ps, s),
+            },
+        }
+    }
+
+    #[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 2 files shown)


sdk/src/env.rs
```diff
@@ -254,6 +254,89 @@ impl Env {
...
+    fn clone_self_and_catch_panic<F, T>(&self, f: F) -> (Env, std::thread::Result<T>)
+    where
+        F: FnOnce(Env) -> T,
+    {
+        let hook = std::panic::take_hook();
+        std::panic::set_hook(Box::new(|_| ()));
+        let deep_clone = self.deep_clone();
+        let res = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| f(deep_clone.clone())));
+        std::panic::set_hook(hook);
+        (deep_clone, res)
```
leighmcculloch, line 268: ❔ Is the deep clone happening so that if the panic occurs we have a complete copy of the host that is safe to use regardless of what happened within the host? That's pretty cool.

sdk/src/env.rs
```diff
@@ -254,6 +254,89 @@ impl Env {
...
         self.env_impl.invoke_function(hf, args).unwrap()
     }
+
+    #[cfg(not(target_family = "wasm"))]
+    fn clone_self_and_catch_panic<F, T>(&self, f: F) -> (Env, std::thread::Result<T>)
+    where
+        F: FnOnce(Env) -> T,
+    {
+        let hook = std::panic::take_hook();
+        std::panic::set_hook(Box::new(|_| ()));
```
leighmcculloch, line 264: ❔ What's the significant of setting a panic handling that does nothing during the clone? If a panic occurs during clone for some reason won't it be gobbled and ignored?

sdk/src/env.rs
```diff
@@ -254,6 +254,89 @@ impl Env {
...
+        let deep_clone = self.deep_clone();
+        let res = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| f(deep_clone.clone())));
+        std::panic::set_hook(hook);
+        (deep_clone, res)
+    }
+
+    #[cfg(not(target_family = "wasm"))]
+    pub fn assert_panic_with_string<F, T: Debug>(&self, s: &str, f: F)
```
leighmcculloch, line 272: 💡 I think we're making more assumptions than we need to about how developers will want to assert on panic / status errors by providing functions that retrieve the value and assert on it. I think we could defer the decision to developers on how they will assert on the error by simply returning the error to them in a `Result`.

This will have two advantages:

- For tests that have a return value as a Result, and for calls that the developer doesn't want to fail, they can simply chuck a `?` on the end and they can surface a `Status` error.
   ```rust
   let e = Env::with_empty_recording_storage();
   let got = e.capture_status(|env| ...)?;
   ```

- For developers that want to assert more complex things than a simple equality, they can do that however they like. And equality checks will still be easy:
   ```rust
   let e = Env::with_empty_recording_storage();
   let got: Result<_, Status> = e.capture_status(|env| ...);
   assert_eq!(got, Err("panic string here"))
   ```

I think both examples above work with strings and statuses. And then we can apply the same pattern to the `Env::invoke_contract` functions so that whenever you invoke a contract you get back a `Result<_, Status>`.

cc @graydon @jonjove

leighmcculloch (commented):
I'm lacking some knowledge about how some of this works so I've asked some questions (❔).

I also wonder if we should tweak the interface we're providing to simply provide data and let the developer assert how they'd like, see the 💡 comment inline.

leighmcculloch (approved):
Regardless of above ☝🏻, 👍🏻 🎉.

PR #291, opened 2022-07-23
Add [u8; N] => ArrayBinary conversion

leighmcculloch:
### What
Add [u8; N] => ArrayBinary conversion.

### Why
The token contract seems to need to do a lot of these conversions in tests.

### Known Limitations
The conversion being added is one way. I think we should probably just add these as we go so as not to burn too much time on them right now.

The conversion has to be fallible because the type of length of an array and the ArrayBinary type are different, being `usize` and `u32` respectively. We should rethink this, because it is very inconvenient and this conversion could be infallible otherwise. (cc @jonjove interested to hear what you think about this)

### DIFF
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -6,9 +6,12 @@ use core::{
 };
 
 use super::{
-    env::internal::Env as _, xdr::ScObjectType, ConversionError, Env, EnvObj, EnvVal, RawVal,
-    RawValConvertible,
+    env::internal::Env as _, env::EnvType, xdr::ScObjectType, ConversionError, Env, EnvObj, EnvVal,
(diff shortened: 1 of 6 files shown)


sdk/src/binary.rs
```diff
@@ -403,6 +403,25 @@ impl<const N: u32> FixedLengthBinary for ArrayBinary<N> {
     }
 }
 
+impl<const N: usize, const M: u32> TryFrom<EnvType<[u8; N]>> for ArrayBinary<M> {
```
leighmcculloch, line 406: This probably looks a bit odd, and we probably need a better name for `EnvType`, but what this conversion is saying is that you can convert from an `[u8; N]` to an `ArrayBinary<N>` if you also have an `Env` present.

Issue #173, opened 2022-07-06
API: Polish SDK API and Types Interfaces to Host Fns


leighmcculloch:

leighmcculloch:
I'll be keeping this comment updated with the list of things.

Some things to do:
- Get rid of Fixed/Var Binary traits #294 
- Make Binary/ArrayBinary fns consistent with Vec #295 
- Change `invoke_contract` accept Into<ScVal> https://github.com/stellar/rs-stellar-contract-sdk/issues/296
- Fix vec, map https://github.com/stellar/rs-stellar-contract-sdk/issues/282
- Move all of the testutils types and functionality into a testutils package.
- Inherent impls for BigInt, Vec, Map for trait impls
- Rename Vec/Binary `put` to `set`
- Remove repetitive words like `contract` from env fns
- Rename ArrayBinary to FixedBinary
- Update token contract and example contracts for all the above things

Some things to investigate:
- Change BitSet, Status, Symbol in the SDK to be struct types wrapping env types, to make the docs 100x better for these types.
- Renaming Binary to Vec<u8> (and FixedBinary to FixedVec<u8>)

PR #294, opened 2022-07-25
Remove FixedLengthBinary and VariableLengthBinary traits

leighmcculloch:
### What
Remove `FixedLengthBinary` and `VariableLengthBinary` traits. Also implement `Borrow`, `AsRef`, and `AsMut` for `ArrayBinary<N>` => `&Binary`.

### Why
The traits `FixedLengthBinary` and `VariableLengthBinary` add complexity to the concepts a developer needs to learn when interacting with binary that is not outsized by the benefits the type add.

The types make it possible for functions to indicate if when accepting a mutable reference to a binary if that function is going to modify the binary in a way that will change its size or not. In conversations with @jonjove he mentioned he has only found a use for this benefit in tests and not really in contract code.

I think we can support most use cases with the `Binary` and `ArrayBinary` types, without the traits.

For any function wanting to accept either a `Binary` or `ArrayBinary` the traits are unnecessary, because the function can accept either by accepting `Borrow<Binary>`, or `AsRef<Binary>`.

For #173

### DIFF
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -114,10 +114,18 @@ use super::{
 };
 
 #[cfg(not(target_family = "wasm"))]
+impl TryFrom<&Binary> for ScVal {
+    type Error = ConversionError;
+    fn try_from(v: &Binary) -> Result<Self, Self::Error> {
+        (&v.0).try_into().map_err(|_| ConversionError)
+    }
+}
+
+#[cfg(not(target_family = "wasm"))]
 impl TryFrom<Binary> for ScVal {
     type Error = ConversionError;
     fn try_from(v: Binary) -> Result<Self, Self::Error> {
-        v.0.try_into().map_err(|_| ConversionError)
+        (&v).try_into()
     }
 }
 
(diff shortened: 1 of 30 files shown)


sdk/src/binary.rs
```diff
@@ -371,35 +350,45 @@ impl<const N: u32> Ord for ArrayBinary<N> {
...
+impl<const N: u32> Borrow<Binary> for &mut ArrayBinary<N> {
+    fn borrow(&self) -> &Binary {
+        &self.0
     }
+}
 
-    #[inline(always)]
-    fn front(&self) -> u8 {
-        self.0.front()
+impl<const N: u32> BorrowMut<Binary> for &mut ArrayBinary<N> {
```
jonjove, line 377: Isn't this unsafe? I could change the length of an `ArrayBinary` with this.

sdk/src/binary.rs
```diff
@@ -371,35 +350,45 @@ impl<const N: u32> Ord for ArrayBinary<N> {
...
-    #[inline(always)]
-    fn back(&self) -> u8 {
-        self.0.back()
+impl<const N: u32> AsRef<Binary> for ArrayBinary<N> {
+    fn as_ref(&self) -> &Binary {
+        &self.0
+    }
+}
+
+impl<const N: u32> AsMut<Binary> for ArrayBinary<N> {
```
jonjove, line 389: Isn't this unsafe? I could change the length of an `ArrayBinary` with this.

jonjove (commented):
This concept seems fine, but aren't some of these conversions unsafe?

sdk/src/binary.rs
leighmcculloch: You're right. I'll remove the BorrowMut and AsMut.

PR #295, opened 2022-07-25
Change interface of Binary to match Vec

leighmcculloch:
### What

Change interface of Binary to match Vec.

### Why

The Binary interface was always supposed to match the Vec interface, and they did originally, but then Vec got changed and Binary didn't.

For https://github.com/stellar/rs-stellar-contract-sdk/issues/173


jonjove (approved):
Agreed, these should definitely match. I haven't confirmed that this PR matches them perfectly, but I trust you have done that.

PR #303, opened 2022-07-25
Impl debug for `BigInt`

jayz22:
### What

Resolves https://github.com/stellar/rs-stellar-contract-sdk/issues/219

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/sdk/src/bigint.rs
+++ b/sdk/src/bigint.rs
@@ -4,11 +4,9 @@ use core::{
     ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Neg, Not, Rem, Shl, Shr, Sub},
 };
 
-use stellar_contract_env_host::RawValConvertible;
-
 use super::{
     env::internal::Env as _, xdr::ScObjectType, Binary, ConversionError, Env, EnvBase, EnvObj,
-    EnvVal, RawVal, TryFromVal,
+    EnvVal, RawVal, RawValConvertible, TryFromVal,
 };
 
 #[repr(transparent)]
--- a/sdk/src/bigint.rs
+++ b/sdk/src/bigint.rs
@@ -1,16 +1,44 @@
 use core::{
     cmp::Ordering,
+    fmt::Debug,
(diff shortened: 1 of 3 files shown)


sdk/src/bigint.rs
```diff
@@ -1,17 +1,43 @@
...
-    RawVal, TryFromVal,
+    env::internal::Env as _, xdr::ScObjectType, Binary, ConversionError, Env, EnvBase, EnvObj,
+    EnvVal, RawVal, RawValConvertible, TryFromVal,
 };
 
 #[repr(transparent)]
 #[derive(Clone)]
 pub struct BigInt(EnvObj);
 
+impl Debug for BigInt {
```
leighmcculloch, line 16: We could add a `Display` version of this that just outputs the number without the `BigInt(` prefix and `)` suffix. cc @paulbellamy mentioned today he'd like this.

sdk/src/bigint.rs
```diff
@@ -1,17 +1,43 @@
...
+        let bi = self.0.to_tagged();
+        let obj: EnvObj = env.bigint_to_radix_be(bi, 10u32.into()).in_env(env);
+        write!(f, "BigInt(")?;
+        if let Ok(bin) = TryInto::<Binary>::try_into(obj) {
+            let sign = env.bigint_cmp(bi, env.bigint_from_u64(0));
+            match unsafe { <i32 as RawValConvertible>::unchecked_from_val(sign) } {
+                -1 => write!(f, "-")?,
+                0 => {
+                    write!(f, "0)")?;
+                    return Ok(());
```
leighmcculloch, line 28: This feels like a gotcha that we exit early here and append the `)` and not on other cases. Is it necessary to exit early here? Looks like it should work fine to continue since the binary should be empty?

leighmcculloch (commented):
One concern (❗), otherwise LTGM 👏🏻

sdk/src/bigint.rs
jayz22: You are right. Fixed.

sdk/src/bigint.rs
jayz22: Added.

leighmcculloch (approved):

PR #310, opened 2022-07-25
Move Binary serialize/deserialize into traits

leighmcculloch:
### What
Move Binary serialize/deserialize into traits onto all the types.

### Why
So that the more natural `my_value.serialize(&env)` that is more similar to the interfaces that libraries like serde provide, and can be used instead of `env.serialize_to_binary`.

For #173

### DIFF
--- a/tests/udt/src/lib.rs
+++ b/tests/udt/src/lib.rs
@@ -36,7 +36,127 @@ impl Contract {
 #[cfg(test)]
 mod test {
     use super::{UdtEnum, UdtStruct, __add};
-    use stellar_contract_sdk::{vec, xdr::ScVal, Env, IntoVal, TryFromVal};
+    use stellar_contract_sdk::{vec, xdr::ScVal, Binary, Env, IntoVal, TryFromVal};
+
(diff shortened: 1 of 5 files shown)


tests/udt/src/lib.rs
```diff
@@ -36,7 +36,127 @@ impl Contract {
...
+            bin.push(0);
+            bin.push(0);
+            bin.push(0);
+            bin.push(0);
+            bin.push(0);
+            bin.push(0);
+            bin.push(1);
+            bin
+        })
+    }
```
jayz22, line 159: If you prefer not writing these repetitive lines, the `binary_new_from_linear_memory` may serve you. Here is the example: 
https://github.com/stellar/rs-stellar-contract-sdk/pull/133/files#diff-4a1250d4e98a8c8930b456dfeb1b323756724a164cba250df28c0d71bbabe722R11-R12

leighmcculloch: Thanks, I'll follow up with fixing that up.

PR #312, opened 2022-07-25
Make Sign work with TryInto<ScVal>

jonjove:
### What

Change `M: Into<ScVal>` to `M: TryInto<ScVal>`.

### Why

Support UDTs.

### Known limitations

Conversion error reporting is a bit weird.


leighmcculloch (approved):
I'd like to circle back on trying to remove the dependence on std::error::Error, because it's confusing to me that this is compatible with conversion errors from the SDK, given that the SDK doensn't support std. But if this works 👏🏻

PR #314, opened 2022-07-25
Make UDT ScMaps sorted

leighmcculloch:
### What
Sort the ScMap's that are generated from user-defined types.

### Why
If they aren't sorted they're invalid. If they aren't sorted it can cause many different problems for applications that do things with the serialized version of the map. One example is signing or hashing the structure will be inconsistent with the host version which is always sorted internally.


jonjove (approved):
Just tested this, it fixes the issue.

PR #316, opened 2022-07-25
Add call_external functions to derive_fn macro

### LINKED ISSUE #243: Test Fwk: Generate another wrap method that calls the contract fn via the env.invoke_function (macros)

jonjove:
Resolves #243

### DIFF
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -77,7 +77,6 @@ pub fn derive_fn(
                         #ident
                     ).unwrap()
                 };
-                let pat_ty = &pat_type.ty;
                 let invoke_arg = FnArg::Typed(PatType {
                     attrs: vec![],
                     pat: Box::new(Pat::Ident(PatIdent {
@@ -88,7 +87,7 @@ pub fn derive_fn(
                         subpat: None,
                     })),
                     colon_token: Colon::default(),
-                    ty: Box::new(Type::Verbatim(quote! { super::#pat_ty })),
+                    ty: pat_type.ty.clone(),
                 });
                 let invoke_call = quote! { #ident };
                 (spec, arg, call, invoke_arg, invoke_call)
@@ -160,6 +159,8 @@ pub fn derive_fn(
         pub static #spec_ident: [u8; #spec_xdr_len] = *#spec_xdr_lit;
 
         pub mod #mod_ident {
+            use super::*;
+
             #export_name
             pub fn call_raw(env: stellar_contract_sdk::Env, #(#wrap_args),*) -> stellar_contract_sdk::RawVal {
                 #use_trait;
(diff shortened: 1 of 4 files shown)


macros/src/derive_fn.rs
```diff
@@ -77,11 +77,24 @@ pub fn derive_fn(
...
+                        ident: ident.clone(),
+                        attrs: vec![],
+                        by_ref: None,
+                        mutability: None,
+                        subpat: None,
+                    })),
+                    colon_token: Colon::default(),
+                    ty: Box::new(Type::Verbatim(quote! { super::#pat_ty })),
+                });
+                (spec, arg, call, invoke_arg, quote! { #ident })
```
leighmcculloch, line 93: It helps with readability in the long tuple (too long!) if we name the args here.
```suggestion
                let invoke_call = quote! { #ident };
                (spec, arg, call, invoke_arg, invoke_call)
```

macros/src/derive_fn.rs
```diff
@@ -164,6 +177,21 @@ pub fn derive_fn(
             ) -> stellar_contract_sdk::RawVal {
                 call_raw(env, #(#slice_args),*)
             }
+
+            #[cfg(feature = "testutils")]
+            #[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
+            pub fn call_external(
```
leighmcculloch, line 183: We don't need to resolve this immediately right now before merging, but we need to come up with a good naming strategy for the different ways we call things, because I think "external" is ambiguous. It's possible we can't do better than this though. I at least can't think of anything better in this moment.

macros/src/derive_fn.rs
```diff
@@ -164,6 +177,21 @@ pub fn derive_fn(
...
+                e: &mut stellar_contract_sdk::Env,
+                contract_id: &stellar_contract_sdk::Binary,
+                #(#invoke_args),*
+            ) {
+                e.invoke_contract(
+                    stellar_contract_sdk::xdr::HostFunction::Call,
+                    (contract_id, #wrap_export_name, #(#invoke_idents),*).try_into().unwrap()
+                )
+                .try_into()
+                .unwrap()
```
leighmcculloch, line 193: We might want to return the `Result` given that the function will be used in testing, there's a good chance people will be asserting on the return value, and it is easy to assert on a `Result` which makes testing a bit clearer imo.

leighmcculloch (approved):
👏🏻 This is wonderful. Left comments, but none need to block merging this.

leighmcculloch:
@jonjove Looks like there is a build error. I don't think it was the commit I pushed as the original PR commit failed too.

macros/src/derive_fn.rs
jonjove: Opened #329

macros/src/derive_fn.rs
jonjove: Opened #330

PR #320, opened 2022-07-26
Change u32 to usize for indexes and lengths

leighmcculloch:
### What
Change u32 to usize for indexes and lengths of collections.

### Why
It is inconvenient that the lengths and indexes of these types are not usize. For example, if I create conversion from a fixed size array to a Vec, or a Binary, I can't do so relying on the contract to guarantee that the size of both, for fixed size structures, are identical. I have to rely on a runtime check. In general it seems a bit odd we'd not use usize at the seams, assuming there's not some cost to doing so.

### Affect on Contract Size

For some reason the `create_contract` test went down in size, unclear why.

As expected the `linear_memory` test went up in size, likely due to the additional `u32::MAX` checks, not substantially though.

#### Before
```-rw-r--r--  1 leighmcculloch  staff  345 Jul 25 17:31 example_add_i32.wasm
-rw-r--r--  1 leighmcculloch  staff  541 Jul 25 17:31 example_add_i64.wasm
-rw-r--r--  1 leighmcculloch  staff  617 Jul 25 17:31 example_contract_data.wasm
-rw-r--r--  1 leighmcculloch  staff  566 Jul 25 17:31 example_create_contract.wasm
-rw-r--r--  1 leighmcculloch  staff  1163 Jul 25 17:31 example_linear_memory.wasm
-rw-r--r--  1 leighmcculloch  staff  2121 Jul 25 17:31 example_udt.wasm
-rw-r--r--  1 leighmcculloch  staff  103 Jul 25 17:31 stellar_contract_sdk.wasm
```

#### After
```-rw-r--r--  1 leighmcculloch  staff  345 Jul 25 17:30 example_add_i32.wasm
-rw-r--r--  1 leighmcculloch  staff  541 Jul 25 17:30 example_add_i64.wasm
-rw-r--r--  1 leighmcculloch  staff  617 Jul 25 17:30 example_contract_data.wasm
-rw-r--r--  1 leighmcculloch  staff  548 Jul 25 17:30 example_create_contract.wasm
-rw-r--r--  1 leighmcculloch  staff  1197 Jul 25 17:30 example_linear_memory.wasm
-rw-r--r--  1 leighmcculloch  staff  2121 Jul 25 17:30 example_udt.wasm
-rw-r--r--  1 leighmcculloch  staff  103 Jul 25 17:30 stellar_contract_sdk.wasm
```

### DIFF
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -618,7 +618,7 @@ impl<const N: usize> FixedBinary<N> {
     }
 }
 
-impl<const N: u32> IntoIterator for FixedBinary<N> {
+impl<const N: usize> IntoIterator for FixedBinary<N> {
     type Item = u8;
 
     type IntoIter = BinIter;
--- a/sdk/src/binary.rs
+++ b/sdk/src/binary.rs
@@ -6,6 +6,8 @@ use core::{
     ops::{Bound, RangeBounds},
 };
 
+use crate::u32usize::u32_to_usize;
+
 use super::{
     env::internal::Env as _, env::EnvType, xdr::ScObjectType, ConversionError, Env, EnvObj, EnvVal,
     Object, RawVal, RawValConvertible,
@@ -212,7 +214,7 @@ impl Binary {
     pub fn len(&self) -> usize {
         let env = self.env();
         let val = env.binary_len(self.0.to_tagged());
-        u32::try_from(val).unwrap() as usize
+        u32_to_usize(u32::try_from(val).unwrap())
     }
 
     #[inline(always)]
@@ -378,6 +380,7 @@ impl Iterator for BinIter {
         } else {
             let val = self.0.env().binary_front(self.0 .0.to_object());
             self.0 = self.0.slice(1..);
+            #[allow(clippy::cast_possible_truncation)]
             let val = unsafe { <u32 as RawValConvertible>::unchecked_from_val(val) } as u8;
             Some(val)
         }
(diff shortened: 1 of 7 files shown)


sdk/src/binary.rs
```diff
@@ -401,86 +407,77 @@ impl FusedIterator for BinIter {}
...
-            return Err(ConversionError);
-        }
+impl<const N: usize> From<EnvType<[u8; N]>> for FixedBinary<N> {
+    fn from(ev: EnvType<[u8; N]>) -> Self {
         let mut bin = Binary::new(&ev.env);
         for b in ev.val {
             bin.push(b);
         }
-        bin.try_into()
+        FixedBinary::<N>(bin)
```
leighmcculloch, line 476: This is the main advantage.

tests/linear_memory/src/lib.rs
```diff
@@ -23,7 +23,7 @@ impl Contract {
         let lm_pos: u32 = unsafe { buf.as_ptr().add(buf_off as usize) as u32 };
         e.binary_copy_to_linear_memory(b.clone(), b_pos, lm_pos, len);
         for idx in lm_pos..buf.len() as u32 {
-            assert!(buf[idx as usize] == b.get_unchecked(b_pos + idx));
+            assert!(buf[idx as usize] == b.get_unchecked((b_pos + idx) as usize));
```
leighmcculloch, line 26: This is the main cost. There will be places that u32 needs force converting to usize in contracts.

leighmcculloch:
I'll add another example that exercises Binary and see if I can whittle the size difference away.

leighmcculloch:
Ok, so this is very doable. But has some downsides I didn't expect. Specifically, if someone wants to pass an index through a contract fns arguments it'll need to be a u32, which means developers need to cast to usize. We can't surface usize as a RawVal as it would present too many gotchas on the host side I think. So I'm just going to close this and keep it as u32 which presents mostly internal annoyances.

leighmcculloch:
I realized there's a simpler solution to my problem: https://github.com/stellar/rs-stellar-contract-sdk/pull/322.

PR #322, opened 2022-07-26
Use usize for FixedBinary<N> but u32 at runtime

leighmcculloch:
### What
Use usize for the const generic in FixedBinary<N>, but continue to use u32 at runtime.

### Why
We can't do some infallible conversions when the const generic is u32, like the one changed in this PR. We need u32 at runtime for the reasons expressed in #320.


leighmcculloch:
fyi @graydon @jonjove

PR #334, opened 2022-07-26
Add Account type and move fns from Env

leighmcculloch:
For #173


leighmcculloch:
Holding merging this until @jonjove finishes tweaking the invoke_contract calls.

PR #356, opened 2022-07-27
Add call_internal

jonjove:
### What

Add `call_internal` that is analogous to `call_external` but uses `invoke_contract` instead of `invoke_contract_external_raw`.

### Why

Makes cross-contract calls easy.

### Known limitations

This probably makes `call_external` obsolete, but I'm not removing it now.

### DIFF
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -185,6 +185,17 @@ pub fn derive_fn(
                 call_raw(env, #(#slice_args),*)
             }
 
+            pub fn call_internal(
+                e: &stellar_contract_sdk::Env,
+                contract_id: &stellar_contract_sdk::Binary,
+                #(#invoke_args),*
+            ) #output {
+                use stellar_contract_sdk::{EnvVal, IntoVal, Symbol, Vec};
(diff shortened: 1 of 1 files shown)


jonjove:
cc @leighmcculloch because I mentioned this idea to you. Follow-on PRs coming in contracts that will show how this make things a lot better.

leighmcculloch (commented):
Excellent. We have four ways to call contracts now, and I think we should delete any that we don't want people to use, or, at least mark them as deprecated. Thoughts? cc @jonjove

jonjove:
Already opened #357 for that, do you want to get that done in this iteration?

leighmcculloch:
> do you want to get that done in this iteration?

There's value in aggressively reducing this I think, mostly so that people don't accidentally use things we plan to take away anyway. But if you think it isn't clear which is better, I think it's fine to keep them around, but lets maybe add a deprecated notice to them.

I also think we should think about what it should look like if there was only `call_internal`. If there was only that fn, it would probably just be `call` or `invoke` (if we use the same terminology as elsewhere in the SDK).

leighmcculloch:
I've opened #358 instead.

PR #358, opened 2022-07-27
Rename the invoke functions

leighmcculloch:
### What
Rename the functions available for invoking contracts:

These functions are intended for contract developer use:
| From | To |
| --- | --- |
| `__fn::call_internal` | `fn::invoke` |
| `__fn::call_external` | `fn::invoke_xdr` |

These functions are not intended for contract developer use:
| From | To |
| --- | --- |
| `__fn::call_raw` | `__fn::invoke_raw` |
| `__fn::call_raw_slice` | `__fn::invoke_raw_slice` |

### Why
In examples it appears we expect people to call some of these functions. We should remove the double underscore.

Also, we should change `call` to `invoke`, because everywhere else we call this invoking.

Instead of `internal` and `external`, we should use pre-existing terms. The external mode uses XDR types, so let's use XDR. And the non-XDR form is the form we expect them to use, and since it uses the same type system that the contract itself uses, I think we can drop any suffix and simply call it `invoke`.

### DIFF
--- a/macros/src/derive_fn.rs
+++ b/macros/src/derive_fn.rs
@@ -125,6 +125,7 @@ pub fn derive_fn(
     // Generated code parameters.
     let wrap_export_name = format!("{}", ident);
     let mod_ident = format_ident!("__{}", ident);
+    let pub_mod_ident = format_ident!("{}", ident);
     let env_call = if env_input.is_some() {
         quote! { env.clone(), }
     } else {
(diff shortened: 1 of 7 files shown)


jonjove:
@leighmcculloch I generally feel good about these suggestions. At a minimum, they are better than where we are now. One question: `invoke` doesn't make it clear that it's a cross-contract call, so should we make that more obvious?

leighmcculloch:
> One question: `invoke` doesn't make it clear that it's a cross-contract call, so should we make that more obvious?

Is it across contact call? It's not always a cross contract call.

PR #360, opened 2022-07-28
rename crates to soroban

graydon:
A big search and replace.

(Also note @leighmcculloch I very slightly updated README.md to not mention the stellar-core command-line interface to invoking contracts, I don't think it's useful at this point and will probably just make people sad if they try to build or use it)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ version = "0.0.0"
 dependencies = [
  "example_add_i32",
  "proptest",
- "stellar-contract-sdk",
+ "soroban-sdk",
 ]
 
 [[package]]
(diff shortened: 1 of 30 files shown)


graydon:
Note: this will need an additional step once the repos are renamed to adjust the URLs in `Cargo.toml`, I left commented-out versions of what they should be.

leighmcculloch:
>(Also note @leighmcculloch I very slightly updated README.md to not mention the stellar-core command-line interface to invoking contracts, I don't think it's useful at this point and will probably just make people sad if they try to build or use it)

+1 We don't need core instructions in the SDK. I'll replace the readme with links to docs, etc.

PR #361, opened 2022-07-28
Add Env docs

leighmcculloch:


sdk/src/env.rs
```diff
@@ -238,6 +283,27 @@ impl Env {
...
+    ///     pub fn hello(env: Env, recipient: stellar_contract_sdk::Symbol) -> stellar_contract_sdk::Symbol {
+    ///         todo!()
+    ///     }
+    /// }
+    ///
+    /// # fn main() {
+    /// let env = Env::default();
+    /// let contract_id = Binary::from_array(&env, [0; 32]);
+    /// env.register_contract(contract_id.clone(), HelloContract);
+    /// # }
```
leighmcculloch, line 305: fyi @jonjove @graydon this appears to work for examples in docs that use the macros. Providing the main fn is required.

PR #368, opened 2022-07-28
Use FixedBinary for contact IDs

leighmcculloch:
### What
Use FixedBinary<32> for contract IDs.

### Why
It's annoying when writing contract tests and not realizing that the contract ID has to be a specific length. This tripped me up when writing my first contracts, and then dealing with panics. 

### Known Limitation
@jonjove has pointed out that use of FixedBinary is in general questionable because it adds one additional branch to any code that converts from another type into the FixedBinary. This is because of the length check that occurs. We should try and make FixedBinary as performant as Binary, or remove it in the future all together. Follow up issue for addressing this issue is https://github.com/stellar/rs-soroban-sdk/issues/369.

### DIFF
--- a/sdk/src/env.rs
+++ b/sdk/src/env.rs
@@ -195,7 +195,11 @@ impl Env {
     }
 
     #[doc(hidden)]
-    pub fn create_contract_from_contract(&self, contract: FixedBinary<32>, salt: Binary) -> FixedBinary<32> {
+    pub fn create_contract_from_contract(
+        &self,
+        contract: FixedBinary<32>,
+        salt: Binary,
+    ) -> FixedBinary<32> {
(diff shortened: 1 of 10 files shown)


sdk/src/env.rs
```diff
@@ -197,7 +195,11 @@ impl Env {
     }
 
     #[doc(hidden)]
-    pub fn create_contract_from_contract(&self, contract: Binary, salt: Binary) -> FixedBinary<32> {
+    pub fn create_contract_from_contract(
+        &self,
+        contract: FixedBinary<32>,
+        salt: Binary,
```
jonjove, line 201: `salt` is also 32 bytes.

leighmcculloch: Fixed, thanks for finding that.

Issue #383, opened 2022-08-01
Section contractenvmetav0 not always emitted depending on profile


jonjove:
It is emitted when building target. This is unchanged if I remove the calls to `wasm-opt`.

leighmcculloch:
After reading https://github.com/rust-lang/rust/issues/56639, I now believe my summary above about nightly vs stable is incorrect.

The comment here is likely the issue:
>Long story short a custom section always goes into the current codegen unit, but if nothing else is referenced from the codegen unit then the custom section doesn't get included in the final output because the linker never looks at the codegen unit/object file.

Ref: https://github.com/rust-lang/rust/issues/56639#issuecomment-445924692

In the release profile we recommend, a single codegen unit is used:
```toml
codegen-units = 1
```
This means the custom section is preserved because the unit is always passed to the linker.

Given the custom section is never referenced right now, for anyone not using a single codegen unit it will be a bit random for if this issue occurs or not, because it really just depends how the Rust build system divides logic into codegen units.

There is a common in the linked issue about in what cases the custom section will be included:

> In which crate is the custom section defined? | Where is the function containing the section? | How is the function containing the custom section called? | Is it generated?
> -- | -- | -- | --
> External crate | Submodule | Not called | No
> External crate | Submodule | Indirectly | No
> External crate | Submodule | Directly | Yes
> External crate | In root | Not called | No
> External crate | In root | Indirectly | Yes
> External crate | In root | Directly | Yes
> Main crate | Any | Any | Yes

Ref: https://github.com/rust-lang/rust/issues/56639#issue-388956758

jonjove:
Reopening this for the moment based on https://github.com/stellar/rs-soroban-sdk/pull/386#issuecomment-1201610591.

leighmcculloch:
@jonjove It looks like the token contract is still using an old version of soroban-sdk-macros, so it doesn't have the fix that was committed in #386. However, even with the macros crate updated in the token contract it doesn't make a difference, it is still missing.

PR #386, opened 2022-08-01
Fix contractmetav0 custom section not emitted

leighmcculloch:
### What
Move contractmetav0 custom section link section static variable into function that is exported instead of being a crate root static, and reference that exported function in the crates of contracts.

### Why
The custom section is not always in contracts when contracts are built with `codegen-units` greater than 1. Because the custom section is not in the contract crate itself, it could end up in an object file for the SDK that doesn't end up getting linked. When `codegen-units` is 1 this isn't an issue because there should only be one.

We recommend that `codegen-units` should always be 1 to improve the optimization of the contract builds, but this is an annoying footgun for anyone who misses that.

By moving the link section into an exported function that gets referenced in the contract crate, the object file should always be included in linking. We reference it in every contract function because there is no way to just do it once for all functions since macros have no contract state.

I tested this change with the examples to make sure that it didn't increase the size of the binary. No binaries changed.

Close https://github.com/stellar/rs-soroban-sdk/issues/383

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -802,7 +802,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-sdk"
-version = "0.0.2"
+version = "0.0.3"
 dependencies = [
  "ed25519-dalek",
  "hex",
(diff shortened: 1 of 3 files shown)


jonjove:
@leighmcculloch as far as I can tell, this doesn't actually cause the section to be generated on the token contract (when codegen-units > 1). Am I doing something wrong?

PR #394, opened 2022-08-02
Rename Binary=>Bytes, FixedBinary=>BytesN

### LINKED ISSUE #385: Rename Binary to Bytes
I think we should consider renaming the `Binary` type to `Bytes`.

In Solidity, their binary type uses the terms byte. https://docs.soliditylang.org/en/v0.8.11/types.html#fixed-size-byte-arrays

In Vyper, the new Ethereum lang, their binary type is named Bytes. https://vyper-by-example.org/values/

The term is immediately recognizable and less ambiguous as Binary. Binary can be ambiguous because people more often use the term to refer to executable binaries than they do byte slices. The term byte almost always shows up in any description I've casually given to what a Binary is.

Thoughts @jonjove @sisuresh @paulbellamy @tomerweller ?

leighmcculloch:
### What
Rename Binary=>Bytes and FixedBinary=>BytesN.

### Why
Binary is a somewhat confusing term. When I see it in the docs it doesn't scream at me that this is the type for storing byte slices or byte arrays. Doing a quick survey of Ethereum, they use the term Bytes in both Solidity and Vyper. Thinking back to whenever I have explained Binary to others or in my head, I immediately reach for the term "bytes" or "byte slice".

This change helps others gracefully update their code because is still contains the Binary and FixedBinary types, with deprecated notices to encourage folks to switch to the new types.

Close #385


jonjove:
Skimmed it and it looks reasonable enough. Should the spec type names change too?

Issue #408, opened 2022-08-09
Revisit pattern for defining if contractimpl is exported


leighmcculloch:
Today we specify if a contractimpl is exported with the `export_if` attribute which accepts a feature. This requires the contract developer to proactively add a feature to their crate that is used to gate whether the contract impl is exported. If the attribute isn't provided it is assumed to always be exported.

I think we need to revisit this pattern. The pattern is simple to implement for us, and flexible for contract developers, but it has one undesirable quality for contract developers:

It is opt-in. Developers who don't set an `export_if` always export, which means that their crate is likely never importable by other contracts. Anyone importing their crate will possibly unknowingly re-export their exports, and there is little that the importer can do about it.

Ideally the solution we use for controlling whether the functions are exported is not opt-in, and by default the `contractimpl` macro sets a contract such that it is importable safely, but when built for itself is always exported.

Ideally the solution would involve automatically detecting if a crate is being built as a dependency or as the primary package.

It's unclear if there is a better solution than what we have now given constraints of rust/cargo, but I think it's worth us having a look to see.

leighmcculloch:
I thought maybe we could use a `cfg` on `crate_type` as a way to conditionally compile in the exports so they are only exported for `cdylib` and not `rlib`, but that is currently not possible with cargo:
- https://github.com/rust-lang/rust/issues/45182
- https://github.com/rust-lang/rust/issues/20267.

leighmcculloch:
Going to close this in favor of #409.

PR #409, opened 2022-08-09
Shift export feature out of contractimpl

leighmcculloch:
### What
Shift export feature out of contractimpl and make it a boolean parameter that the developer can control using features or however they like.

### Why
A little experimentation with seeing how we can remove the export feature from the contractimpl macro.

### DIFF
--- a/tests/udt/src/lib.rs
+++ b/tests/udt/src/lib.rs
@@ -19,7 +19,7 @@ pub struct UdtStruct {
 pub struct Contract;
 
 #[cfg_attr(feature = "export", contractimpl(export = true))]
-#[cfg_attr(not(feature = "export"), contractimpl(export = false)]
+#[cfg_attr(not(feature = "export"), contractimpl(export = false))]
 impl Contract {
     pub fn add(a: UdtEnum, b: UdtEnum) -> i64 {
         let a = match a {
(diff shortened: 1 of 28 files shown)


tests/udt/src/lib.rs
```diff
@@ -18,7 +18,8 @@ pub struct UdtStruct {
 
 pub struct Contract;
 
-#[contractimpl(export_if = "export")]
+#[cfg_attr(feature = "export", contractimpl(export = true))]
+#[cfg_attr(not(feature = "export"), contractimpl)]
```
leighmcculloch, line 22: Fyi @jonjove I took a quick stab of an alternative approach that shifts the conditional decision on whether to export out of the contractimpl and into user code. It's pretty verbose though. This doesn't actually address any of the issues from #408, it was just an experiment in an alternative way to provide the feature selection.

jonjove:
This is an interesting design and I'm glad you did this experiment.

What I like: Much more flexible, can avoid the use of an "export" feature if it's not needed
What I don't like: Quite a bit more verbose if you do want it feature based

leighmcculloch:
> What I like: Much more flexible, can avoid the use of an "export" feature if it's not needed

There is the practical fact that it is more flexible, but the main driver for me with this is that it conceptually looks more like Rust code I see about, and therefore even though it is more verbose, it is instantly understandable that I can control exports with features, as opposed to, "if I pass a string, that will end up feature gating exports."

I'm leaning to merging this. Lmk if you disagree strongly.

PR #411, opened 2022-08-10
Add macros contractimport, contractclient, contractfile

### LINKED ISSUE #118: Contract Spec: Add macros that generates a cross-contract calling impl
Given a wasm file path, or a contract ID and a Horizon URL, generate contract calling code and testing capabilities.

leighmcculloch:
### What
Add the `contractimport` macro for importing contracts via their wasm into a contract to use them for:
- deploying them
- invoking them
- testing against them

Add the `contractclient` macro for generating a client that can:
- invoke contract functions for a contract defined by a trait

Add the `contractfile` macro for including the WASM bytes for a file into code, using the same file relative loading that contractimport uses.

### Why
Today building against other existing contracts is complex and dependent on some assumptions, because we require importing their rust code which might not be available. Depending on the rust code requires contracts to have all this internal machinery like features to control what is exported, what isn't, and this makes the developer experience developing contracts noisy. Also, it introduces multiple ways for contracts to interact natively and via wasm invokes, which is not ideal since testing is frequently occurring with the approach not used on chain.

This smooths over the story for how to embed and interact with other contracts.

You might notice that `contractfile` doesn't do a whole lot when compared to the built in `include_bytes!`. All it does extra is sha256 hash verification. The main reason we need `contractfile` is because Rust doesn't provide a way in stable to get the relative location of a proc-macro invocation, and so there's no way to replicate the relative file loading behavior of `include_bytes`. `contractimport` needs to load the file into memory inside the proc-macro, so it has to use a workspace relative path, and we need the way the binary gets loaded into variable to use the same path otherwise the developer would have to enter two paths. One day if https://github.com/rust-lang/rust/issues/54725 is stabalized we can replicate the `include_bytes!` relative path in `contractimport`. Either way we add sha256 verification so we do get some benefit of it.

Close #118

### Status
This is a work-in-progress.


leighmcculloch:
We appear to have an issue with .wasm files being built differently in different situations.

CI builds using the following commands, approximately:
```
cargo clean && cargo hack build --target wasm32-unknown-unknown --profile release && cargo test --no-default-features -p soroban-sdk
```
Which result in a hash of `target/wasm32-unknown-unknown/release/example_add_i32.wasm`:
```
0ef130ce81f57cb03a9e20b0e8b5c9e47bf8c50cf0a811e4b25535ad95b92a7e
```

When building and testing with `make build test`:
```
58f98f171987ffff001c4c7f15357bb2c2a4870ec9294c28a401d6c0143ed0e5
```

leighmcculloch:
🤦🏻 Of course, the Makefile runs the files through wasm-opt.

PR #426, opened 2022-08-11
Fix up the add_bigint example build

paulbellamy:
### What

Fix the build for tests/add_bigint

### Why

So it gets built with the other examples

### Known limitations

[TODO or N/A]

### DIFF
--- a/tests/add_bigint/Cargo.toml
+++ b/tests/add_bigint/Cargo.toml
@@ -4,9 +4,19 @@ version = "0.0.0"
 authors = ["Stellar Development Foundation <info@stellar.org>"]
 license = "Apache-2.0"
 edition = "2021"
+publish = false
+rust-version = "1.62"
 
 [lib]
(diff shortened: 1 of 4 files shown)


tests/add_bigint/Cargo.toml
```diff
@@ -4,9 +4,19 @@ version = "0.0.0"
...
 crate-type = ["cdylib", "rlib"]
+doctest = false
 
 [dependencies]
-stellar-contract-sdk = {path = "../../soroban-sdk"}
+soroban-sdk = {path = "../../soroban-sdk"}
+
+[dev-dependencies]
+proptest = "1.0.0"
+soroban-sdk = {path = "../../soroban-sdk", features = ["testutils"]}
```
leighmcculloch, line 19: You can remove this line, the self reference dev dep is no longer required.

leighmcculloch: I was wrong about this, this is fine.

PR #454, opened 2022-08-18
Rename Env::create_contract_from_contract to Env::deploy_contract

leighmcculloch:
### What
Rename `Env::create_contract_from_contract` to `Env::deploy_contract`.

Rename `Env::create_token_from_contract` to `Env::deploy_token_contract`.

### Why
The distinction between creating a contract from an ed25519 or a contract is not particularly relevant within the SDK. Contracts can only create contracts from contracts. To put it in the developers words, a contract can deploy contracts. We use the term deploy in other places that the developer sees, like the CLI. We can use that term here.

There could be an argument that we also need a function for creating from ed25519 for tests, but I think that is unclear. And, if we do add that later, I think that is still doable and these function names still stand well with a change like that.

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -188,16 +188,26 @@ impl Env {
             .unwrap()
     }
 
-    #[doc(hidden)]
-    pub fn create_contract_from_contract(&self, contract: Bytes, salt: BytesN<32>) -> BytesN<32> {
+    /// Deploy contract takes the given contract bytes and stores it on ledger
+    /// with a new contract ID that is generated using a combination of the
+    /// currently executing contract ID and the provided salt.
+    ///
+    /// The given contract bytes must be a valid WASM Soroban contract.
+    ///
+    /// Returns the contract ID of the deployed contract.
+    pub fn deploy_contract(&self, contract: Bytes, salt: BytesN<32>) -> BytesN<32> {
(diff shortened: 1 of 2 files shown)


leighmcculloch:
+1

> This PR also suggests that we probably aren't doing a good enough job explaining how existing functionality can be used.

I think it goes deeper than a missing explanation. The way that we are framing contract creation irrespective of documentation is not intuitive.

For example, the terms `_from_contract` and `_from_ed25519` are confusing. What does it mean to be "from" one of these things. Even when using an ed25519 public key to create, you point out that we can create "with ed25519 from a contract," so the "from contract" is now even more confusing.

The term `salt` is also ambiguous. While I agree we are technically using the term correctly, it isn't intuitive, and its name doesn't communicate important properties of its value, such as its uniqueness in the scope of the contract or ed25519.

I think we need to find some better term to define the "identifier", or "sub-identifier", which is currently the "salt". And a better term to define the "scope" of uniqueness that the identifier applies within, which is either a contract or an ed25519.

cc @sisuresh

jonjove:
@leighmcculloch I totally agree with what you wrote. Let's think all of this through and put it back together in a more sane way.

leighmcculloch:
Moving conversation to https://github.com/stellar/rs-soroban-env/issues/348 and closing this PR in favor of discussing the problem more completely before any change.

PR #459, opened 2022-08-19
Compile error when contracttype struct field name too long

leighmcculloch:
### What
Add compile error when contracttype struct field name too long.

### Why
Compile errors will render in a user friendly way to the developer, and can highlight the area of the code that is the problem.


sisuresh (approved):
LGTM. Thanks!

Issue #427, opened 2022-08-11
Unit variants of enums should convert to single element vector


jonjove:
### What problem does your feature solve?

```rust
enum Example {
    V0,
    V1(u32),
    V2(RawVal),
}
```
maps as
```
Example::V0 -> ["V0", Void]
Example::V1(3) -> ["V1", 3]
```

This design doesn't naturally extend to enums with longer tuple variants. It's also ambiguous for `V2` because
```
Example::V2(Void) -> ["V2", Void]
```
could mean variant V2 as unit variant or variant V2 with the associated `RawVal` void.

### What would you like to see?

It should map as
```
Example::V0 -> ["V0"]
Example::V1(3) -> ["V1", 3]
```

### What alternatives are there?

TODO

PR #461, opened 2022-08-22
Convert unit enums to single element vector

leighmcculloch:
### What

Convert unit enums to single element vectors and ignore extra fields when decoding.

### Why

For a number of reasons:
- Encode unit enums into their most efficient form.
- To make it easier to scale enum encoding to support tuple variants with more than a single value.
- To make it easier to make changes to enums by adding tuple values.

Close https://github.com/stellar/rs-soroban-sdk/issues/427

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,7 +831,6 @@ dependencies = [
 [[package]]
 name = "soroban-env-common"
 version = "0.0.3"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=a9c1396a#a9c1396ac7936da22f447b1d1fb5b737a7ae1262"
 dependencies = [
  "soroban-env-macros",
  "soroban-wasmi",
@@ -842,7 +841,6 @@ dependencies = [
 [[package]]
 name = "soroban-env-guest"
 version = "0.0.3"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=a9c1396a#a9c1396ac7936da22f447b1d1fb5b737a7ae1262"
 dependencies = [
  "soroban-env-common",
  "static_assertions",
@@ -851,7 +849,6 @@ dependencies = [
 [[package]]
 name = "soroban-env-host"
 version = "0.0.3"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=a9c1396a#a9c1396ac7936da22f447b1d1fb5b737a7ae1262"
 dependencies = [
  "backtrace",
  "dyn-fmt",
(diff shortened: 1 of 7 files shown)


leighmcculloch:
> * To make it easier to make changes to enums by adding tuple values.

@jonjove I'm thinking of making the same changes to tuples, so a 4 length tuple can be decoded into a 3 element tuple. I think the downsides are minimal. I'm typically opposed to non-symmetric encoding of types, such as u64 -> RawVal -> i64 which can create situations that won't work, but I don't think tuples will experience the same footguns. Lmk what you think.

jonjove:
> I'm thinking of making the same changes to tuples, so a 4 length tuple can be decoded into a 3 element tuple
I want to make sure I understand. Instead of checking that the `Vec` has the exact length of the target tuple, you just want to check that the `Vec` is at least long enough?

jonjove:
I see why one might want this, but I feel like it's more likely to be a foot-gun. In general, I don't think you should retroactively modify your enum variants--you should just add new ones.

leighmcculloch:
It's already possible to retroactively modify structs by adding new keys, so this makes enums more like structs in that regard.

jonjove:
Maybe we should be going the other direction and making it impossible to retroactively modify structs? That also seems likely to be a foot-gun.

leighmcculloch:
To do that for structs would require additional wasm code that doesn't otherwise serve the user. We'd be adding an additional length check in the decoding code doesn't need. I feel like this is the sort of thing we can tell people not to do, but then they'll go out of their way to do it anyway, so I'm not sure it is worth it vs just providing best practices on how to do it safely, which is no different to general database migration best practices.

leighmcculloch:
The reason I think u64 -> RawVal -> i64 is too much of a footgun is because there are ranges of u64 that this will work for, and ranges it won't work for, and code may work fine in tests if those ranges aren't discovered in tests.

On the contrary it seems harder for (T0, T1, T2) -> RawVal -> (T0, T1) to be a footgun since the encoding and decoding types are concretely specified in the code at compile time. i.e. Runtime can't change the (T0, T1, T2) into (T0). This is why I don't think this is so bad.

jonjove:
I don't think I agree with you. The really dangerous thing in this case is

1. storage contains `(0u32, 1u32, 2u32)`
2. read from storage into `let mut x: (u32, u32) = e.contract_data().get(key) // (0u32, 1u32)` because code is not aware that the tuple actually had length 3
3. modify tuple element 0 `x.0 += 1`
4. write to storage `e.contract_data().put(key, x) // (1u32, 1u32)` which means code accidentally truncated storage, losing data

In this case, the code looks like I modified an element of a tuple. But I actually modified the structure of the tuple too. Data was lost, and future code might panic if it depends on that tuple having length 3.

leighmcculloch:
Ah good point, and nice example. Indeed, the footgun remains.

leighmcculloch:
Follow change is in https://github.com/stellar/rs-soroban-sdk/issues/462

PR #464, opened 2022-08-22
Expose host functions for event publishing

### LINKED ISSUE #451: Add `Env` functions for publishing events via host functions

jayz22:
### What

Resolves #451.

### Why

### Known limitations

[TODO or N/A]


leighmcculloch (approved):

soroban-sdk/src/contract_event.rs
```diff
@@ -0,0 +1,66 @@
+use core::fmt::Debug;
+
+use crate::{env::internal, Env, IntoVal, RawVal, Vec};
+
+/// ### Examples
```
leighmcculloch, line 5: Could you add some lead in text describing the type?

soroban-sdk/src/contract_event.rs
```diff
@@ -0,0 +1,66 @@
...
+/// #     let env = Env::default();
+/// #     let contract_id = FixedBinary::from_array(&env, [0; 32]);
+/// #     env.register_contract(&contract_id, Contract);
+/// #     f::invoke(&env, &contract_id);
+/// # }
+/// # #[cfg(not(feature = "testutils"))]
+/// # fn main() { }
+/// ```
+#[derive(Clone)]
+pub struct ContractEvent(Env);
```
leighmcculloch, line 35: I think it might be worth naming this `ContractEvents`, or maybe even just `Events`, since it doesn't represent a single event and instead represents contract events the functionality. I don't feel strongly.

leighmcculloch (approved):
Had a request and thought, but either way this is good to merge and if we decide to make those changes they can be follow up.

PR #465, opened 2022-08-22
Add Set type

### LINKED ISSUE #228: Type: Set: Add type
Add a `Set` type to the SDK, that has similar interface to the `Vec` type, but is backed by a `Map` where the elements in the set are keys in the map.

accordeiro:
Fixes #228 – Add a Set type to the SDK. This `Set` implementation is basically syntax sugar on top of the `Map` type, where all `Set` entries are `Map` keys (so effectively each entry is unique).

This is my first code ever written in Rust, so all feedback is welcome! :)


soroban-sdk/src/set.rs
```diff
@@ -0,0 +1,276 @@
...
+macro_rules! set {
+    ($env:expr) => {
+        $crate::Set::new($env)
+    };
+    ($env:expr, $($x:expr),+ $(,)?) => {
+        $crate::Set::from_array($env, [$($x), +])
+    };
+}
+
+/// Set is a contiguous growable array type composed of unique items.
```
leighmcculloch, line 29: It's not contiguous.
```suggestion
/// Set is an ordered growable collection composed of unique items.
```

soroban-sdk/src/set.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<T> Set<T>
+where
+    T: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal>,
+{
+    pub(crate) fn env(&self) -> &Env {
+        self.0.env()
+    }
+
+    pub fn new(env: &Env) -> Set<T> {
+        let map = Map::<T, ()>::new(env);
```
leighmcculloch, line 67: Fyi the type annotations shouldn't be required since they are derived from the destination.
```suggestion
        let map = Map::new(env);
```

leighmcculloch (approved):
This looks great! Thanks!!!

There are a few things to follow up with:
- Add an iterator for the type.
- Add conversion functions for the Set itself so it can be stored in itself and inside other collection types. (Hold off on doing this because I'm currently reworking conversions.)
- Add some functions like add, subtract, etc, that are common on sets. These can probably be added as operators.

soroban-sdk/src/set.rs
accordeiro: I went back and forth on whether I should use the `ordered` word here. When I think of something **ordered**, I think that "the order in which I add items matters". And I would think `set(1, 2, 3) != set(2, 1, 3)`. 

I tried to be more specific about this implicit ordering a few lines below in the doc:

> Set imposes a fixed ordering, so Set<1, 2, 3> == Set<2, 3, 1>

What do you think of removing the `ordered` qualifier here and just say:

```suggestion
/// Set is a growable collection composed of unique items.
```
?

leighmcculloch: Good point, it's so ambiguous to what it actually means. +1 to removing the word ordered here.

PR #467, opened 2022-08-23
Remove need for Vec<EnvVal> on events

leighmcculloch:
### What
Change topic list from Vec<EnvVal> to tuple. Also add some docs.

### Why
To remove one case of surfacing EnvVal in the SDK API.


jayz22:
@leighmcculloch Thank you for making this follow up change. I was also working up a solution yesterday but didn't finish before the end of day: https://github.com/stellar/rs-soroban-sdk/pull/469/files. It is pretty similar to what you have, except I made a trait and impl'ed on those tuples, such that you don't need the `publish_n` functions. I also added a couple of validity checks on each of the topic. 
I'm good with either approach. So feel free to keep either one. I've marked the other one in draft as I still have a thing to fix and it depends on some light change on the env crate.

leighmcculloch:
Closing in favor of #469.

PR #469, opened 2022-08-23
Replace `Vec<RawVal>` with tuples of variable sizes in events publishing.

jayz22:
### What
Follows up to https://github.com/stellar/rs-soroban-sdk/pull/464 which resolves https://github.com/stellar/rs-soroban-sdk/issues/451
- Replace `Vec<RawVal>` with tuples of 0-4 sizes. 
- Address a couple comments in the previous PR. 

Note: this is duplicate work with https://github.com/stellar/rs-soroban-sdk/pull/467 with slightly different approach. Only one needs to be merged.  

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-sdk/src/contract_event.rs
+++ b/soroban-sdk/src/contract_event.rs
@@ -1,13 +1,21 @@
 use core::fmt::Debug;
 
-use crate::{env::internal, Env, IntoVal, RawVal, Vec};
+use super::xdr::ScObjectType;
+
+// TODO: consolidate with host::events::TOPIC_BYTES_LENGTH_LIMIT
+const TOPIC_BYTES_LENGTH_LIMIT: u32 = 32;
+
+use crate::{
+    env::{internal, EnvObj},
+    Bytes, Env, IntoVal, Object, RawVal, Vec,
+};
 
 /// ### Examples
 ///
 /// ```
 /// use soroban_sdk::Env;
 ///
-/// # use soroban_sdk::{contractimpl, vec, map, RawVal, FixedBinary};
+/// # use soroban_sdk::{contractimpl, vec, map, RawVal, BytesN};
 /// #
 /// # pub struct Contract;
 /// #
@@ -15,7 +23,7 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
 /// # impl Contract {
 /// #     pub fn f(env: Env) {
 /// let event = env.contract_event();
-/// let topics = vec![&env, 0u32, 1u32];
+/// let topics = (0u32, 1u32);
 /// let data = map![&env, (1u32, 2u32)];
 /// event.publish(topics, data)
 /// #     }
@@ -24,7 +32,7 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
 /// # #[cfg(feature = "testutils")]
 /// # fn main() {
 /// #     let env = Env::default();
-/// #     let contract_id = FixedBinary::from_array(&env, [0; 32]);
+/// #     let contract_id = BytesN::from_array(&env, [0; 32]);
 /// #     env.register_contract(&contract_id, Contract);
 /// #     f::invoke(&env, &contract_id);
 /// # }
@@ -32,35 +40,83 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
 /// # fn main() { }
 /// ```
 #[derive(Clone)]
-pub struct ContractEvent(Env);
+pub struct Events(Env);
 
-impl Debug for ContractEvent {
+impl Debug for Events {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "ContractEvent")
+        write!(f, "Events")
     }
 }
 
-impl ContractEvent {
+pub trait IntoTopics {
+    fn into_topics(self, env: &Env) -> Object;
+}
+
+// 0 topics
+impl IntoTopics for () {
+    fn into_topics(self, env: &Env) -> Object {
+        Vec::<RawVal>::new(env).to_object()
+    }
+}
+
+macro_rules! impl_for_tuple {
+    ( $($typ:ident $idx:tt)* ) => {
+        impl<$($typ),*> IntoTopics for ($($typ,)*)
+        where
+            $($typ: IntoVal<Env, RawVal> + Clone),*
+        {
+            fn into_topics(self, env: &Env) -> Object {
+                $({
+                    let ev = self.$idx.clone().into_env_val(env);
+                    match TryInto::<EnvObj>::try_into(ev) {
+                        Ok(obj) => {
+                            if obj.as_object().is_obj_type(ScObjectType::Vec) {
+                                panic!("topic cannot be a vec")
+                            } else if obj.as_object().is_obj_type(ScObjectType::Map) {
+                                panic!("topic cannot be a map")
+                            } else if obj.as_object().is_obj_type(ScObjectType::Bytes) {
+                                let bytes = unsafe { Bytes::unchecked_new(obj.clone()) };
+                                if bytes.len() > TOPIC_BYTES_LENGTH_LIMIT {
+                                    panic!("topic exceeds bytes length limit")
+                                }
+                            }
+                        }
+                        Err(_) => (),
+                    }
+                })*
+                self.into_val(env)
+            }
+        }
+    };
+}
+
+// 1-4 topics
+impl_for_tuple! { T0 0 }
+impl_for_tuple! { T0 0 T1 1 }
+impl_for_tuple! { T0 0 T1 1 T2 2 }
+impl_for_tuple! { T0 0 T1 1 T2 2 T3 3 }
+
+impl Events {
     #[inline(always)]
     pub(crate) fn env(&self) -> &Env {
         &self.0
     }
 
     #[inline(always)]
-    pub(crate) fn new(env: &Env) -> ContractEvent {
-        ContractEvent(env.clone())
+    pub(crate) fn new(env: &Env) -> Events {
+        Events(env.clone())
     }
 
     /// Publishes `topics` and `data` into a contract event.
     /// `topics` is expected to have length <= 4 and cannot contain Vecs, Maps,
     /// or Binaries > 32 bytes
     #[inline(always)]
-    pub fn publish<T, D>(&self, topics: Vec<T>, data: D)
+    pub fn publish<T, D>(&self, topics: T, data: D)
     where
-        T: IntoVal<Env, RawVal>,
+        T: IntoTopics,
         D: IntoVal<Env, RawVal>,
     {
         let env = self.env();
-        internal::Env::contract_event(env, topics.to_object(), data.into_val(env));
+        internal::Env::contract_event(env, topics.into_topics(env), data.into_val(env));
     }
 }
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -45,7 +45,7 @@ pub type EnvType<V> = internal::EnvVal<Env, V>;
 pub type EnvVal = internal::EnvVal<Env, RawVal>;
 pub type EnvObj = internal::EnvVal<Env, Object>;
 
-use crate::ContractEvent;
+use crate::Events;
 use crate::{Bytes, BytesN, ContractData, Ledger};
 
 /// The [Env] type provides access to the environment the contract is executing
@@ -114,11 +114,11 @@ impl Env {
         Ledger::new(self)
     }
 
-    /// Get a [ContractEvent] for publishing events associated with the
+    /// Get a [Events] for publishing events associated with the
     /// currently executing contract.
     #[inline(always)]
-    pub fn contract_event(&self) -> ContractEvent {
-        ContractEvent::new(self)
+    pub fn contract_event(&self) -> Events {
+        Events::new(self)
     }
 
     /// Get the 32-byte hash identifier of the current executing contract.
--- a/soroban-sdk/src/lib.rs
+++ b/soroban-sdk/src/lib.rs
@@ -77,7 +77,7 @@ pub use bigint::BigInt;
 pub use bytes::{Binary, FixedBinary};
 pub use bytes::{Bytes, BytesN};
 pub use contract_data::ContractData;
-pub use contract_event::ContractEvent;
+pub use contract_event::Events;
 pub use ledger::Ledger;
 pub use map::Map;
 pub use set::Set;


soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
+
+// 0 topics
+impl IntoTopics for () {
+    fn into_topics(self, env: &Env) -> Object {
+        Vec::<RawVal>::new(env).to_object()
     }
 }
 
-impl ContractEvent {
+macro_rules! impl_for_tuple {
```
jonjove, line 62: Nitpick: can you give this a name that indicates what is being `impl`ed?

I also think you should evaluate the impact of this approach on size of WASM binaries. One different approach that might or might not be more size efficient would be to write a private function that takes `Vec<EnvVal>` (or something like it) and contains all the logic then just provide public tuple wrappers to that function.

jayz22: Yeah, this is a great point! I will definitely check the sizes. This might also have a larger size than @leighmcculloch 's [approach](https://github.com/stellar/rs-soroban-sdk/pull/467) since it has extra topic validation logic.

leighmcculloch: I think we can keep the validation logic in the host side. If the WASM side is just going to panic, the host can do that.

soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
-impl Debug for ContractEvent {
+impl Debug for Events {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "ContractEvent")
+        write!(f, "Events")
+    }
+}
+
+pub trait IntoTopics {
+    fn into_topics(self, env: &Env) -> Object;
```
leighmcculloch, line 52: Could you add docs to the type and the into_topics function?
```suggestion
pub trait IntoTopics {
    fn into_topics(self, env: &Env) -> Vec<RawVal>;
```

soroban-sdk/src/contract_event.rs
```diff
@@ -1,21 +1,29 @@
...
+
+// TODO: consolidate with host::events::TOPIC_BYTES_LENGTH_LIMIT
+const TOPIC_BYTES_LENGTH_LIMIT: u32 = 32;
+
+use crate::{
+    env::{internal, EnvObj},
+    Bytes, Env, IntoVal, Object, RawVal, Vec,
+};
 
 /// ### Examples
```
leighmcculloch, line 13:
```suggestion
/// Events publishes events for the currently executing contract.
 ///
/// ### Examples
```

soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
+                            } else if obj.as_object().is_obj_type(ScObjectType::Map) {
+                                panic!("topic cannot be a map")
+                            } else if obj.as_object().is_obj_type(ScObjectType::Bytes) {
+                                let bytes = unsafe { Bytes::unchecked_new(obj.clone()) };
+                                if bytes.len() > TOPIC_BYTES_LENGTH_LIMIT {
+                                    panic!("topic exceeds bytes length limit")
+                                }
+                            }
+                        }
+                        Err(_) => (),
```
leighmcculloch, line 84: The validations here should happen host side. There is no error handling on the WASM side so I don't think we gain much by doing the checks here.

soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
 
     /// Publishes `topics` and `data` into a contract event.
     /// `topics` is expected to have length <= 4 and cannot contain Vecs, Maps,
     /// or Binaries > 32 bytes
     #[inline(always)]
-    pub fn publish<T, D>(&self, topics: Vec<T>, data: D)
+    pub fn publish<T, D>(&self, topics: T, data: D)
     where
-        T: IntoVal<Env, RawVal>,
+        T: IntoTopics,
```
leighmcculloch, line 116: Nice! I like this over what I suggested in #467. My only concern is if this doesn't render as clearly in the docs, but as long as we provide good examples of how you can pass in different tuple sizes I think that's fine. Can you add examples?

soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
     #[inline(always)]
-    pub(crate) fn new(env: &Env) -> ContractEvent {
-        ContractEvent(env.clone())
+    pub(crate) fn new(env: &Env) -> Events {
+        Events(env.clone())
     }
 
     /// Publishes `topics` and `data` into a contract event.
     /// `topics` is expected to have length <= 4 and cannot contain Vecs, Maps,
     /// or Binaries > 32 bytes
```
leighmcculloch, line 112: Keep the first line of the docs short and brief because it is rendered in places in the docs separate to the rest.

```suggestion
     /// Publish an event.
     ///
     /// Event data is specified in `data`. Data may be any value or
     /// type, including types defined by contracts using [contracttype].
```

leighmcculloch (commented):
I closed #467 in favor of this.

soroban-sdk/src/env.rs
```diff
@@ -114,11 +114,11 @@ impl Env {
         Ledger::new(self)
     }
 
-    /// Get a [ContractEvent] for publishing events associated with the
+    /// Get a [Events] for publishing events associated with the
```
leighmcculloch, line 117:
```suggestion
    /// Get [Events] for publishing events associated with the
```

soroban-sdk/src/contract_event.rs
leighmcculloch: For the above doc comment you'll need to also include this in the code:
```rust
 #[cfg(doc)]
 use crate::{contracttype, Bytes, BytesN, Map};
```

soroban-sdk/src/contract_event.rs
leighmcculloch, line 52: I think IntoTopics could probably just be a marker trait, it doesn't actually need a function or any functionality.


i.e.
```suggestion
pub trait Topics {
```

Then the macro impl for the tuples is just doing:

```rust
impl<T0, T1, ...> Topics for (T0, T1, ...) {}
```

And the publish function can be defined as:

```rust
     pub fn publish<T, D>(&self, topics: T, data: D)
     where
         T: Topics + IntoVal<Env, RawVal>,
```

soroban-sdk/src/contract_event.rs
```diff
@@ -24,43 +32,91 @@ use crate::{env::internal, Env, IntoVal, RawVal, Vec};
...
+        write!(f, "Events")
+    }
+}
+
+pub trait IntoTopics {
+    fn into_topics(self, env: &Env) -> Object;
+}
+
+// 0 topics
+impl IntoTopics for () {
```
leighmcculloch, line 56: We should `impl<T> Topics for Vec<T>` too, so that someone can use a Vec as a topic list too.

soroban-sdk/src/env.rs
```diff
@@ -114,11 +114,11 @@ impl Env {
...
     /// currently executing contract.
     #[inline(always)]
-    pub fn contract_event(&self) -> ContractEvent {
-        ContractEvent::new(self)
+    pub fn contract_event(&self) -> Events {
```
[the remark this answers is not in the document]
jayz22: Would you prefer me to rename `contract_event.rs` to `events.rs` as well?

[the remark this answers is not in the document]
leighmcculloch: It's probably a good idea. The diff might be easier to read though if done in a separate PR. Up to you.

soroban-sdk/src/contract_event.rs
jayz22: To clarify, `contracttype` *can* be a topic? It is listed in the "must not contain" category in your comment.

leighmcculloch: No, can't be, because contract types are Vecs and Maps.

soroban-sdk/src/contract_event.rs
jayz22: This makes sense, I've removed the validation logic.

soroban-sdk/src/env.rs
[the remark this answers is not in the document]
jayz22: Done. The renaming of the file is its separate commit (last one) to make it less difficult for diff.

soroban-sdk/src/contract_event.rs
jayz22: I've adopted the marker trait suggestion which removes need for the "into" function.

soroban-sdk/src/contract_event.rs
jayz22: Fixed.

soroban-sdk/src/contract_event.rs
jayz22: Yeah this makes sense, I've removed it.

soroban-sdk/src/contract_event.rs
jayz22: I've added several examples in the `doctest` using tuple of different sizes and a vec to represent the topics. Let me know if this is clear. Happy to add more if needed.

soroban-sdk/src/contract_event.rs
jayz22: Thanks, updated the comment.

soroban-sdk/src/env.rs
jayz22: Fixed.

soroban-sdk/src/contract_event.rs
jayz22: I like this suggestion! I've changed the `IntoTopics` to `Topics: IntoVal<Env, Object>` as you suggested. The `IntoVal` part replaces the need for the previous `into_topics` function.
I had to adapt the `Vec<T>` and `()` to impl the `IntoVal` trait. The `Vec` I think is fine. But `()` I'm a little unsure this is right place where it should go.

soroban-sdk/src/contract_event.rs
jayz22: Done!

soroban-sdk/src/events.rs
```diff
@@ -0,0 +1,118 @@
...
+    ( $($typ:ident $idx:tt)* ) => {
+        impl<$($typ),*> Topics for ($($typ,)*)
+        where
+            $($typ: IntoVal<Env, RawVal>),*
+        {
+        }
+    };
+}
+
+impl IntoVal<Env, Object> for () {
```
jayz22, line 71: Need `impl IntoVal` so that `()` can be converted to an empty `Vec` as topics.

soroban-sdk/src/contract_event.rs
leighmcculloch: 👀

soroban-sdk/src/events.rs
leighmcculloch: This conversion could come back to bite us, because it isn't obvious that a `()` converted to an `Object` should be a `Vec` under the hood. Any conversion to `Object` should consistently convert to `RawVal` as the same value, but here we have:

- `()` => `Object` => `RawVal` = `Vec`
- `()` => `RawVal` => `Static::Void`

`()` should exclusively converts to void when it comes to converting to `RawVal`/`Object`.

Could you replace this with `impl<T> IntoVal<Env, Vec<T>> for ()`?

soroban-sdk/src/events.rs
```diff
@@ -0,0 +1,118 @@
...
+#[derive(Clone)]
+pub struct Events(Env);
+
+impl Debug for Events {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "Events")
+    }
+}
+
+pub trait Topics: IntoVal<Env, Object> {}
```
leighmcculloch, line 59: I think this should be Vec<T>. That's ultimately what we're needing all the values to be. It'd mean adding some functions to tuples I think here in the SDK. Wdyt?
```suggestion
pub trait Topics: IntoVal<Env, Vec<T>> {}
```

jayz22: I'm trying it out but I think there is a difference. 
Ultimately what goes into the host a `Vec`, i.e., heterogeneous vector, which tuple `(T1, T2, T3, T4)` gets converted to. 
But `Vec<T>` implies this must be a homogenous typed vector. This goes back to the issue at the beginning and why this PR started. 
I'll experiment a little bit to see if there's an elegant way, if not might have to fall back to the `IntoTopics` solution.

leighmcculloch: Oh sorry, I should have said `Vec<RawVal>`. I think all these types are convertible into that?

leighmcculloch: The main reason to do this is to support https://github.com/stellar/rs-soroban-sdk/pull/469/files#r953184081.

jayz22: Ah yes, that should work.

jayz22: I changed the trait to `IntoVal<Env, Vec<RawVal>>` as you suggested. It works fairly nicely for the tuples. 
However, I found it much less elegant to work with `Vec<T>` since it requires 
```
impl<T> IntoVal<Env, Vec<RawVal>> for Vec<T>
where
    T: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal>,
```
which ironically I found it hard to do. Ideally it should just do something like `self.iter().map(f).collect()` into `Vec<RawVal>` but that didn't work (or maybe I just didn't spend enough time).
@leighmcculloch If you are ok with it, I think we should merge this as is (since we are done with Tuples) and we can add a followup issue to support vector.

leighmcculloch: Agreed adding vector was just a nicety. 

Fwiw to make it work I think you need a `impl<T> From<Vec<T>> for Vec<RawVal> where T: IntoVal<Env, RawVal>` and inside that conversion you just do an unsafe unchecked_new since we can trust the types are RawVal.

soroban-sdk/src/events.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl Debug for Events {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "Events")
+    }
+}
+
+pub trait Topics: IntoVal<Env, Vec<RawVal>> {}
+
+impl IntoVal<Env, Vec<RawVal>> for () {
```
leighmcculloch, line 59: Todo: move this and the macro for tuples into tuple.rs or similar since it doesn't only apply to events but more generally is for supporting zero length tuple to Vec conversions.

leighmcculloch (approved):

leighmcculloch:
Merged. There's two things to follow up with that aren't urgent:
- https://github.com/stellar/rs-soroban-sdk/pull/469#discussion_r953244649
- https://github.com/stellar/rs-soroban-sdk/pull/469#discussion_r953246490

soroban-sdk/src/events.rs
jayz22: I don't think tuple.rs is the right place since `Vec<T>` is defined in the sdk, so maybe somewhere in the sdk (`vec.rs`?).

leighmcculloch: I was meaning in tuple.rs in the SDK, a new file. Inside vec.rs is okay too, that file is just getting big.

jayz22: Ah, Ok. Yeah that makes sense.

PR #484, opened 2022-08-24
Self invocations in tests use contractclient

### LINKED ISSUE #436: Macro: contractimpl: Move generated invoke functions into a Client type
The `contractimpl` generated invoke functions end up in a module. This makes sense for the hidden non-user extern functions that must be module level, but doesn't make sense for the invoke functions we expect users to use in tests. It would make more sense, I think, if those functions looked identical to the Client interface we're going to generate for invoking other contracts. (See #118) This way there is one way to invoke functions in tests and outside tests, and it is the same for the current contract and others.

leighmcculloch:
### What

Mark the `<fn>::invoke` functions as deprecated, and add as replacement with the developer calling their own contract using a generated client, the same client they'll use when calling other contracts.

##### Before

```rust
const ADD_CONTRACT_ID: [u8; 32] = [0; 32];
mod addcontract {
    soroban_sdk::contractimport!(
        file = "target/wasm32-unknown-unknown/release/example_add_i32.wasm"
    );
}

pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn add_with(env: Env, x: i32, y: i32) -> i32 {
        addcontract::ContractClient::add(&env, &BytesN::from_array(&env, &ADD_CONTRACT_ID), x, y)
    }
}

#[test]
fn test_functional() {
    let e = Env::default();

    let add_contract_id = BytesN::from_array(&e, &ADD_CONTRACT_ID);
    e.register_contract_wasm(&add_contract_id, addcontract::WASM);

    let contract_id = BytesN::from_array(&e, &[1; 32]);
    e.register_contract(&contract_id, Contract);

    let x = 10i32;
    let y = 12i32;
    let z = add_with::invoke(&e, &contract_id, &x, &y);
    assert!(z == 22);
}
```

##### After

```rust
const ADD_CONTRACT_ID: [u8; 32] = [0; 32];
mod addcontract {
    soroban_sdk::contractimport!(
        file = "target/wasm32-unknown-unknown/release/example_add_i32.wasm"
    );
}

pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn add_with(env: Env, x: i32, y: i32) -> i32 {
        addcontract::ContractClient::add(&env, &BytesN::from_array(&env, &ADD_CONTRACT_ID), x, y)
    }
}

#[test]
fn test_functional() {
    let e = Env::default();

    let add_contract_id = BytesN::from_array(&e, &ADD_CONTRACT_ID);
    e.register_contract_wasm(&add_contract_id, addcontract::WASM);

    let contract_id = BytesN::from_array(&e, &[1; 32]);
    e.register_contract(&contract_id, Contract);

    let x = 10i32;
    let y = 12i32;
    let z = ContractClient::add_with(&e, &contract_id, x, y);
    assert!(z == 22);
}
```

##### Diff

```diff
@@ -31,7 +32,7 @@ fn test_functional() {
 
     let x = 10i32;
     let y = 12i32;
-    let z = add_with::invoke(&e, &contract_id, &x, &y);
+    let z = ContractClient::add_with(&e, &contract_id, x, y);
     assert!(z == 22);
 }
```

### Why

The `add_with::invoke` syntax is pretty odd and not intuitive. The client syntax is something we're surfacing for calling other contracts. It seems usable as a way to call a developers own contract in tests, and then it gives them one way to do something for all contracts.

Close #436

### Known limitations

N/A

### DIFF
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -1,17 +1,81 @@
 use proc_macro2::TokenStream;
-use quote::{format_ident, quote};
-use syn::{spanned::Spanned, Error, FnArg, Pat, TraitItemMethod, Type, TypePath};
+use quote::{format_ident, quote, ToTokens};
+use syn::{
+    parse::{Parse, ParseStream},
+    punctuated::Punctuated,
+    spanned::Spanned,
+    token::Comma,
+    Error, FnArg, Ident, ItemImpl, ItemTrait, Pat, ReturnType, Token, Type, TypePath,
+};
 
-pub fn derive_client(name: &str, methods: &[&TraitItemMethod]) -> TokenStream {
+use crate::syn_ext;
+
+pub enum ClientItem {
+    Trait(ItemTrait),
+    Impl(ItemImpl),
+}
+
+impl ClientItem {
+    pub fn fns(&'_ self) -> Vec<ClientFn> {
+        match self {
+            ClientItem::Trait(t) => syn_ext::trait_methods(t)
+                .map(|m| ClientFn {
+                    ident: &m.sig.ident,
+                    inputs: &m.sig.inputs,
+                    output: &m.sig.output,
+                })
+                .collect(),
+            ClientItem::Impl(i) => syn_ext::impl_pub_methods(i)
+                .map(|m| ClientFn {
+                    ident: &m.sig.ident,
+                    inputs: &m.sig.inputs,
+                    output: &m.sig.output,
+                })
+                .collect(),
+        }
+    }
+}
+
+impl Parse for ClientItem {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        _ = input.parse::<Token![pub]>();
+        let lookahead = input.lookahead1();
+        if lookahead.peek(Token![trait]) {
+            input.parse().map(ClientItem::Trait)
+        } else if lookahead.peek(Token![impl]) {
+            input.parse().map(ClientItem::Impl)
+        } else {
+            Err(lookahead.error())
+        }
+    }
+}
+
+impl ToTokens for ClientItem {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        match self {
+            ClientItem::Trait(t) => t.to_tokens(tokens),
+            ClientItem::Impl(i) => i.to_tokens(tokens),
+        }
+    }
+}
+
+pub struct ClientFn<'a> {
+    pub ident: &'a Ident,
+    pub inputs: &'a Punctuated<FnArg, Comma>,
+    pub output: &'a ReturnType,
+}
+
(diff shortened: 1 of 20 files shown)


leighmcculloch:
At the moment in this PR the `#[contractclient]` attribute must be added manually to the impl, but it could be argued that we could make this happen automatically within the `#[contractimpl]`. The main reason I have kept it separate is so that the developer is required to provide the name of the generated type that will be the client. We could autogenerate it based on some fixed value like the name of the thing being `impl` with `Client` appended, but I dunno. Feedback appreciated. cc @jonjove @sisuresh

sisuresh:
I'd prefer autogenerating a name instead of using `#[contractclient(name = "Client")]`. I don't see the advantage of specifying the name since the user will already need to know how to interact with this client through documentation.

leighmcculloch:
Sounds good to me, we just need to figure out the rules by which it auto-generates. `contractclient` gets used in several contexts, so we need a rule system that is consistent:

1. It can be used on a single contract impl block.
   ```rust
   #[contractclient]
   impl Contract { ... }
   ```

2. Or it can be used across multiple, and for multiple we need to generate unique separate clients.
   ```rust
   #[contractclient]
   impl Contract { ... }
   #[contractclient]
   impl TokenContract for Contract { ... }
   ```

3. Or it can be used on traits. On traits we already hardcode it to `"Client"` because it is embedded inside `contractimport`.
   ```rust
   #[contractclient]
   trait Contract { ... }
   ```

It's the (2) scenario that makes auto defining a name tricky. If we put `contractclient` inside `contractimpl` we could have the `contractimpl` version take the name out of the impl, and make it `ContractClient` and `ContractTokenContractClient`. Wdyt?

sisuresh:
I like the simplicity of putting `contractclient` inside `contractimpl` and autogenerating the name, but I have a question about (2). Would there be a conflict in that scenario? I'd expect it to look like `TokenContract::Client` and `Contract::Client`.

leighmcculloch:
> I'd expect it to look like `TokenContract::Client` and `Contract::Client`.

Unfortunately Rust doesn't support nesting types in other types. If it fits, I agree this would be ideal.



> I have a question about (2). Would there be a conflict in that scenario?

There'd be a conflict if we gave the clients the same name. I think if we use the name of the trait being implemented in the client name, we'll avoid most collisions.

i.e. Use a name in the form:
```
[Type][Trait]Client
```

e.g. For `impl TokenContract for MyContract`:
```
MyContractTokenContractClient
```

😒 It looks pretty long and the repetition isn't great be. I'll edit ok on this.

leighmcculloch:
I've kept the existing way of invoking functions in the SDK for this release, but they are marked as deprecated, so developers will see this:
Screen Shot 2022-08-25 at 10 02 16 PM

leighmcculloch:
I added a TODO for the issue of multiple traits, but have decided to defer solving that problem to a future release, because the contract function set derive is another part of the macros that still prevents us from doing multiple implementations anyway.

PR #488, opened 2022-08-25
Add sym! macro for const compile time Symbol creation

leighmcculloch:
### What
Add sym! macro that creates a Symbol that is generated at compile time.

### Why
Symbol::from_str is a const fn and it can be used in a const context to generate the Symbol at compile time. However in Rust const fns are only evaluated at compile time if they are used explicitly in a const context. This means if someone uses Symbol::from_str in a non-const context inside a contract, they're introducing a couple hundred bytes of WASM instructions for the conversion logic.

### DIFF
--- a/soroban-sdk/src/symbol.rs
+++ b/soroban-sdk/src/symbol.rs
@@ -8,20 +8,20 @@ use crate::Symbol;
 /// ### Examples
 ///
 /// ```
-/// use soroban_sdk::{sym, Symbol};
+/// use soroban_sdk::{symbol, Symbol};
 ///
 /// let symbol = symbol!("a_str");
 /// assert_eq!(symbol, Symbol::from_str("a_str"));
 /// ```
 ///
 /// ```
-/// use soroban_sdk::{sym, Symbol};
+/// use soroban_sdk::{symbol, Symbol};
 ///
 /// const symbol: Symbol = symbol!("a_str");
 /// assert_eq!(symbol, Symbol::from_str("a_str"));
 /// ```
 #[macro_export]
-macro_rules! sym {
+macro_rules! symbol {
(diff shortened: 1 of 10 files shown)


leighmcculloch:
This saw reduction in size of two examples that uses Symbols.

Before:
```
-rw-r--r--  1  191 Aug 25 11:50 example_hello.wasm
-rw-r--r--  1 1130 Aug 25 11:50 example_invoke_contract.wasm
```
After:
```
-rw-r--r--  1  470 Aug 25 11:51 example_hello.wasm
-rw-r--r--  1 1401 Aug 25 11:51 example_invoke_contract.wasm
```

leighmcculloch:
This macro could live in the `soroban-env-common` crate, and I might move it there, but I've put it here for the moment because I really want the docs for it to be about using it in the SDK.

soroban-sdk/src/symbol.rs
```diff
@@ -0,0 +1,29 @@
...
+/// ```
+///
+/// ```
+/// use soroban_sdk::{sym, Symbol};
+///
+/// const symbol: Symbol = sym!("a_str");
+/// assert_eq!(symbol, Symbol::from_str("a_str"));
+/// ```
+#[macro_export]
+macro_rules! sym {
```
leighmcculloch, line 24: Even if it is a few more characters it might improve readability if this macro was `symbol!`. Wdyt @jonjove @sisuresh @paulbellamy?

leighmcculloch: I updated the change in 8836979d to use `symbol!`.

jonjove: +1, I would have chosen `symbol!` as well.

PR #490, opened 2022-08-25
Add JSON representation for a soroban sc spec

tamirms:
### What

Add JSON representation for a soroban smart contract spec. Checkout https://github.com/stellar/soroban-cli/pull/91/files for how `SpecType` is used to serialize the smart contract into JSON.

For the following function `pub fn add(a: BigInt, b: BigInt) -> BigInt` this is what the JSON output looks like:

```
{"type":"function","name":"add","inputs":[{"name":"a","value":{"type":"bigInt"}},{"name":"b","value":{"type":"bigInt"}}],"outputs":[{"type":"bigInt"}]}
```

### Why

This is necessary to implement https://github.com/stellar/soroban-cli/issues/56 .

### Known limitations

[N/A]

### DIFF
--- a/soroban-spec/src/gen/json.rs
+++ b/soroban-spec/src/gen/json.rs
@@ -15,11 +15,9 @@ impl TryFrom<&ScSpecUdtStructFieldV0> for StructField {
     type Error = Error;
 
     fn try_from(f: &ScSpecUdtStructFieldV0) -> Result<Self, Self::Error> {
-        let name = f.name.to_string()?;
-        let value = SpecType::try_from(&f.type_)?;
         Ok(StructField {
-            name: name,
-            value: value,
+            name: f.name.to_string()?,
+            value: SpecType::try_from(&f.type_)?,
         })
     }
 }
@@ -35,11 +33,9 @@ impl TryFrom<&ScSpecFunctionInputV0> for FunctionInput {
     type Error = Error;
 
     fn try_from(f: &ScSpecFunctionInputV0) -> Result<Self, Self::Error> {
-        let name = f.name.to_string()?;
-        let value = SpecType::try_from(&f.type_)?;
         Ok(FunctionInput {
-            name: name,
-            value: value,
+            name: f.name.to_string()?,
+            value: SpecType::try_from(&f.type_)?,
         })
     }
 }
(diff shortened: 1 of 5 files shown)


soroban-spec/src/gen/json.rs
```diff
@@ -0,0 +1,310 @@
...
+        "name": "UdtEnum"
+      }
+    }
+  ],
+  "outputs": [
+    {
+      "type": "i64"
+    }
+  ]
+}"#
```
leighmcculloch, line 307: @paulbellamy I think you were the stakeholder who asked for an easier to digest JSON format. Would you make any changes to this format? We could probably simplify it a little by removing the explicit 'userDefined' types and converting some objects inside arrays into objects. For example, the fields list could be a map/object of name to type. I don't feel strongly though.

leighmcculloch (approved):
Looks good to me. I pushed a test to see what it looks like. Feel free to add more tests if you think it would be helpful. Great first contribution 👏🏻.

cc @paulbellamy See my comment below for any feedback.

@tamirms Let me know if you're otherwise ready for this to be merged and I'll merge it.

tamirms:
@leighmcculloch thanks for the review, please go ahead and merge unless @paulbellamy finds any issues that should be fixed

soroban-spec/src/gen/json.rs
```diff
@@ -0,0 +1,310 @@
...
+    type Error = Error;
+
+    fn try_from(spec: &ScSpecTypeDef) -> Result<Self, Self::Error> {
+        match spec {
+            ScSpecTypeDef::Map(inner) => {
+                let key = SpecType::try_from(inner.key_type.as_ref());
+                let value = SpecType::try_from(inner.value_type.as_ref());
+                Ok(SpecType::Map {
+                    key: Box::new(key?),
+                    value: Box::new(value?),
```
leighmcculloch, line 132: I'm going to make a quick change to this PR before merging. It is helpful if the `?`'s appear near the thing that can fail, so rather than putting `?` on `r` it's helpful if it exists up near the try_from.

soroban-spec/src/gen/json.rs
paulbellamy: Yeah, here I think it's more important to be unambiguous without context. But I don't have a strong opinion either.

soroban-spec/src/gen/json.rs
```diff
@@ -0,0 +1,283 @@
...
+    },
+    Struct {
+        name: String,
+        fields: Vec<StructField>,
+    },
+    Union {
+        name: String,
+        cases: Vec<UnionCase>,
+    },
+}
```
leighmcculloch, line 107: @tamirms I was about to merge but noticed the entries were bundled with the list of types. I split them out separately because I think they're two different things, it doesn't have any material impact on the JSON though.

leighmcculloch (approved):
🎉 Thanks @tamirms, this is a great first contribution, and I liked how you used the TryFrom conversions, error handling. The resulting JSON is a good balance of succinct and easy to work with. Also, the use of `tag` with serde is cool, TIL.

paulbellamy (approved):

soroban-spec/src/gen/json.rs
tamirms: thanks @leighmcculloch I agree that it's much cleaner with your modifications!

PR #491, opened 2022-08-25
Remove use of iterator from enum decoding

leighmcculloch:
### What
Remove use of iterator from enum decoding

### Why
It adds additional WASM for little to no gain. The length of the vec is always known due to the length check which is required anyway. So there's no reason to add additional branches checking if the value is present or not, since we know it is.


jonjove:
Right now the length check is inequality rather than exact equality. Should we change to exact equality now?

leighmcculloch:
I'm going to close this, because this change makes this code more brittle than it already is. I think we should address this when we add support for other types of enum variants that have more than one value in the tuple.

PR #503, opened 2022-08-27
Add Deployer

leighmcculloch:
### What
Add Deployer

### Why
Close https://github.com/stellar/rs-soroban-env/issues/348


leighmcculloch:
I paired with @jonjove this afternoon a little on this, and we came to the resolution that:
 - Move getting the contract ID into the deployer rather than separately outside.
 - Move the salt into the list of args that define the ID, rather than separating the salt from the other params.
 - Try removing the generics and just put the different deployers on functions.

I solo'd a bit on it after this and got it to where it is now. I think in the interest of having something to work with we should merge this and then think about it some more in the next iteration.

I've included some functions we don't have host functions for, and so I've put a `todo!()` in those functions and marked them as doc(hidden).

cc @jonjove @sisuresh

Issue #398, opened 2022-08-03
Type: Vec: Add contains, binary search, and index of fns


jonjove:
### What problem does your feature solve?

`Vec` is missing functions from the `std::Vec` interface. This issue was prompted by a message explicitly asking about `contains`.

### What would you like to see?

Add some of these functions.

### What alternatives are there?

TODO

leighmcculloch:
Adding `contains` and other similar functions requires additional host functions. I opened https://github.com/stellar/rs-soroban-env/issues/317 requesting this.

leighmcculloch:
> Add some of these functions.

The Vec type will likely go through many iterations of improvement when it comes to functions, so let's scope this issue to offering the following functions that relate to `contains` and are likely to be served by the same underlying host fns:
- `contains` - boolean for whether the vec contains the value
- `binary_search` - Similar to `contains`, but assumes a sorted vec, and will also return the insert point if not found.
- `first_index_of` - There is no `std::vec::Vec` version of this. It is a common pattern to find the position of an item in a vec by using an iterator in the form `vec.iter().position(...)`. It is not efficient to use an iterator for this use case since it would move so much logic into the contract, so we should offer a similar function on `Vec` itself. In the SDK layer we can debate the use of the term index vs position. Imo position makes more sense in the context of an iterator, and a little less sense in the context of contiguous vector, but I don't think it would be harmful to go either way.
- `last_index_of` - Similar to above.

leighmcculloch:
Blocked on https://github.com/stellar/rs-soroban-env/issues/317

PR #516, opened 2022-08-29
Add Vec contains, first_index_of, last_index_of, binary_search

leighmcculloch:
### What
Add Vec contains, first_index_of, last_index_of, binary_search

### Why
These are common functions on vector types and was requested by contract developers.

Close https://github.com/stellar/rs-soroban-sdk/issues/398

### DIFF
--- a/soroban-sdk/src/vec.rs
+++ b/soroban-sdk/src/vec.rs
@@ -1,4 +1,5 @@
 use core::{
+    borrow::Borrow,
     cmp::Ordering,
     fmt::Debug,
     iter::FusedIterator,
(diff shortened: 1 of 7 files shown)


soroban-sdk/src/vec.rs
```diff
@@ -563,6 +564,66 @@ where
     }
 }
 
+impl<T> Vec<T>
+where
+    for<'a> &'a T: IntoVal<Env, RawVal>,
+{
+    /// Returns true if the Vec contains the item.
+    #[inline(always)]
+    pub fn contains(&self, item: impl Borrow<T>) -> bool {
```
leighmcculloch, line 573: @jonjove @graydon I'm interested in your feedback on the interface I'm using for these functions where I'm accepting a `Borrow<T>` so that people can pass a `&T` or a `T`. I saw there is a more complicated use of this in the docs at https://doc.rust-lang.org/std/borrow/trait.Borrow.html, but it is unclear to me if we need to go to that length.

jonjove: I'm good with this.

PR #522, opened 2022-08-30
Map RawVal to Val in spec

leighmcculloch:
### What
Map RawVal to Val in spec.

### Why
The RawVal type should be mapped to the val type, not a user defined type.


soroban-sdk-macros/src/map_type.rs
```diff
@@ -44,70 +45,70 @@ pub fn map_type(t: &Type) -> Result<ScSpecTypeDef, Error> {
                 }) => {
                     let args = angle_bracketed.args.iter().collect::<Vec<_>>();
                     match &ident.to_string()[..] {
-                    "Option" => {
-                        let t = match args.as_slice() {
+                        "Option" => {
```
leighmcculloch, line 48: Cargo fmt wanted to reformat all these lines for some reason.

PR #561, opened 2022-09-06
Add iterator traits to Set

accordeiro:
### What

Add iterator for the Soroban `Set` type.

### Known limitations

Soroban `Set` is syntax sugar for maps, so there is an implicit ordering that does not preserve insertion order. This behavior is probably fine and is consistent with other Soroban types like `Map`.


leighmcculloch (approved):
🎉

PR #563, opened 2022-09-06
Add conversion traits for Set

accordeiro:
### What

Add conversion traits and more tests to the Set type.


leighmcculloch (approved):
Nice 👍🏻

PR #575, opened 2022-09-08
Add publish dry run to build on release branches

leighmcculloch:
### What
Add publish dry run to build on release branches.

### Why
So that release branches created by the bump version workflow are fully tested before going to the publish step after merge.


leighmcculloch:
This build won't pass because the SDK is currently dependent on unreleased versions of its deps, which is exactly the type of thing this change is intended to detect and block. Since I'm just adding this change to CI now for future use, I'm closing this change, and opening the same change with a different branch name so that the check won't run now and we can merge it now. See #576.

PR #581, opened 2022-09-09
Fix Ord::cmp impls of all types

leighmcculloch:
### What
Change Ord::cmp impls of all types to use the <EnvVal as Ord>::cmp impl.

### Why
The EnvVal type provides an optimized implementation of the Ord::cmp trait with a couple of fast paths that avoid a host function call. The EnvVal type was designed by @graydon to provide easy comparison between types without us having to reimplement the comparison everywhere ourselves. It looks like this was overlooked when implementing some, but not all, SDK types.

Also, @graydon pointed out that the implementation of these comparison functions in the SDK are incorrect because they consume the result as an i32 encoded in a RawVal when in-fact they are i64 values without any RawVal encoding.

### DIFF
--- a/soroban-sdk/src/account.rs
+++ b/soroban-sdk/src/account.rs
@@ -53,10 +53,7 @@ impl PartialOrd for Account {
 
 impl Ord for Account {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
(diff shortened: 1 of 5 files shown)


graydon (approved):
Pretty sure this is right!

PR #582, opened 2022-09-09
Fix conversions from guaranteed types

leighmcculloch:
### What
Change fallible conversions from guaranteed types to do an unsafe direct conversion. Also, in the process fixed bits to just return the raw u64 value.

### Why
The SDK can trust the host to always return the correct type for returned values that have types that are fixed. If the host says a return value from a function will always be a RawVal wrapping a u32, we can unsafe convert it to a u32 without adding a branch and panic on failure. This reduces some branches in the compiled WASM.

Also the bits function was converting a returned value to a u32 but the value was in fact just a raw u64 value.

### DIFF
--- a/soroban-sdk/src/bigint.rs
+++ b/soroban-sdk/src/bigint.rs
@@ -1094,15 +1094,13 @@ impl BigInt {
     /// Returns true if the [BigInt] is zero.
     pub fn is_zero(&self) -> bool {
         let env = self.env();
-        let is_zero = env.bigint_is_zero(self.0.to_object());
-        bool::try_from(is_zero).unwrap()
+        env.bigint_is_zero(self.0.to_object()).is_true()
     }
 
     /// Returns the minimum number of bits required to store the [BigInt].
-    pub fn bits(&self) -> u32 {
+    pub fn bits(&self) -> u64 {
         let env = self.env();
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/bigint.rs
```diff
@@ -1094,15 +1094,13 @@ impl BigInt {
...
+        env.bigint_is_zero(self.0.to_object()).is_true()
     }
 
     /// Returns the minimum number of bits required to store the [BigInt].
-    pub fn bits(&self) -> u32 {
+    pub fn bits(&self) -> u64 {
         let env = self.env();
-        let bits = env.bigint_bits(self.0.to_object());
-        u32::try_from(bits).unwrap()
+        env.bigint_bits(self.0.to_object())
```
leighmcculloch, line 1103: @jayz22 @graydon Why does the `bigint_bits` function return `u64`? In Rust built-in primitive types like u32, u64, all define their number of bits in `u32`. I find it highly unlikely our BigInt type would realistically be able to hold more than u32::MAX bits either.

leighmcculloch: Reposted this question at https://github.com/stellar/rs-soroban-env/issues/423. Let's discuss there.

PR #588, opened 2022-09-10
Add BigInt mutable ref for PartialEq, PartialOrd

leighmcculloch:
### What
Add BigInt mutable ref for PartialEq, PartialOrd.

### Why
So that disambiguating the impl to use is not required for comparing &mut BigInts.

### DIFF
--- a/soroban-sdk/src/bigint.rs
+++ b/soroban-sdk/src/bigint.rs
@@ -1258,60 +1258,70 @@ impl PartialEq for BigInt {
         self.partial_cmp(other) == Some(Ordering::Equal)
     }
 }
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/bigint.rs
```diff
@@ -1671,16 +1681,16 @@ mod test {
 
         let mut b = bigint!(&env, 1);
         b += &1;
-        assert_eq!(b.clone(), bigint!(&env, 2));
+        assert_eq!(b, bigint!(&env, 2));
```
leighmcculloch, line 1684: The clone was being used to disambiguate which impl to use.

Issue #532, opened 2022-08-31
Add test facility to collect events and assert on them


leighmcculloch:
We need a way in tests to be able to assert on events having been published and their contents.

PR #606, opened 2022-09-14
Add ability to assert on events

leighmcculloch:
### What
Add ability to assert on events.

### Why
So that contract tests can assert on events published.

Close https://github.com/stellar/rs-soroban-sdk/issues/532

### DIFF
--- a/soroban-sdk/src/events.rs
+++ b/soroban-sdk/src/events.rs
@@ -117,12 +117,12 @@ impl Events {
 }
 
 #[cfg(feature = "testutils")]
-use crate::{xdr, BytesN, TryIntoVal};
+use crate::{xdr, TryIntoVal};
 
 #[cfg(feature = "testutils")]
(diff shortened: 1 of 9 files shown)


soroban-sdk/src/events.rs
```diff
@@ -112,3 +115,43 @@ impl Events {
...
+                    contract_id: Some(contract_id),
+                    body: xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, data }),
+                    ..
+                }) = e
+                {
+                    vec.push_back((
+                        contract_id.0.into_val(env),
+                        topics.try_into_val(env).unwrap(),
+                        data.try_into_val(env).unwrap(),
+                    ))
```
leighmcculloch, line 152: @jonjove I know we had some doubts on whether the ScVal <=> RawVal conversions were worth it, but they do allow for some nice conveniences in the SDK. Check this out ☝🏻 where we can do relatively seamless conversions from ScVal types to SDK types.

Issue #482, opened 2022-08-24
Support debug events in the SDK


leighmcculloch:

leighmcculloch:
@graydon I think I might need some help from you to understand the best way, or if it is possible to utilise things like debug events in the SDK to gain more insight into things like conversion errors or panics. I know you had a vision for how debug events would work, but I'm not sure if it carries outside the host and into contracts.

I put this issue in the September iteration.

PR #611, opened 2022-09-15
Add logger and debug events

leighmcculloch:
### What
Add debugger and debug events, and hook up to conversion errors.

### Why
To complete the debug events / logging story.

So that developers can output their own debug events.

So that builtin conversion errors are easier to understand.

This layers on top of the debug events that @graydon added sometime ago.

Close https://github.com/stellar/rs-soroban-sdk/issues/482


leighmcculloch:
Here's an example of what it looks like to use:
Screen Shot 2022-09-15 at 3 15 06 PM
cc @graydon

leighmcculloch:
@graydon It is worth noting that the debug output of Vecs is pretty underwhelming in debug events ☝🏻. In the SDK Vec's have a very different output as their Debug implementation outputs every individual element. Maybe we should do something similar in the host?

leighmcculloch:
Also, the logs aren't attributable to a contract when multiple contracts are running. There's a follow up issue for that: https://github.com/stellar/rs-soroban-env/issues/438.

PR #624, opened 2022-09-16
Do not export contractimport!-ed contracttypes!

### LINKED ISSUE #622: Contract imported types can collide in contract spec with local contract types
The `contractimport!` generates types for contract types and also exports all the types. This means if you import a contract, all its types are always exported, which isn't ideal, because you might just want to call the contract not export the types.

We should make the types not exported by default.

We should let the developer choose which to reexport, but that is out of scope of this issue, and is covered by https://github.com/stellar/rs-soroban-sdk/issues/623.

leighmcculloch:
### What
Do not export contractimport!-ed contracttypes!

### Why
Close #622

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -103,13 +103,13 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
 #[derive(Debug, FromMeta)]
 struct ContractTypeArgs {
     lib: Option<String>,
+    export: Option<bool>,
 }
 
 /// Generates conversions from the struct/enum from/into a `RawVal`.
 ///
 /// Includes the type in the contract spec so that clients can generate bindings
 /// for the type.
-///
 #[proc_macro_attribute]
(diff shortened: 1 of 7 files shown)


soroban-sdk-macros/src/lib.rs
```diff
@@ -119,7 +119,13 @@ pub fn contracttype(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
     let input = parse_macro_input!(input as DeriveInput);
     let ident = &input.ident;
-    let gen_spec = matches!(input.vis, Visibility::Public(_));
+    // If the export argument has a value, do as it instructs regarding
+    // exporting. If it does not have a value, export if the type is pub.
+    let gen_spec = if let Some(export) = args.export {
+        export
+    } else {
+        matches!(input.vis, Visibility::Public(_))
+    };
```
leighmcculloch, line 128: I really wanted to write a test for this, but it is rather difficult without setting up a full crate that builds a wasm then analyzes its contract spec, which is not efficient. I think we will need to catch issues with this code via examples and test vectors, which we do have in other repos. It isn't ideal.

PR #626, opened 2022-09-16
Use log_value in logging temporarily until Guest Env supports debug events

### LINKED ISSUE #625: Fix logs temporarily until debug events support guest
I misunderstood how debug events work and didn't realize they don't work in wasm builds because the guest hasn't implemented them. I had tested them only in native builds, but wasm builds is where the logging was designed to be used.

I've opened issue https://github.com/stellar/rs-soroban-env/issues/447 for adding support to the guest env so it will work in wasm development builds.

That issue will take some time to realize though, so we should change the SDK to push the logs to the `log_value` host function which is a primitive function that accepts just a RawVal.

leighmcculloch:
### What
Use log_value in logging temporarily until Guest Env supports debug events

### Why
Close #625

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -269,11 +269,6 @@ impl Env {
         let stack = internal::Env::get_current_call_stack(self);
         stack.try_into_val(self).unwrap()
     }
-
-    #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Closing this

Issue #181, opened 2022-07-06
Test Fwk: Add Env fns for setting/modifying/inspecting ledger state.


leighmcculloch:

leighmcculloch:
Similar to #180, @jonjove has suggested that we should just have contract developers create ledger state using the host interface in the same way that they would as part of their contract, without touching XDR, etc.

I think there are use cases setting specific ledger state, but I see a developer can get a long way by just doing the above. So I think we should move this to the next iteration. I'm moving this to the August iteration. cc @jonjove @graydon @tomerweller

PR #677, opened 2022-09-27
Use AccountId instead of BytesN<32> for account ids

leighmcculloch:
### What
Use AccountId instead of BytesN<32> for account ids.

### Why
The env is changing to use AccountId because that's what is used for the invoker.

Close https://github.com/stellar/rs-soroban-env/issues/477


leighmcculloch:
Note that this PR has limited tests because we don't have a reasonable way to create accounts, although that's coming in https://github.com/stellar/rs-soroban-sdk/issues/181.

sisuresh (approved):

PR #690, opened 2022-09-29
Make classic account auth more in line with the Core.

dmkozh:
### What

- Limit the signature count (in core the limit is in XDR)
- Limit the weight to be not more than 255 (as thresholds are `u8`)

### Why

Make this more consistent with what Core does and with classic account limitations. That's an SDK counterpart to https://github.com/stellar/rs-soroban-env/pull/493

### Known limitations

Can't cover this in tests yet.


leighmcculloch (approved):
🎉 This looks good to me. Thanks for following this up so quickly.

I'll merge this when https://github.com/stellar/rs-soroban-env/pull/493 is approved by @sisuresh or @jonjove and merged.

soroban-auth/src/lib.rs
```diff
@@ -67,10 +76,9 @@ fn verify_account_signatures(env: &Env, auth: &AccountSignatures, name: Symbol,
         }
 
         env.verify_sig_ed25519(&sig.public_key, &msg_bytes, &sig.signature);
-
-        weight = weight
-            .checked_add(acc.signer_weight(&sig.public_key))
-            .expect("weight overflow");
+        // Signature weight can be at most 255, hence overflow isn't possible
+        // here as 255 * MAX_ACCOUNT_SIGNATURES is < u32::MAX.
+        weight += acc.signer_weight(&sig.public_key);
```
leighmcculloch, line 81: Once https://github.com/stellar/rs-soroban-env/pull/493#discussion_r984255735 is changed to return 0 from the signer_weight function, we need to update this code to test for `0` and panic.

leighmcculloch (commented):
Just one thing to do in this change and I think it's good to merge. 👏🏻

soroban-auth/src/lib.rs
dmkozh: Done.

leighmcculloch (approved):

PR #695, opened 2022-09-30
Remove init_wrap method from token spec

leighmcculloch:
### What
Remove init_wrap method from token spec and rename init_token to init.

### Why
Contracts never need to call init_wrap, so we keep the interface simpler if we don't advertise it to them. Since they only need to use one init function, we should simplify the name.

Follow up to https://github.com/stellar/rs-soroban-env/issues/509.


paulbellamy:
BTW, we should have updated soroban-cli with the new method name.

leighmcculloch:
Thanks @paulbellamy!

PR #707, opened 2022-10-06
Rename Invoker -> Address

leighmcculloch:
### What
Rename the `Invoker` SDK type to `Address`.

### Why
@tomerweller was looking at one of the examples and was pretty confused by the use of the `Invoker` as an argument. He pointed out to me how it was confusing because it looked like the argument would contain the invoker of this contract, when in fact it was just an argument that anyone could pass. This signals, I think, that using the term `Invoker` for the type could be a reasonably large footgun. The environment and the contract invocation will have an invoker, and that invoker is an address of an account or a contract.

Another alternative name to use would be an `Id`. Note that we can't really use the longer `Identifier` because it would overlap with the `soroban-auth` `Identifier`.

Another reason to use the term `Address` as it can be said to imply that it points to something tangible on network. Where-as the more generic `Identifier` in `soroban-auth` points to any identifier that may not address something on network.

I don't love making a name change this late, but this seems warranted.

@tomerweller @paulbellamy @sisuresh Wdyt?

Note that I'm not updating the XDR spec type from Invoker to Address. I'll do this at a later point, as it has very little impact or visibility.


tomerweller (approved):
I strongly support this change. `Invoker` is confusing and having an `Address` that can be an account or a contract is well aligned with other ecosystems.

leighmcculloch:
This change will be minimal impact because soroban-auth shields most examples from the change. It will require updating one example.

It should require zero changes to the dapp, because the dapp uses soroban-auth for token auth.

PR #725, opened 2022-10-11
Add conversion from `Address` to `Identifier`.

dmkozh:
### What

Add conversion from `Address` to `Identifier`.

### Why

Simplify code that uses `env.invoker` for getting identifiers.

### Known limitations

N/A

### DIFF
--- a/soroban-auth/src/public_types.rs
+++ b/soroban-auth/src/public_types.rs
@@ -75,6 +75,15 @@ impl From<&AccountId> for Identifier {
     }
 }
 
+impl From<Address> for Identifier {
+    fn from(addr: Address) -> Self {
+        match addr {
+            Address::Account(a) => Identifier::Account(a),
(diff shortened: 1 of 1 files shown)


soroban-auth/src/public_types.rs
```diff
@@ -75,6 +75,15 @@ impl From<&AccountId> for Identifier {
     }
 }
 
+impl From<Address> for Identifier {
```
leighmcculloch, line 78: Could you also add `From<&Address> for Identifier` too? These elements are cheap to clone.

leighmcculloch (approved):

soroban-auth/src/public_types.rs
dmkozh: Added.

leighmcculloch: Thanks!!!

leighmcculloch (approved):

Issue #642, opened 2022-09-21
Env::assert_panic_with_status does not catch HostError's containing Status


leighmcculloch:

leighmcculloch:
In https://github.com/stellar/rs-soroban-sdk/pull/641 I added use of the `assert_panic_with_status` function which raises a panic containing a `HostError` that contains a `Value` that is `Status`, but the assert function doesn't capture it, for some reason it sees the panic as a `String`.

leighmcculloch:
This test sees the following exception:

https://github.com/stellar/rs-soroban-sdk/blob/6292b9da98ddb8e09cef9b0ac96461cc0e8aa8c3/tests/result/src/lib.rs#L50-L53

```
thread 'test::hello_error' panicked at 'inner function panicked with "called `Result::unwrap()` on an `Err` value: HostError
Value: Status(ContractError(1))

Debug events (newest first):
   0: "Status(ContractError(1))"

Backtrace (newest first):
   0: backtrace::backtrace::libunwind::trace
             at /Users/leighmcculloch/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/backtrace/libunwind.rs:93:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/leighmcculloch/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/backtrace/mod.rs:66:5
   1: backtrace::backtrace::trace
             at /Users/leighmcculloch/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/backtrace/mod.rs:53:14
   2: backtrace::capture::Backtrace::create
             at /Users/leighmcculloch/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/capture.rs:176:9
   3: backtrace::capture::Backtrace::new_unresolved
             at /Users/leighmcculloch/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/capture.rs:170:9
   4: <soroban_env_host::host::error::HostError as core::convert::From<T>>::from
             at /Users/leighmcculloch/.cargo/git/checkouts/rs-soroban-env-230c62c6b1f3d3f1/a1f3a5c/soroban-env-host/src/host/error.rs:113:25
   5: <T as core::convert::Into<U>>::into
[...]

PR #746, opened 2022-10-28
Remove panic-catching and fix tests that use newly-working native try_call

graydon:
fixes https://github.com/stellar/rs-soroban-env/issues/462
fixes https://github.com/stellar/rs-soroban-sdk/issues/642

This "fixes" the above bugs by removing the panic-catching machinery from the SDK as discussed with @leighmcculloch -- the host now does a `catch_panic` inside `try_call` and calls `rollback` itself, so it functions "like the VM". This change is mostly just removing stuff from the SDK and tweaking tests to expect correct behaviour.

I think the only slightly hinky thing in here is that the tests assert on a _specific_ status that differs depending on whether the panic was caught natively by the host or due to a vm trap. I think this is the best we can do here and now, but we might want to make a note of it somewhere as a consideration when assigning the final error-code repertoire: it might be better to merge these two error codes and just restrict the difference between a failed VM `try_call` and a failed native `try_call` to differences in `DebugEvent`s

cc @jayz22


soroban-sdk/src/tests/contract_invoke.rs
```diff
@@ -21,9 +20,15 @@ fn test_invoke() {
 }
```
leighmcculloch, line 20: @graydon Could we make this test pass by retaining the panic string and plopping it into a debug event?

There are plenty of examples that rely on the panic string in the test. I would like us to move away from that type of testing and instead use errors, but I don't want to take away the ability to should on the error string if we don't have to.

leighmcculloch (commented):
🎉 This looks great. I have one ask inline, that I think requires a tweak to the env.

soroban-sdk/src/tests/contract_invoke.rs
graydon: Env tweaked! Posting update.

leighmcculloch (approved):
🎉🎉🎉🎉🎉🎉

leighmcculloch:
> it might be better to merge these two error codes and just restrict the difference between a failed VM try_call and a failed native try_call to differences in DebugEvents

+1 I think doing something like this might be better. The more I think about I wonder if the distinction of "vm errors" vs non-vm errors really matters. It feels too much like an implementation detail anyway. Users of the host shouldn't need the distinction.

PR #749, opened 2022-10-31
Add assert_with_error! macro

leighmcculloch:
### What
Add `assert_with_error!` macro that works like `assert!` but uses `panic_error!`.

### Why
I've noticed that reviewing examples written by the folks experimenting with Soroban that folks like to use `assert!`. We provide a `panic_error!` function for folks who like to panic on error, so that they can panic with an error code instead of a string. Given the patterns I'm seeing with `assert!` I think it's worth also providing same for error codes, and seeing if folks find error codes preferrable with the same assertion pattern.

### DIFF
--- a/soroban-sdk/src/lib.rs
+++ b/soroban-sdk/src/lib.rs
@@ -129,7 +129,30 @@ pub use soroban_sdk_macros::symbol;
 macro_rules! panic_error {
     ($env:expr, $error:expr) => {{
         $env.panic_error($error);
-        unreachable!()
+        unreachable!();
+    }};
+}
+
+/// Assert a condition and panic with the given error if it is false.
+///
+/// The first argument in the list must be a reference to an [Env].
+///
+/// The second argument is an expression that if resolves to `false` will cause
+/// a panic with the error in the third argument.
+///
+/// The third argument is an error value. The error value will be given to any
+/// calling contract.
+///
+/// Equivalent to `assert!`, but with an error value instead of a string. The
+/// error value will be given to any calling contract.
+///
+/// See [`contracterror`] for how to define an error type.
+#[macro_export]
+macro_rules! assert_error {
+    ($env:expr, $cond:expr, $error:expr) => {{
+        if !($cond) {
+            $crate::panic_error!($env, $error);
+        }
     }};
 }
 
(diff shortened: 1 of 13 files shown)


soroban-sdk/src/lib.rs
```diff
@@ -129,7 +129,30 @@ pub use soroban_sdk_macros::symbol;
...
+///
+/// The third argument is an error value. The error value will be given to any
+/// calling contract.
+///
+/// Equivalent to `assert!`, but with an error value instead of a string. The
+/// error value will be given to any calling contract.
+///
+/// See [`contracterror`] for how to define an error type.
+#[macro_export]
+macro_rules! assert_error {
```
leighmcculloch, line 151: I chose the name `assert_error` because it parallels `panic_error` (which parallels `panic_any`). I don't love the name `assert_error`, because it reads more like we're asserting that an error should occur, rather than creating an assertion that will cause an error if false. I've gone with consistency, but open to suggestions.

willemneal: I find it a little misleading. You are not asserting that there is an error, rather the opposite.  For near's sdk, they went with `require`, which I would advocate for.

Also I can't wait until macro's can be bound to types! `env.require!(...)` reads better to me.

leighmcculloch: > Also I can't wait until macro's can be bound to types! `env.require!(...)` reads better to me.

Is this in development? Do you have a link to the tracking PR / RFC? I can't find it.

willemneal: It's called [postfix macros](https://github.com/rust-lang/rfcs/pull/2442)  [polyfill here](https://github.com/est31/postfix-macros).

leighmcculloch: @graydon suggested `assert_with_error` or `assert_or_error`.

On it's own I think I prefer `assert_or_error` over require because it retains the assert language.

If we wanted panic_error to also follow, then the `_with_error` variant would probably fit better, i.e. `assert_with_error` and `panic_with_error`.

Thoughts? cc @willemneal @graydon @paulbellamy

willemneal: `*_with_error` makes sense to me.

leighmcculloch: Follow up: https://github.com/stellar/rs-soroban-sdk/pull/752

PR #750, opened 2022-10-31
Reintroduce an optimized aborting unwrap

leighmcculloch:
### What
Add an `unwrap_optimized` function to `Option` and `Result` that on wasm builds aborts immediately.

### Why
Sometime ago we had a form of this, and we removed it (https://github.com/stellar/rs-soroban-env/commit/330014541e30eb2565736061f9161c54b5d3f29c) in favor of using nightly to strip panic_fmt information that gets stuck in the resulting binary. That is still the preferred route to preventing panic info ending up in wasm32 builds because we want developers to as much as possible to have regular experience writing Rust code. But we call .unwrap() in a number of places internally in the SDK, and we could optimize those individual calls in the SDK, so we may as well so that non-nightly builds are smaller and not bloated as a result of type unwraps at function boundaries.

In the past when we had this we had it in the rs-soroban-env repo, however I think this should only belong in the SDK because it should only exist in wasm32 builds of contracts, so I'm adding it here this time.

To be clear this optimized unwrap is doc hidden and not intended for contract developers to use. We made a decision a while ago not to augment developers use of unwrap, and I think that stands, this is just for internal use.

### DIFF
--- a/soroban-auth/src/lib.rs
+++ b/soroban-auth/src/lib.rs
@@ -18,7 +18,9 @@
 
 mod tests;
 
-use soroban_sdk::{serde::Serialize, BytesN, Env, IntoVal, RawVal, Symbol, Vec};
+use soroban_sdk::{
+    serde::Serialize, unwrap::UnwrapOptimized, BytesN, Env, IntoVal, RawVal, Symbol, Vec,
+};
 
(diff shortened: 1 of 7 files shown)


graydon (approved):
LGTM

PR #751, opened 2022-10-31
Remove type checks for values delivered from the host

leighmcculloch:
### What
Remove type checks for values delivered by the host to the guest.

### Why
We said a while ago that values coming from the host to the guest would be trusted. We make this assumption in many places already. In various we still do a type check on the guest side, which seems to be inconsistent and unnecessary. This removes unnecessary branches from contracts and reduces their size.

### Merging
This is intended to be merged to `main` after #750.


soroban-sdk/src/env.rs
```diff
@@ -207,9 +206,12 @@ impl Env {
...
     pub fn verify_sig_ed25519(&self, pk: &BytesN<32>, msg: &Bytes, sig: &BytesN<64>) {
-        internal::Env::verify_sig_ed25519(self, msg.to_object(), pk.to_object(), sig.to_object())
-            .try_into()
-            .unwrap()
+        let _ = internal::Env::verify_sig_ed25519(
+            self,
+            msg.to_object(),
+            pk.to_object(),
+            sig.to_object(),
+        );
```
leighmcculloch, line 214: @jayz22 I'm pretty sure I don't need to check the return value of this host function as the host should stop the contract inside the host. Is that right?

graydon (approved):
LGTM

PR #764, opened 2022-11-14
Add testutils to build ed25519 keypairs

heytdep:
### What

Test utils to build ed25519 keypairs from a Stellar public key and secret key. 

### Why

There currently is no way to build signatures for existing ed25519 keypairs, so if I wanted to generate signatures in tests or in a script to use them later with the CLI, I'd have to build the keypair and strkey decoding myself. With this PR users are provided with a function to build the keypair from a Stellar public and secret key (as you can see in tests).

### Known limitations

N/A


soroban-auth/src/testutils.rs
```diff
@@ -38,6 +38,25 @@ pub mod ed25519 {
...
+    pub fn signer(
+        env: &Env,
+        secret_key: &[u8; 32],
+    ) -> (
+        IdentifierValue,
+        impl Identifier + Sign<SignaturePayload, Signature = [u8; 64]> + Debug,
+    ) {
+        let secret = ed25519_dalek::SecretKey::from_bytes(secret_key).unwrap();
+        let public = ed25519_dalek::PublicKey::from(&secret);
+        let signer = ed25519_dalek::Keypair { secret, public };
```
leighmcculloch, line 51: @heytdep I renamed the `build` function to `signer` because I really like this idea of being able to construct the types, and it seems useful to side-by-side add a `identifier` function to get an identifier from the same type of input.

I changed the inputs of `build` to be a single secret key because the public key can be derived from the secret key. That way in the test you only have to provide one parameter, not two, which is a little more convenient. This also ensures the public key and secret key aren't out of sync, because there's no validation they are a match.

soroban-auth/src/testutils.rs
```diff
@@ -38,6 +38,25 @@ pub mod ed25519 {
...
+        (signer.identifier(env), signer)
+    }
+
+    /// Create an identifier from a ed25519 public key.
+    pub fn identifier(env: &Env, public_key: &[u8; 32]) -> IdentifierValue {
+        IdentifierValue::Ed25519(public_key.into_val(env))
+    }
+
```
leighmcculloch, line 59: @heytdep I suggest we remove the part that added the strkey package in this change, because we can use those types directly by importing them. I get the strkey package is not as easy to use though, and if you have ideas for how to make it easier to use I'd be really interested. It's code is here: https://github.com/stellar/rs-stellar-strkey.

sisuresh (approved):
LGTM

PR #766, opened 2022-11-28
Update contract deployment to match the Env changes

dmkozh:
### What

This change makes SDK compatible with the host/env changes to the contract deployment:

- Use new host function API to register WASM/token contracts
- Deploy contracts using WASM hash instead of WASM blobs
- Remove ED25519 deployer for the time being (until env supports it)
- Do the respective test updates

### Why

Supporting the host changes.

### Known limitations

This PR removes the possibility for registering WASM/tokens with arbitrary user-provided IDs. I don't think it's a good idea to allow that, given that in reality it's not possible to deploy contracts with arbitrary IDs. However, there could be a case for specifying the exact ID preimage for the contract in order to get a stable ID. If we think there is a use case for that, I'm open to suggestions as to what would be the best interface for that functionality.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -353,14 +353,9 @@ struct ContractImportArgs {
 /// #[test]
 /// fn test() {
 ///     let env = Env::default();
-///
-///     // Define IDs for contract A and B.
-///     let contract_a_id = BytesN::from_array(&env, &[0; 32]);
-///     let contract_b_id = BytesN::from_array(&env, &[1; 32]);
-///
 ///     // Register contract A using the imported WASM.
-///     env.register_contract_wasm(&contract_a_id, contract_a::WASM);
-///
+///     let contract_a_id = env.register_contract_wasm(&contract_a_id, contract_a::WASM);
+///     let contract_b_id = BytesN::from_array(&env, &[1; 32]);
 ///     // Register contract B defined in this crate.
 ///     env.register_contract(&contract_b_id, ContractB);
 ///
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -354,7 +354,7 @@ struct ContractImportArgs {
 /// fn test() {
 ///     let env = Env::default();
 ///     // Register contract A using the imported WASM.
-///     let contract_a_id = env.register_contract_wasm(&contract_a_id, contract_a::WASM);
+///     let contract_a_id = env.register_contract_wasm(contract_a::WASM);
 ///     let contract_b_id = BytesN::from_array(&env, &[1; 32]);
 ///     // Register contract B defined in this crate.
 ///     env.register_contract(&contract_b_id, ContractB);
--- a/soroban-sdk/src/deploy.rs
+++ b/soroban-sdk/src/deploy.rs
@@ -40,7 +40,7 @@
 //! # #[cfg(not(feature = "testutils"))]
 //! # fn main() { }
 //! ```
-use crate::{env::internal::Env as _, Bytes, BytesN, Env, IntoVal, TryFromVal};
+use crate::{env::internal::Env as _, Bytes, BytesN, Env, IntoVal};
 
 /// Deployer provides access to deploying contracts.
 pub struct Deployer {
@@ -84,22 +84,6 @@ impl Deployer {
             salt: salt.into_val(env),
         }
     }
-
-    #[doc(hidden)]
-    /// Get a deployer for contracts that derive their contract IDs from the
-    /// given ed25519 public key and the provided salt.
-    pub fn with_ed25519(
-        &self,
-        public_key: impl IntoVal<Env, BytesN<32>>,
-        salt: impl IntoVal<Env, Bytes>,
-    ) -> DeployerWithEd25519 {
-        let env = self.env();
-        DeployerWithEd25519 {
-            env: env.clone(),
-            public_key: public_key.into_val(env),
-            salt: salt.into_val(env),
-        }
-    }
 }
 
 /// A deployer that deploys a contract that has its ID derived from the current
@@ -122,10 +106,12 @@ impl DeployerWithCurrentContract {
     /// will be used to derive a contract ID for the deployed contract.
     ///
     /// Returns the deployed contract's ID.
-    pub fn deploy(&self, wasm: impl IntoVal<Env, Bytes>) -> BytesN<32> {
+    pub fn deploy(&self, wasm_hash: impl IntoVal<Env, BytesN<32>>) -> BytesN<32> {
         let env = &self.env;
-        let id = env
-            .create_contract_from_contract(wasm.into_val(env).to_object(), self.salt.to_object());
+        let id = env.create_contract_from_contract(
+            wasm_hash.into_val(env).to_object(),
+            self.salt.to_object(),
+        );
         unsafe { BytesN::<32>::unchecked_new(id.in_env(env)) }
     }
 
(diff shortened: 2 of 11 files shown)


sisuresh (approved):

soroban-sdk-macros/src/lib.rs
```diff
@@ -353,14 +353,9 @@ struct ContractImportArgs {
...
 ///     let env = Env::default();
-///
-///     // Define IDs for contract A and B.
-///     let contract_a_id = BytesN::from_array(&env, &[0; 32]);
-///     let contract_b_id = BytesN::from_array(&env, &[1; 32]);
-///
 ///     // Register contract A using the imported WASM.
-///     env.register_contract_wasm(&contract_a_id, contract_a::WASM);
-///
+///     let contract_a_id = env.register_contract_wasm(&contract_a_id, contract_a::WASM);
```
leighmcculloch, line 357: `&contract_a_id` is still a parameter to the function. That looks like it is in error.

soroban-sdk/src/deploy.rs
```diff
@@ -22,11 +18,10 @@
...
 //! # impl Contract {
-//! #     pub fn f(env: Env) {
-//! # let wasm = [0u8; 32];
-//! # let salt = [0u8; 32];
-//! let deployer = env.deployer().with_current_contract(&salt);
-//! let contract_id = deployer.deploy(&wasm);
+//! #     pub fn f(env: Env, wasm_id: BytesN<32>) {
+//! #         let salt = [0u8; 32];
+//! #         let deployer = env.deployer().with_current_contract(&salt);
+//! #         let contract_id = deployer.deploy(&wasm_id);
```
leighmcculloch, line 24: These lines were uncommented so that they are visible in the Rust docs. Is there a reason they are being hidden? They're the lines we're demonstating to the user.
```suggestion
//! let deployer = env.deployer().with_current_contract(&salt);
//! let contract_id = deployer.deploy(&wasm_id);
```

soroban-sdk/src/testutils.rs
```diff
@@ -67,3 +70,21 @@ pub trait Accounts {
     /// Remove an account.
     fn remove(&self, id: &AccountId);
 }
+
+pub fn random_id() -> [u8; 32] {
```
leighmcculloch, line 74: What type of `id` is this intended to be? Or is this intended to be just a random 32-byte byte slice? It feels a bit arbitrary.

Instead of making this about "ids" how about we change it to:
```suggestion
pub fn random_id<const N: usize>() -> [u8; N] {
```

soroban-sdk/src/testutils.rs
```diff
@@ -67,3 +70,21 @@ pub trait Accounts {
...
+}
+
+pub fn random_account_id() -> xdr::AccountId {
+    xdr::AccountId(xdr::PublicKey::PublicKeyTypeEd25519(xdr::Uint256(
+        random_id(),
+    )))
+}
+
+impl Env {
+    pub fn random_id_bytes(&self) -> BytesN<32> {
```
leighmcculloch, line 87: Could you add Rust docs to these new functions?

leighmcculloch, line 87: Same question with this function, it is named somewhat arbitrary, since there are different types of IDs, and in the testutils API it isn't clear the scope this should be used for.

I suggest we change this to:
```suggestion
    pub fn random_bytes<const N: usize>(&self) -> BytesN<N> {
```

soroban-sdk/src/env.rs
```diff
@@ -439,26 +436,41 @@ impl Env {
...
-        contract_wasm: &[u8],
-    ) -> BytesN<32> {
-        let contract_id = if let Some(contract_id) = contract_id.into() {
-            contract_id.clone()
-        } else {
-            let mut contract_id = [0u8; 32];
-            rand::thread_rng().fill_bytes(&mut contract_id);
-            BytesN::from_array(self, &contract_id)
-        };
+    pub fn install_contract_wasm<'a>(&self, contract_wasm: &[u8]) -> BytesN<32> {
```
leighmcculloch, line 442: This lifetime looks unused.
```suggestion
    pub fn install_contract_wasm(&self, contract_wasm: &[u8]) -> BytesN<32> {
```

soroban-sdk/src/env.rs
```diff
@@ -439,26 +436,41 @@ impl Env {
...
+    /// use soroban_sdk::{BytesN, Env};
+    ///
+    /// const WASM: &[u8] = include_bytes!("../doctest_fixtures/contract.wasm");
+    ///
+    /// # fn main() {
+    /// let env = Env::default();
+    /// env.register_contract_wasm(WASM);
+    /// # }
+    /// ```
+    pub fn register_contract_wasm<'a>(&self, contract_wasm: &[u8]) -> BytesN<32> {
```
leighmcculloch, line 469:
```suggestion
    pub fn register_contract_wasm(&self, contract_wasm: &[u8]) -> BytesN<32> {
```

soroban-sdk/src/env.rs
```diff
@@ -475,24 +487,38 @@ impl Env {
...
+    /// env.register_contract_token();
     /// # }
     /// ```
-    pub fn register_contract_token<'a>(
-        &self,
-        contract_id: impl Into<Option<&'a BytesN<32>>>,
-    ) -> BytesN<32> {
-        let contract_id = if let Some(contract_id) = contract_id.into() {
-            contract_id.clone()
+    pub fn register_contract_token<'a>(&self) -> BytesN<32> {
```
leighmcculloch, line 493:
```suggestion
    pub fn register_contract_token(&self) -> BytesN<32> {
```

soroban-sdk/src/env.rs
```diff
@@ -475,24 +487,38 @@ impl Env {
...
+        self.register_contract_with_source(xdr::ScContractCode::Token)
+    }
+
+    fn register_contract_with_source<'a>(&self, source: xdr::ScContractCode) -> BytesN<32> {
```
leighmcculloch, line 497:
```suggestion
    fn register_contract_with_source(&self, source: xdr::ScContractCode) -> BytesN<32> {
```

soroban-sdk/src/env.rs
```diff
@@ -423,13 +419,14 @@ impl Env {
         contract_id
     }
 
-    /// Register a contract in a WASM file with the [Env] for testing.
+    /// Install the contract WASM code to the [Env] for testing.
     ///
-    /// Passing a contract ID for the first arguments registers the contract
-    /// with that contract ID. Providing `None` causes a random ID to be
-    /// assigned to the contract.
+    /// Returns the identifier of the installed code that can be then used for
```
leighmcculloch, line 424: We refer to the value returned as the "identifier" here, but in other places in this PR it is referred to as the "wasm_hash". We need to stick with one term for the thing otherwise it will be somewhat confusing to a developer. I suggest we stick with hash because we should expect people to be able to hash their contract file independently to find the value so it seems odd to abstract the concept. If it's not just a simple hash of the wasm though, identifier is better.

soroban-sdk/src/env.rs
```diff
@@ -423,13 +419,14 @@ impl Env {
...
+    /// the contract deployment.
     ///
-    /// Returns the contract ID of the registered contract.
+    /// This is mostly useful for contract factory testing, otherwise use
+    /// `register_contract_wasm` function that installs and deploys the contract
+    /// in a single call.
```
leighmcculloch, line 429: Avoid "This" in docs. It can be used in every paragraph/sentence and makes for more succinct explanations and docs.
```suggestion
    /// Useful for contract factory testing, otherwise use
    /// `register_contract_wasm` function that installs and deploys the contract
    /// in a single call.
```

leighmcculloch (commented):
Couple things to tweak, otherwise looks great.

A follow up, we should also remove the contract ID param from the other register function, because it's a bit odd that one has it and the other doesn't.

soroban-sdk-macros/src/lib.rs
dmkozh: Fixed.

soroban-sdk/src/deploy.rs
dmkozh: Oh, I just thought that was unintentionally incorrect formatting, I didn't know that we show only a snippet of the example...

soroban-sdk/src/env.rs
dmkozh: Updated to use hash everywhere.

soroban-sdk/src/env.rs
dmkozh: Updated.

soroban-sdk/src/env.rs
dmkozh: Removed.

soroban-sdk/src/env.rs
dmkozh: Removed.

soroban-sdk/src/testutils.rs
dmkozh: Updated.

soroban-sdk/src/testutils.rs
dmkozh: Added.

dmkozh: Updated.

soroban-sdk/src/deploy.rs
leighmcculloch: It's a pretty cool feature of rust docs. It helps to shave our examples down while keeping them fully functional.

leighmcculloch (approved):

soroban-sdk/src/testutils.rs
```diff
@@ -67,3 +70,24 @@ pub trait Accounts {
...
+pub fn random_account_id() -> xdr::AccountId {
+    xdr::AccountId(xdr::PublicKey::PublicKeyTypeEd25519(xdr::Uint256(
+        random_bytes_array(),
+    )))
+}
+
+impl Env {
+    /// Generates a Host-owned array of `N` random bytes.
+    pub fn random_bytes<const N: usize>(&self) -> BytesN<N> {
+        BytesN::from_array(self, &random_bytes_array())
```
leighmcculloch, line 91: I just realized there's a couple things inconsistent with the placement of these functions with the way other test utilities are structured, so I'm going to follow up with addressing that. The issues are:
- One of the functions is exposing XDR types, but we try and expose only SDK types if we can. All SDK types are convertible to XDR types anyway, and it prevents users getting confused about what types they can or can't use in different situations.
- We avoid putting functions like "random account id" on Env, and put them on the type itself, so that the Env isn't like an everything object of sorts.

leighmcculloch: Follow up is in https://github.com/stellar/rs-soroban-sdk/pull/767.

PR #767, opened 2022-11-29
Reorganize recently added random functions

leighmcculloch:
### Why
Most other test utilities relating to types are on the types themselves on a testutils only trait. These random functions were just added but we added them to the global and Env namespaces. The Env namespace will have a tendency to become a default space to put everything. While some thing should belong there, testutilities relating to types are better namespaced on their own type using the existing pattern to avoid crowding.

### DIFF
--- a/soroban-sdk/src/accounts.rs
+++ b/soroban-sdk/src/accounts.rs
@@ -204,6 +204,14 @@ impl TryFrom<AccountId> for ScVal {
 }
 
 #[cfg(not(target_family = "wasm"))]
+impl TryFrom<AccountId> for super::xdr::AccountId {
+    type Error = ConversionError;
+    fn try_from(v: AccountId) -> Result<Self, Self::Error> {
+        (&v).try_into()
+    }
+}
+
+#[cfg(not(target_family = "wasm"))]
 impl TryFromVal<Env, ScVal> for AccountId {
     type Error = ConversionError;
     fn try_from_val(env: &Env, val: ScVal) -> Result<Self, Self::Error> {
@@ -262,6 +270,18 @@ impl AccountId {
 
     pub fn to_object(&self) -> Object {
         self.0.to_object()
+    }
+}
+
+#[cfg(feature = "testutils")]
+#[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
+impl crate::testutils::AccountId for AccountId {
+    fn random(env: &Env) -> AccountId {
+        xdr::AccountId(xdr::PublicKey::PublicKeyTypeEd25519(xdr::Uint256(
+            crate::testutils::random(),
+        )))
+        .try_into_val(env)
+        .unwrap()
     }
 }
 
@@ -373,7 +393,7 @@ impl Account {
 }
 
 #[cfg(any(test, feature = "testutils"))]
-use crate::testutils;
+use crate::testutils::{self, AccountId as _};
 
 #[cfg(any(test, feature = "testutils"))]
 #[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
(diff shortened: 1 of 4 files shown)


soroban-sdk/src/accounts.rs
```diff
@@ -265,6 +273,18 @@ impl AccountId {
     }
 }
 
+#[cfg(feature = "testutils")]
+#[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
+impl crate::testutils::AccountId for AccountId {
+    fn random(env: &Env) -> AccountId {
```
leighmcculloch, line 279: Turns out we already had afunction for this, `Accounts::generate`, so I moved that logic into a random functions on `AccountId`, and that replaces the `crate::testutils::random_account_id`.

soroban-sdk/src/bytes.rs
```diff
@@ -1066,6 +1066,14 @@ impl<const N: usize> BytesN<N> {
     }
 }
 
+#[cfg(feature = "testutils")]
+#[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
+impl<const N: usize> crate::testutils::BytesN<N> for BytesN<N> {
+    fn random(env: &Env) -> BytesN<N> {
```
leighmcculloch, line 1072: `BytesN::random` replaces `crate::testutils::random_bytes`.

soroban-sdk/src/env.rs
```diff
@@ -495,19 +497,18 @@ impl Env {
...
         };
-        self.env_impl.set_source_account(random_account_id());
+        self.env_impl
+            .set_source_account(AccountId::random(self).try_into().unwrap());
 
         let contract_id: BytesN<32> = self
             .env_impl
             .invoke_function(xdr::HostFunction::CreateContract(xdr::CreateContractArgs {
-                contract_id: xdr::ContractId::SourceAccount(xdr::Uint256(random_bytes_array())),
+                contract_id: xdr::ContractId::SourceAccount(xdr::Uint256(random())),
```
dmkozh, line 511: nit: I think unlike the other changed functions (which are both more concise and clear) this one is ambiguous. Since it doesn't belong to any 'namespace'/struct, it's hard to tell what exactly does it generate, so I would probably go with something more specific (like the previous name or at least `random_array` or something like that).

soroban-sdk/src/testutils.rs
```diff
@@ -48,46 +45,54 @@ pub trait Logger {
...
+    // use in tests.
+    fn random(env: &Env) -> crate::BytesN<N>;
 }
 
-/// Generates a Rust array of N random bytes.
-pub fn random_bytes_array<const N: usize>() -> [u8; N] {
+/// Generates an array of N random bytes.
+///
+/// The value returned is not cryptographically secure and is only intended for
+/// use in tests.
```
dmkozh, line 92: nit: I think crypto-security comment is appropriate, but the 'only intended for use in tests' seems redundant, given that all these functions belong to the test utils IIUC and can't be used inside the contracts. Up to you though.

dmkozh (approved):
Looks good, just a couple naming/doc nitpicks.

soroban-sdk/src/testutils.rs
leighmcculloch: I just want to emphasise that someone shouldn't import testutils to get this capability into other applications for other uses. Just being overly cautious.

leighmcculloch: Updated in ec85e79.

soroban-sdk/src/env.rs
leighmcculloch: I'm in two minds about whether we should even expose this function at this stage. I think I'd rather folks import a RNG that they choose to use in their applications, so I'll make `random()` `pub(crate)` so it isn't part of the public API.

leighmcculloch: I guess we have random on types now, so exposing this here is not that big of a deal.

dmkozh: Ok, that's probably fine then, given that the function isn't public.

leighmcculloch: I'll remove it, then lets see how much a need continues to arise for it. Most uses of random in contract tests appear to be people generating contract IDs, but that's all going away.

leighmcculloch: Removed in d11aad01.

sisuresh (approved):

PR #768, opened 2022-11-29
Make `contract_id` public in contract clients.

dmkozh:
### What

Make `contract_id` public in contract clients.

### Why

That's useful e.g. in testing when we want to pass the contract client around, but we also need to its identifier. The current workaround is to pass around a tuple, which is inconvenient.

### Known limitations

N/A

### DIFF
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -187,7 +187,7 @@ pub fn derive_client(name: &str, fns: &[ClientFn]) -> TokenStream {
     quote! {
         pub struct #client_ident {
             env: soroban_sdk::Env,
-            contract_id: soroban_sdk::BytesN<32>,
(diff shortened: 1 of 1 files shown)


soroban-sdk-macros/src/derive_client.rs
```diff
@@ -187,7 +187,7 @@ pub fn derive_client(name: &str, fns: &[ClientFn]) -> TokenStream {
     quote! {
         pub struct #client_ident {
             env: soroban_sdk::Env,
-            contract_id: soroban_sdk::BytesN<32>,
+            pub contract_id: soroban_sdk::BytesN<32>,
             #[cfg(any(test, feature = "testutils"))]
             source_account: Option<soroban_sdk::AccountId>,
```
leighmcculloch, line 192: I think all the fields could be made `pub`.

leighmcculloch (approved):

leighmcculloch (approved):

PR #774, opened 2022-12-01
Add back ability to register wasm contracts at any id

leighmcculloch:
### What
Add back ability to register wasm contracts at any id

### Why
To avoid breaking the API when we need the ability in the future to register wasm contracts at specific addresses.

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -548,7 +548,7 @@ impl Env {
                         &key,
                         &LedgerEntry {
                             ext: xdr::LedgerEntryExt::V0,
-                            last_modified_ledger_seq: 0, // TODO: Decide on what value to place here.
+                            last_modified_ledger_seq: 0,
                             data: xdr::LedgerEntryData::ContractData(xdr::ContractDataEntry {
                                 contract_id: hash.clone(),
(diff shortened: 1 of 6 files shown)


soroban-sdk/src/env.rs
```diff
@@ -528,14 +530,42 @@ impl Env {
     ///
     /// # fn main() {
     /// let env = Env::default();
-    /// env.register_contract_wasm(WASM);
+    /// env.register_contract_wasm(None, WASM);
```
dmkozh, line 533: Could you please add a comment that this replaces the code in case if it already exists? I don't think it's obvious from the name (same for another variant of this function).

dmkozh (approved):

soroban-sdk/src/env.rs
leighmcculloch: I'm going to leave the comment out, because I want to write more tests that actually confirm that replacing works, but this works given current behavior expected as is, so I'd like to merge it now for this release, and circle back to improve the testing.

i.e. This restores the existing API and existing behavior, and provides some potential new behavior that the docs aren't committing to (replacing).

I'll follow up this PR with tests around replacing.

tomerweller (approved):

PR #789, opened 2022-12-07
Add ability to get current Budget from env in tests

leighmcculloch:
### What
Add ability to get current Budget from env in tests.

### Why
It would be helpful for developers to be able to query their resource usage inside tests. While some things won't be completely equal to a WASM contract running, such as the WASM instruction count, other costs are measurable.

@tyvdh mentioned this would be useful for upcoming quests.

### DIFF
--- a/soroban-sdk/src/tests/budget.rs
+++ b/soroban-sdk/src/tests/budget.rs
@@ -20,7 +20,8 @@ fn test_budget() {
 
     e.budget().reset();
     let b = client.add();
-    println!("{}", e.budget());
+    e.budget().print();
+
     assert_eq!(e.budget().input(CostType::BytesAppend), 36);
     assert_eq!(
         b,
--- a/soroban-sdk/src/tests/budget.rs
+++ b/soroban-sdk/src/tests/budget.rs
@@ -1,12 +1,14 @@
 use crate as soroban_sdk;
-use soroban_sdk::{contractimpl, testutils::CostType, Env};
+use soroban_sdk::{contractimpl, testutils::budget::CostType, Bytes, Env};
 
 pub struct Contract;
 
 #[contractimpl]
 impl Contract {
-    pub fn add(a: i32, b: i32) -> i32 {
-        a + b
+    pub fn add(e: Env) -> Bytes {
+        let mut b = Bytes::from_array(&e, b"abcdefghijklmnopqrstuvwyxz");
+        b.append(&Bytes::from_array(&e, b"0123456789"));
+        b
     }
 }
 
@@ -16,16 +18,12 @@ fn test_budget() {
     let contract_id = e.register_contract(None, Contract);
     let client = ContractClient::new(&e, &contract_id);
 
-    let a = 10i32;
-    let b = 12i32;
-    e.reset_budget();
-    let c = client.add(&a, &b);
-    assert_eq!(c, 22);
-
-    let budget = e.budget();
-    eprintln!("Cpu Insns: {}", budget.get_cpu_insns_count());
-    eprintln!("Mem Bytes: {}", budget.get_mem_bytes_count());
-    for cost_type in CostType::variants() {
-        eprintln!("Cost ({cost_type:?}): {}", budget.get_input(*cost_type));
-    }
+    e.budget().reset();
+    let b = client.add();
+    println!("{}", e.budget());
+    assert_eq!(e.budget().get(CostType::BytesAppend), 36);
+    assert_eq!(
+        b,
+        Bytes::from_array(&e, b"abcdefghijklmnopqrstuvwyxz0123456789")
+    );
 }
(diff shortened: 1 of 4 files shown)


soroban-sdk/src/testutils.rs
```diff
@@ -29,6 +31,70 @@ pub trait Ledger {
...
+    /// # }
+    /// # #[cfg(not(feature = "testutils"))]
+    /// # fn main() { }
+    /// ```
+    pub struct Budget(crate::env::internal::budget::Budget);
+
+    impl Display for Budget {
+        fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            writeln!(f, "CPU Instructions: {}", self.get_cpu_instruction_count())?;
+            writeln!(f, "Memory Bytes: {}", self.get_memory_bytes_count())?;
```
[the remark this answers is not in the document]
jayz22: Any work that is done by via the host (calling host function, accessing storage etc) are accurate. The main difference when running non-WASM (native) contract is the `WasmInsnExec`, `WasmMemAlloc`  and `Vm*` ones will not be counted, because there is no Vm running. Therefore the cost numbers will underestimate the actual contract running cost. But given most of the heavy lifting is done by the host anyway, they should be close enough to give the user some sense of the actual cost.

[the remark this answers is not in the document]
leighmcculloch: I'll add notes about this in the docs.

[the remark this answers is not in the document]
leighmcculloch: I added some docs explaining this in 51cb8fe3.

soroban-sdk/src/testutils.rs
```diff
@@ -29,6 +31,83 @@ pub trait Ledger {
...
+        /// code compared to running the WASM equivalent.
+        pub fn get_memory_bytes_count(&self) -> u64 {
+            self.0.get_mem_bytes_count()
+        }
+
+        /// Get other cost counts.
+        ///
+        /// Note that VM cost types are likely to be underestimated when running
+        /// Rust code compared to running the WASM equivalent.
+        pub fn get(&self, cost_type: CostType) -> u64 {
```
jayz22, line 105: I would name this `get_input` to be consistent and more explicit, since the input is different from the cost. The comment needs to be updated too.
The input is what goes into the budget model to produce the cost (which has two dimensions  "cpu instructions" and "memory bytes"). For example in a vector clone, the input is the number of elements of the vector and the cost is the cpu instructions and mem bytes allocated to perform the vector clone action.

soroban-sdk/src/testutils.rs
```diff
@@ -29,6 +31,83 @@ pub trait Ledger {
...
 
+pub mod budget {
+    use core::fmt::Display;
+
+    #[doc(inline)]
+    pub use crate::env::internal::budget::CostType;
+
+    /// Budget that tracks the resources consumed for the environment.
+    ///
+    /// Note that some budgets – CPU instructions, memory, and VM cost types –
```
jayz22, line 42: This comment can be more clarified. The budget consist of two dimensions – CPU instructions, memory bytes. Each dimension consist of cost types, the WASM/VM-related cost types are the ones being omitted(underestimated).

soroban-sdk/src/tests/budget.rs
```diff
@@ -0,0 +1,29 @@
+use crate as soroban_sdk;
+use soroban_sdk::{contractimpl, testutils::budget::CostType, Bytes, Env};
+
+pub struct Contract;
+
+#[contractimpl]
+impl Contract {
+    pub fn add(e: Env) -> Bytes {
+        let mut b = Bytes::from_array(&e, b"abcdefghijklmnopqrstuvwyxz");
```
jayz22, line 9: What is this bytes object?

soroban-sdk/src/testutils.rs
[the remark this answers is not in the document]
silence48: would it be possible to detect when it is running non-wasm contract and print a notice to the user?

soroban-sdk/src/tests/budget.rs
silence48: looks like it is a filler for the test

leighmcculloch: > What is this bytes object?

It is the SDK Bytes type.

soroban-sdk/src/testutils.rs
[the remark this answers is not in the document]
leighmcculloch: I don't think so. Also, a test might interact with both wasm and rust invocations. They can be mixed.

soroban-sdk/src/testutils.rs
leighmcculloch: Thanks I updated the docs.

soroban-sdk/src/testutils.rs
leighmcculloch: I renamed it to input. Thanks for the explanation, this helped a lot.

jayz22 (approved):

soroban-sdk/src/tests/budget.rs
jayz22: Oh I see, this is just testing `BytesAppend`. Makes sense.

graydon (approved):

PR #793, opened 2022-12-13
Auth concept v1

dmkozh:
Auth concept v1


soroban-sdk/src/account.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    pub fn env(&self) -> &Env {
+        self.0.env()
+    }
+
+    pub fn address(&self) -> Address {
+        self.0.env().get_account_address(&self)
+    }
+
+    pub fn authorize(&self, args: Vec<RawVal>) {
```
leighmcculloch, line 142: What does this function do? It is unclear.

soroban-sdk/src/env.rs
```diff
@@ -127,28 +111,28 @@ impl Env {
...
     #[inline(always)]
     pub fn deployer(&self) -> Deployer {
         Deployer::new(self)
     }
 
-    /// Get the 32-byte hash identifier of the current executing contract.
-    pub fn current_contract(&self) -> BytesN<32> {
-        let id = internal::Env::get_current_contract(self);
-        unsafe { BytesN::<32>::unchecked_new(id.in_env(self)) }
+    pub fn current_contract_account(&self) -> Account {
```
leighmcculloch, line 120: What does it mean to be able to access the "current contract account". Do all contracts have accounts?

soroban-sdk/src/env.rs
```diff
@@ -522,6 +488,21 @@ impl Env {
         }
         contract_id
     }
+
+    pub fn verify_account_authorization(
```
leighmcculloch, line 492: When would a contract need to manually verify account authorization, vs relying on the system to verify an invocation or not?

soroban-sdk/src/testutils.rs
```diff
@@ -45,37 +45,6 @@ pub trait Logger {
...
-    /// Generate a random account ID.
-    //
-    // The value filled is not cryptographically secure.
-    fn random(env: &Env) -> crate::AccountId;
-}
-
-/// Test utilities for [`Accounts`][crate::accounts::Accounts].
-pub trait Accounts {
-    /// Generate an account ID.
-    fn generate(&self) -> crate::AccountId;
```
leighmcculloch, line 59: I assume we'd circle back on adding this before merging, but for testing there would need to be a way to generate distinct accounts / addresses that exist and don't exist, and to also use them for signing.

soroban-sdk/src/account.rs
dmkozh: Sorry, I spend much time on docs, given that this is a prototype/iteration. That's the main authorization function: it authorizes a call of the current contract function with the given args and with the current authorized call stack (see examples). More strictly, calling `authorize` requires the account to have an invocation in signature payload that consists of the call stack of all the  `(contract_id, fn_name)` pairs that have called `authorize` for this account, and `args` provided by the current contract.
Basically, if you want to make an account to have authorized your contract's invocation, you call this. That's all the contract needs to do for auth. Everything else is on host/account contract.

soroban-sdk/src/env.rs
dmkozh: Yeah, this is an `Account` object corresponding to the currently running contract. As all the invocations now require `Account` there needs a way to do something on behalf of the currently running contract. This is functionally equivalent to the contract invoker now, but unlike invoker it's extensible (e.g. it's easy to allow contract to safely authorize subcontract calls if needed).

soroban-sdk/src/env.rs
dmkozh: This is a testutils-only function (which is a bit tricky to notice from the diff). The idea is that tests have auth in the recording mode. Then it's possible to write the tests without worrying about auth at all. But, of course, it's important to cover the authorizations of your own contract at least once, hence we provide a function like this one that allows to check that an account would need to authorize something. See the example tests for the examples of using this function.

soroban-sdk/src/testutils.rs
dmkozh: Yeah, we may want this in some fashion, but for now I've just used `Account::random()`. I'm not sure we need more though; the regular contracts don't need to worry about the existence of an account and signatures are just not needed for the regular contract testing (which I think is a big win). Maybe wallet contracts would need more specific test utils; not sure about that yet as that's still WIP.

soroban-sdk/src/account.rs
leighmcculloch: I'm still catching up on how this works so I might be way off, and I suspect I'm misunderstanding how this works, so please bear with me!

It's a little confusing that this function returns nothing. If the function is just loading up the authorizations that were in the invocation, why can't the host do that automatically when it receives the invocation?
Otherwise, what if a contract forgets to call it, or is tricked into not calling it. And if it is always the first thing a contract calls, we could just do it host side before calling any WASM.

If there was a function to call for verifying auth, I expect it would give me what was authenticated, like the identity, or ID, or something.

dmkozh: > It's a little confusing that this function returns nothing. 

Currently it would just panic if authorization fails in enforcing mode. It could maybe return `bool/Result`, but I'm not convinced if it's ever a good idea to gracefully handle the authorization failure...

> Otherwise, what if a contract forgets to call it, 

Then no authorization happens. That's up to the contract basically. Nothing prevents the contract from e.g. passing just an `Address` instead of `Account` etc. I'm not sure if it's possible to make this completely fool-proof without seriously compromising flexibility.

> And if it is always the first thing a contract calls, we could just do it host side before calling any WASM

We don't know the args beforehand and we may not even always want to call this and consume nonce. We may add some SDK sugar (maybe with some host backing) for the simple cases where there is just a single account for which we need to authorize the invocation with all the function args. This draft currently only implements the generic case that allows passing the accounts through and passing the custom args to `authorize` (see atomic swap/multiswap examples).

Something like the previous roles concept is not out of the question, but I'm not sure how to make it non-confusing.

dmkozh: >  I'm not convinced if it's ever a good idea to gracefully handle the authorization failure.

I've just realized that `try_call` kind of allows that anyway. So maybe we should really make this return result or add `try_authorize`... I'll think more about that.

PR #794, opened 2022-12-13
Update env to include delete-im, remove-EnvVal changes

jayz22:
### What

Update env to include delete-im, remove-EnvVal changes

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-sdk/src/accounts.rs
+++ b/soroban-sdk/src/accounts.rs
@@ -6,7 +6,6 @@ use core::{cmp::Ordering, fmt::Debug};
 use crate::{
     env::internal::xdr,
     env::internal::{Env as _, EnvBase as _, RawVal, RawValConvertible},
-    env::EnvObj,
     BytesN, ConversionError, Env, IntoVal, Object, TryFromVal, TryIntoVal,
 };
 
@@ -76,7 +75,10 @@ impl Accounts {
 ///
 /// In tests account identifiers can be generated using [`Accounts`].
 #[derive(Clone)]
-pub struct AccountId(EnvObj);
+pub struct AccountId {
+    env: Env,
+    obj: Object,
+}
 
 impl Debug for AccountId {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
@@ -111,18 +113,9 @@ impl PartialOrd for AccountId {
 
 impl Ord for AccountId {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        self.0.env.check_same_env(&other.0.env);
-        let v = self
-            .0
-            .env
-            .obj_cmp(self.0.obj.to_raw(), other.0.obj.to_raw());
-        if v == 0 {
-            Ordering::Equal
-        } else if v < 0 {
-            Ordering::Less
-        } else {
-            Ordering::Greater
-        }
+        self.env.check_same_env(&other.env);
+        let v = self.env.obj_cmp(self.obj.to_raw(), other.obj.to_raw());
+        v.cmp(&0)
     }
 }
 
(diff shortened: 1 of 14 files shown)


soroban-sdk/src/accounts.rs
```diff
@@ -111,16 +111,30 @@ impl PartialOrd for AccountId {
...
+            .0
+            .env
+            .obj_cmp(self.0.obj.to_raw(), other.0.obj.to_raw());
+        if v == 0 {
+            Ordering::Equal
+        } else if v < 0 {
+            Ordering::Less
+        } else {
+            Ordering::Greater
+        }
```
leighmcculloch, line 125: I believe this is the same thing:
```suggestion
        v.cmp(&0)
```

soroban-sdk/src/accounts.rs
```diff
@@ -421,7 +435,11 @@ impl testutils::Accounts for Accounts {
                     last_modified_ledger_seq: 0,
                     ext: xdr::LedgerEntryExt::V0,
                 };
-                storage.put(&k, &v)
+                storage.put(
+                    &k,
+                    &v,
+                    soroban_env_host::budget::AsBudget::as_budget(env.host()),
```
leighmcculloch, line 441: What does as budget do in these calls now?

soroban-sdk/src/env.rs
```diff
@@ -49,8 +50,17 @@ pub use internal::TryFromVal;
...
+pub struct EnvVal {
+    pub env: Env,
+    pub val: RawVal,
+}
+
+#[derive(Clone)]
+pub struct EnvObj {
+    pub env: Env,
+    pub obj: Object,
+}
```
leighmcculloch, line 63: I think we can get rid of these types, at least that's what I imagined we'd do, and just put Env and Object inside each types like Bytes, Vec, etc. This is fine too though.

leighmcculloch (approved):
Small suggestions inline, otherwise lgtm.

I would consider getting rid of the EnvVal and EnvObj types. The goal was in part to get rid of unnecessary concepts, but this PR seems to attempt to maintain those two concepts, which I think are unnecessary and I regret adding them. Because they just make one more thing to learn when navigating this code.

jayz22:
> I would consider getting rid of the EnvVal and EnvObj types. 

I would love so as well. I already removed EnvVal.
However, when I replace EnvObj (replacing with `Env` + `Object`) for `Map` and `Vec`: 
`pub struct Map<K, V>(Env, Object, PhantomData<K>, PhantomData<V>);`
`repr(transparent)` was giving me problem "transparent struct needs at most one non-zero-sized field, but has 2". Is it ok that we remove `repr(transparent)`?

leighmcculloch:
> Is it ok that we remove `repr(transparent)`?

I think yes. The attributes guarantees that the ABI of the inner value and struct are the same, so that you could transmute between them. More details [here](https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent). We only need it on types that cross extern/exported functions, and the only types we have that do that are `RawVal` and `Object`. So it is critical that we keep it on those types, but other types no.

soroban-sdk/src/accounts.rs
jayz22: Logically same as before. It's just now a `Budget` is explicitly passed in every storage call, whereas previously the `Storage` owns a `MeteredOrdMap` which owns a `Budget` reference (same one as the host owns). Internally it charges the budget before preforming the corresponding map operation.

soroban-sdk/src/env.rs
jayz22: I have removed both of them.

leighmcculloch (approved):
🎉 👏🏻

PR #802, opened 2022-12-15
Add LedgerSnapshot::update(Host)

### LINKED ISSUE #799: Add fn to ledger snapshot to update from a Host
The `LedgerSnapshot` type can be updated piecemeal by calling `set_ledger_info` and `update_entries`.

@sisuresh pointed out that it would be nice if we could update both in a single call:
- https://github.com/stellar/soroban-tools/pull/309#discussion_r1049045796

There are some advantages to this, mostly around preventing foot-guns where you update one and not the other.

We could have the ledger snapshot type entirely own what it means to snapshot the ledger, by giving it a `&Host` and letting it get the appropriate data out.

This new fn would ideally be used in the CLI.

leighmcculloch:
### What
Add LedgerSnapshot::update function that will update a snapshot with the state within a Host.

### Why
The `LedgerSnapshot` type can be updated piecemeal by calling `set_ledger_info` and `update_entries`.

@sisuresh pointed out that it would be nice if we could update both in a single call:
- https://github.com/stellar/soroban-tools/pull/309#discussion_r1049045796

There are some advantages to this, mostly around preventing foot-guns where you update one and not the other.

Close #799

### DIFF
--- a/soroban-ledger-snapshot/src/lib.rs
+++ b/soroban-ledger-snapshot/src/lib.rs
@@ -8,7 +8,7 @@ use std::{
 use soroban_env_host::{
     storage::SnapshotSource,
     xdr::{LedgerEntry, LedgerKey, ScHostStorageErrorCode, ScStatus},
-    HostError, LedgerInfo,
+    Host, HostError, LedgerInfo,
 };
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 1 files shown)


sisuresh (approved):
LGTM. Thanks!

leighmcculloch:
> LGTM. Thanks!

No problem! It turned out better than I expected, and will make for less footguns in the CLI. 👏🏻

PR #815, opened 2023-01-04
Remove {Try}FromVal

jayz22:
### What

This is the sdk side changes necessary for https://github.com/stellar/rs-soroban-env/pull/623 which is a rebased version of https://github.com/stellar/rs-soroban-env/pull/600.

This won't work because it relies on the conversion between a `Result` and `RawVal`, which is commented out in  https://github.com/jayz22/rs-soroban-env/blob/aae7e792518ccd3f086586263263944ccf3d6e4c/soroban-env-common/src/lib.rs#L51-L54 due to the infinite recursion issue. 

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-sdk/src/map.rs
+++ b/soroban-sdk/src/map.rs
@@ -3,9 +3,9 @@ use core::{cmp::Ordering, fmt::Debug, iter::FusedIterator, marker::PhantomData};
 use crate::iter::{UncheckedEnumerable, UncheckedIter};
 
 use super::{
-    env::internal::{Env as _, EnvBase as _, RawValConvertible},
+    env::internal::{try_convert_to, Env as _, EnvBase as _, RawValConvertible},
     xdr::ScObjectType,
-    ConversionError, Env, IntoVal, Object, RawVal, Status, TryFromVal, TryIntoVal, Vec,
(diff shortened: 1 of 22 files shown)


soroban-sdk/src/map.rs
```diff
@@ -428,30 +429,36 @@ where
     #[inline(always)]
     pub fn iter_unchecked(&self) -> UncheckedIter<MapIter<K, V>, (K, V), ConversionError>
     where
-        K: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal> + Clone,
-        K::Error: Debug,
-        V: IntoVal<Env, RawVal> + TryFromVal<Env, RawVal> + Clone,
-        V::Error: Debug,
+        K: IntoVal<Env, RawVal> + Clone,
+        <RawVal as TryIntoVal<Env, K>>::Error: Debug,
```
jayz22, line 433: I don't like how every "T::Error" now have to be expressed as
"<RawVal as TryIntoVal<Env, K>>::Error", not sure if Rust allows a better way to define this type for an entire struct.

jayz22: Need something like the [inherent associated type](https://github.com/rust-lang/rust/issues/8995).

PR #817, opened 2023-01-05
Add alloc

leighmcculloch:
### What
Add grow only allocator behind an `alloc` feature.

### Why
Give folks more freedom.

To let folks use types needing an allocator, use a simple grow-only allocator.

This allocator was written by @graydon, but then removed in 04ad098e35b3cabb9b93f603c68c7c8e56835311 when we were removing dead code that wasn't in use and that we thought needed nightly. It turns out it doesn't need nightly so lets bring it back to give folks freedom.

The std crate is also included to get the default alloc error handler from the std crate. That default handler is heavier than one we could write, but we can't override it with our own without nightly. In the interest of making the SDK have no dependencies on nightly I compromised with using the std crates handler. The std crate is only imported when the `alloc` feature is in use.


tomerweller (approved):
FREEDOOOOOM!

graydon (approved):
Thanks for the careful iteration on this, glad to be arriving at something interesting and hopefully non-catastrophic for code size!

PR #824, opened 2023-01-13
Adapt SDK to changes to {Try,}{From,Into}Val in env crates.

graydon:
Adapts the SDK to changes in https://github.com/stellar/rs-soroban-env/pull/628 -- this variant finally worked ok. Very minor warts, lots of code removal.

### DIFF
--- a/soroban-sdk-macros/src/derive_enum.rs
+++ b/soroban-sdk-macros/src/derive_enum.rs
@@ -21,7 +21,7 @@ pub fn derive_type_enum(
     let mut errors = Vec::<Error>::new();
 
     let variants = &data.variants;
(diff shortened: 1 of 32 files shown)


soroban-sdk-macros/src/derive_enum.rs
```diff
@@ -163,14 +163,6 @@ pub fn derive_type_enum(
             }
         }
 
-        impl #path::TryIntoVal<#path::Env, #enum_ident> for #path::RawVal {
```
leighmcculloch, line 166: I think we still need the try into val's for cases where we assume every type has the try into val? Or does this get generated by a blanket?

leighmcculloch (approved):

soroban-sdk-macros/src/derive_enum.rs
graydon: TryIntoVal always comes from the blanket TryIntoVal for TryFromVal.

leighmcculloch: Sounds good to me.

PR #827, opened 2023-01-13
Fix vec insert

jayz22:
### What

Fix `Vec::insert`. @tyvdh found out the bug that `Vec::insert` was actually doing what `Vec::set` is supposed to. 

### Why

### Known limitations

[TODO or N/A]


leighmcculloch (approved):

leighmcculloch:
Thanks @jayz22 !

PR #830, opened 2023-01-17
Fix vec pop_front

mootz12:
### What

@markuspluna noticed a bug with `pop_front` where it returns the last result instead of the first. This PR fixed that and adds additional unit tests.

There is an outstanding deviation with `remove` for SDK implementation and the [Rust std version](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove). The SDK implementation does not return the item that was removed. Let me know if this was on purpose, otherwise being the SDK version to parity in this PR as well.

### Why

To fix the bug

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/vec.rs
+++ b/soroban-sdk/src/vec.rs
@@ -477,7 +477,7 @@ where
 
     #[inline(always)]
     pub fn pop_front(&mut self) -> Option<Result<T, T::Error>> {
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

leighmcculloch:
@mootz12 Thanks for finding this bug and fixing it!

PR #833, opened 2023-01-20
Adapt to env-crate changes: fallible Env and TryFromVal using Env::Error

graydon:
This is the SDK-side work adapting to https://github.com/stellar/rs-soroban-env/pull/633

It does two things simultaneously that, after some reflection, I'm not 100% sure should both be done. But I'm maybe 80% sure. The first should be done and the second should _probably_ be done, but we can discuss!

1. Changes `Env` to always return `Result<T, Env::Error>`. There's no more `CheckedEnv` at all, and `Env::Error` changes between `HostError` (when compiled natively `Env=Host`) and `Infallible` (when compiled for wasm `Env=Guest`). This is just part of the general "no panics in the host" work aiming to fix https://github.com/stellar/rs-soroban-env/issues/579 once and for all. As a result of this, there are a lot of unwraps in the SDK now. This is intentional, so they don't happen by accident / automatically in the previously-existing `impl Env for CheckedEnv` (which is being deleted upstream). They're all the same points where that wrapper would have panicked before, but now they're explicit. In the wasm case they should compile to no-ops because the VM will actually trap before it returns -- hence `Infallible` -- but the API still says `Result<>` on the label.
2. Changes `TryFromVal` to no longer have its own `Error`, but to always return `Result<Self, Env::Error>`. This is more contentious. It means that when someone uses `TryFromVal` in wasm code and there's some fallible-in-the-guest conversion code, it will escalate its error to an `Env::Error` before it completes, making the error behave _as if_ it was generated in the environment (and in particular, again, trapping the VM immediately rather than returning an error that a caller can scrutinize). This has some advantages: it makes the APIs less noisy (only one error type most of the time), it makes the resulting errors _richer_ when in native local testing mode (escalating to an `Env::Error=HostError` will capture a backtrace) and it .. sorta makes sense in that `TryFromVal` is explicitly the environment-assisting conversion path. If you want a non-environment-assisting conversion you can use `From` or `TryFrom` . Any time  you call `TryFromVal` you should be expecting a possibly-`Env::Error`-generating path anyways. So your result type should at most be a superset of `Env::Error`-plus-some-other-less-serious-errors?  But that can't be meaningfully represented, so `Env::Error` feels to me like a logical escalation type. But it does mean that paths that return-a-user-visible-result in local testing will trap the VM in deployment. Which is a bit weird.

One followup we might do, if we do go with the second point here, is to strip the `Result<>` returns out of some of the SDK methods that do a `TryFromVal` internally, such as `Map::get` or such, because when building for wasm they'll be returning `Result<T, Infallible>` and this is generally fairly confusing to users as a type altogether -- it can't be converted to or from much else and scrutinizing it produces dead code paths and such. So I think it'l behave nicer if we unwrap those inside the SDK methods as well. I think this is _probably_ what 99% of user code will do anyways. An ambitious user can always program against the `RawVal` API themselves -- try to do some not-trapping parts of their conversion if they want to inspect a value in the `Guest` -- but I expect nobody will do that in almost any code.

Feedback welcome!

## Code size impact

Code size changes are mixed. Some larger, some smaller. Mostly negligible. It does slightly better in the codesize-optimized cases.

### Non-size-optimized

Before:
~~~
-rwxrwxr-x 2 graydon graydon 21386 Jan 21 23:14 soroban_token_spec.wasm
-rwxrwxr-x 2 graydon graydon 505 Jan 21 23:14 test_add_i128.wasm
-rwxrwxr-x 2 graydon graydon 506 Jan 21 23:14 test_add_u128.wasm
-rwxrwxr-x 2 graydon graydon 427 Jan 21 23:14 test_add_u64.wasm
-rwxrwxr-x 2 graydon graydon 2928 Jan 21 23:14 test_alloc.wasm
-rwxrwxr-x 2 graydon graydon 523 Jan 21 23:14 test_contract_data.wasm
-rwxrwxr-x 2 graydon graydon 190 Jan 21 23:14 test_empty.wasm
-rwxrwxr-x 2 graydon graydon 131 Jan 21 23:14 test_empty2.wasm
-rwxrwxr-x 2 graydon graydon 529 Jan 21 23:14 test_errors.wasm
-rwxrwxr-x 2 graydon graydon 507 Jan 21 23:14 test_events.wasm
-rwxrwxr-x 2 graydon graydon 695 Jan 21 23:14 test_import_contract.wasm
-rwxrwxr-x 2 graydon graydon 1313 Jan 21 23:14 test_invoke_contract.wasm
-rwxrwxr-x 2 graydon graydon 190 Jan 21 23:14 test_logging.wasm
-rwxrwxr-x 2 graydon graydon 3026 Jan 21 23:14 test_udt.wasm
~~~ 

After:
~~~
-rwxrwxr-x 2 graydon graydon 21986 Jan 21 22:57 soroban_token_spec.wasm
-rwxrwxr-x 2 graydon graydon 533 Jan 21 22:57 test_add_i128.wasm
-rwxrwxr-x 2 graydon graydon 534 Jan 21 22:57 test_add_u128.wasm
-rwxrwxr-x 2 graydon graydon 393 Jan 21 22:57 test_add_u64.wasm
-rwxrwxr-x 2 graydon graydon 3002 Jan 21 22:57 test_alloc.wasm
-rwxrwxr-x 2 graydon graydon 541 Jan 21 22:57 test_contract_data.wasm
-rwxrwxr-x 2 graydon graydon 190 Jan 21 22:57 test_empty.wasm
-rwxrwxr-x 2 graydon graydon 131 Jan 21 22:57 test_empty2.wasm
-rwxrwxr-x 2 graydon graydon 493 Jan 21 22:57 test_errors.wasm
-rwxrwxr-x 2 graydon graydon 507 Jan 21 22:57 test_events.wasm
-rwxrwxr-x 2 graydon graydon 665 Jan 21 22:57 test_import_contract.wasm
-rwxrwxr-x 2 graydon graydon 1182 Jan 21 22:57 test_invoke_contract.wasm
-rwxrwxr-x 2 graydon graydon 190 Jan 21 22:57 test_logging.wasm
-rwxrwxr-x 2 graydon graydon 3266 Jan 21 22:57 test_udt.wasm
~~~

### Size-optimized

Before:
~~~
-rw-rw-r-- 1 graydon graydon 10320 Jan 21 23:12 soroban_token_spec.wasm
-rw-rw-r-- 1 graydon graydon 430 Jan 21 23:12 test_add_i128.wasm
-rw-rw-r-- 1 graydon graydon 429 Jan 21 23:12 test_add_u128.wasm
-rw-rw-r-- 1 graydon graydon 360 Jan 21 23:12 test_add_u64.wasm
-rw-rw-r-- 1 graydon graydon 1524 Jan 21 23:12 test_alloc.wasm
-rw-rw-r-- 1 graydon graydon 424 Jan 21 23:12 test_contract_data.wasm
-rw-rw-r-- 1 graydon graydon 183 Jan 21 23:12 test_empty.wasm
-rw-rw-r-- 1 graydon graydon 124 Jan 21 23:12 test_empty2.wasm
-rw-rw-r-- 1 graydon graydon 360 Jan 21 23:12 test_errors.wasm
-rw-rw-r-- 1 graydon graydon 458 Jan 21 23:12 test_events.wasm
-rw-rw-r-- 1 graydon graydon 571 Jan 21 23:12 test_import_contract.wasm
-rw-rw-r-- 1 graydon graydon 719 Jan 21 23:12 test_invoke_contract.wasm
-rw-rw-r-- 1 graydon graydon 183 Jan 21 23:12 test_logging.wasm
-rw-rw-r-- 1 graydon graydon 2561 Jan 21 23:12 test_udt.wasm
~~~

After:
~~~
-rw-rw-r-- 1 graydon graydon 9581 Jan 21 23:09 soroban_token_spec.wasm
-rw-rw-r-- 1 graydon graydon 407 Jan 21 23:09 test_add_i128.wasm
-rw-rw-r-- 1 graydon graydon 406 Jan 21 23:09 test_add_u128.wasm
-rw-rw-r-- 1 graydon graydon 282 Jan 21 23:09 test_add_u64.wasm
-rw-rw-r-- 1 graydon graydon 1488 Jan 21 23:09 test_alloc.wasm
-rw-rw-r-- 1 graydon graydon 451 Jan 21 23:09 test_contract_data.wasm
-rw-rw-r-- 1 graydon graydon 183 Jan 21 23:09 test_empty.wasm
-rw-rw-r-- 1 graydon graydon 124 Jan 21 23:09 test_empty2.wasm
-rw-rw-r-- 1 graydon graydon 383 Jan 21 23:09 test_errors.wasm
-rw-rw-r-- 1 graydon graydon 458 Jan 21 23:09 test_events.wasm
-rw-rw-r-- 1 graydon graydon 526 Jan 21 23:09 test_import_contract.wasm
-rw-rw-r-- 1 graydon graydon 767 Jan 21 23:09 test_invoke_contract.wasm
-rw-rw-r-- 1 graydon graydon 183 Jan 21 23:09 test_logging.wasm
-rw-rw-r-- 1 graydon graydon 2578 Jan 21 23:09 test_udt.wasm
~~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -841,7 +841,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-common"
 version = "0.0.12"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=d803b0922d1d9eb4d3cfde5e6d574c7776c23388#d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=0f0fbdbc0fbed2d5bd8ca531986dd393511b0543#0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 dependencies = [
  "crate-git-revision",
  "serde",
@@ -854,7 +854,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-guest"
 version = "0.0.12"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=d803b0922d1d9eb4d3cfde5e6d574c7776c23388#d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=0f0fbdbc0fbed2d5bd8ca531986dd393511b0543#0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 dependencies = [
  "soroban-env-common",
  "static_assertions",
@@ -863,7 +863,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-host"
 version = "0.0.12"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=d803b0922d1d9eb4d3cfde5e6d574c7776c23388#d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=0f0fbdbc0fbed2d5bd8ca531986dd393511b0543#0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 dependencies = [
  "backtrace",
  "curve25519-dalek",
@@ -885,7 +885,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-macros"
 version = "0.0.12"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=d803b0922d1d9eb4d3cfde5e6d574c7776c23388#d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=0f0fbdbc0fbed2d5bd8ca531986dd393511b0543#0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 dependencies = [
  "itertools",
  "proc-macro2",
@@ -911,7 +911,7 @@ dependencies = [
 [[package]]
 name = "soroban-native-sdk-macros"
 version = "0.0.12"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=d803b0922d1d9eb4d3cfde5e6d574c7776c23388#d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=0f0fbdbc0fbed2d5bd8ca531986dd393511b0543#0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 dependencies = [
  "itertools",
  "proc-macro2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,17 +37,17 @@ soroban-token-spec = { version = "0.4.3", path = "soroban-token-spec" }
 [workspace.dependencies.soroban-env-common]
 version = "0.0.12"
 git = "https://github.com/stellar/rs-soroban-env"
-rev = "d803b0922d1d9eb4d3cfde5e6d574c7776c23388"
+rev = "0f0fbdbc0fbed2d5bd8ca531986dd393511b0543"
 
 [workspace.dependencies.soroban-env-guest]
 version = "0.0.12"
 git = "https://github.com/stellar/rs-soroban-env"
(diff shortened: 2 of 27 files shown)


graydon:
Closing in favor of #837

PR #835, opened 2023-01-20
Fix contractimport for custom discriminants (#834)

### LINKED ISSUE #834: contractimport fails when importing custom types with discriminants

`0.4.3` but same with older versions (oldest tested `0.3.2`)

### What did you do?

Create a contract using a custom type enum with custom discriminants.
i.e.
```rust
#[contracttype]
#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
#[repr(u32)]
pub enum MyCoolType {
    ThisTypeIsSadBecauseTheNameIsTooLong = 0,
    Type = 1
}
```

Import the generated WASM file using the ``soroban_sdk::contractimport`` macro.

Everything should compile just fine.

### What did you see instead?
```bash
$ RUSTFLAGS="-Z macro-backtrace" cargo +nightly test
error[E0507]: cannot move out of `*self` which is behind a shared reference
   --> buggy/src/lib.rs:4:5
    |
4   | /     soroban_sdk::contractimport!(
5   | |         file = "../target/wasm32-unknown-unknown/release/soroban_error_types.wasm"
6   | |     );
    | |     ^
    | |     |
    | |_____move occurs because `*self` has type `MyCoolType`, which does not implement the `Copy` trait
    |       in this procedural macro expansion
    |
   ::: /home/kai/.cargo/registry/src/github.com-1ecc6299db9ec823/soroban-sdk-macros-0.3.2/src/lib.rs:123:1
    |
123 |   pub fn contracttype(metadata: TokenStream, input: TokenStream) -> TokenStream {
    |   ----------------------------------------------------------------------------- in this expansion of `#[soroban_sdk::contracttype]`
```

### Addition Information
An example implementation for this bug is given here https://github.com/Smephite/soroban_tests.
It seems to likely be related to #758 .
The bug seems to be fixed when adding the `Copy` trait to the derived values on the contract types.

Smephite:
### What

Add the `Copy` trait to derived custom types when using `contractimport`

### Why

Close #834.

### Known limitations

This should not expose any issues as Integer Variant Enums as described in the [docs](https://soroban.stellar.org/docs/learn/custom-types#enum-integer-variants) should derive `Clone` anyhow.


soroban-spec/src/gen/rust/types.rs
```diff
@@ -31,7 +31,7 @@ pub fn generate_struct(spec: &ScSpecUdtStructV0) -> TokenStream {
         });
         quote! {
             #[soroban_sdk::contracttype(export = false)]
-            #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+            #[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
```
leighmcculloch, line 34: Unfortunately we can't add `Copy` to struct or enum types that can contain other types. `Copy` requires that all contained types are also `Copy`, and not all types a struct or enum can contain are `Copy`.

For example, `Vec` cannot be made `Copy`, and this means that any custom type `struct` that contains a `Vec` would fail to compile.

leighmcculloch (commented):
It looks like this PR adds `Copy` to all types of contract types, however I think we need to limit it to `contracttype` on integer enums, and `contracterror` types.

tyvdh:
What's the recommended work around here? It just seems very odd/unintuitive that you could explicitly declare:
```rust
#[contracttype]
#[derive(Copy, Clone, Eq, PartialEq)]
pub enum MapElement {
    Asteroid,
    FuelPod,
}
```
and yet when getting the bindings from the wasm it's:
```rust
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum MapElement {
    Asteroid,
    FuelPod,
}
```

leighmcculloch:
Only enums with integer values need the Copy derive.

See the examples here: https://soroban.stellar.org/docs/how-to-guides/custom-types#custom-type-enum

Specifically:
```rust
#[contracttype]
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum Enum {
    A,
    B(...),
}
```

```rust
#[contracttype]
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
#[repr(u32)]
pub enum Enum {
    A = 1,
    B = 2,
}
```

Smephite:
I think @tyvdh's question might've been misleading here, as it is only loosely connected to the issue at hand.
Regarding custom discriminants it is true that only enum with integer values need the `Copy` trait, but the note by Tyler was that structs deriving `Copy` in their original source won't derive `Copy` after the wasm compile -> regenerating bindings using soroban flow.  
To resolve this second issue, it is probable that a more sophisticated analysis of the provided struct is required to circumvent the already mentioned issue of the possibility of the struct containing un-copyable types.

As this seems to be only loosely connected to this issue / PR it might be suitable to open a new issue.

Smephite:
> I think @tyvdh's question might've been misleading here, as it is only loosely connected to the issue at hand. Regarding custom discriminants it is true that only enum with integer values need the `Copy` trait, but the note by Tyler was that structs deriving `Copy` in their original source won't derive `Copy` after the wasm compile -> regenerating bindings using soroban flow. To resolve this second issue, it is probable that a more sophisticated analysis of the provided struct is required to circumvent the already mentioned issue of the possibility of the struct containing un-copyable types.
> 
> As this seems to be only loosely connected to this issue / PR it might be suitable to open a new issue.

Another possibility could be to throw a compiler warning when trying to derive `Copy` on a `contract_type` that does not support it.

leighmcculloch:
I think the ideal fix for this issue would be for the `contracttype` proc-macro to 

> the note by Tyler was that structs deriving `Copy` in their original source won't derive `Copy` after the wasm compile -> regenerating bindings using soroban flow.
> To resolve this second issue, it is probable that a more sophisticated analysis of the provided struct is required to circumvent the already mentioned issue of the possibility of the struct containing un-copyable types.

Ah yes, indeed. The contract specification doesn't capture third party trait implementations. We could extend the contract specification to try and capture other things like that. Something worth considering, although not without tradeoffs.

Smephite:
I updated the PR to only add the Copy derive in the `generate_enum` function.

For my own understanding, I have one question regarding the Spec.
```
#[contracttype]
#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
#[repr(u32)]
pub enum MyCoolType {
    ThisTypeIsSadBecauseTheNameIsTooLong = 0,
    Type = 1
}

#[contracttype]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum MyNormalType {
    One(Vec<Symbol>),
    Two
}
```

in a rust sense both of these are `enums`, but when inspecting the wasm using the soroban cli the second enum is of `"type": "union"`.  
What kind of rust enums will be mapped to the soroban enum?  
Is it safe to assume that every enum variant reaching `generate_enum(...)` will not contain an un-copyable value?

leighmcculloch (approved):

PR #836, opened 2023-01-21
Improve compiler errors for UDTs

brson:
### What

- Display the variant name when more than one fields are present.
- Print an error when the variant has named fields.

### Why

Improve usability.

I am also looking at expanding contracttype to support multiple fields and named fields, but won't get that done in the next few days.

### Known limitations

na

### DIFF
--- a/soroban-sdk-macros/src/derive_enum.rs
+++ b/soroban-sdk-macros/src/derive_enum.rs
@@ -38,7 +38,7 @@ pub fn derive_type_enum(
                 errors.push(Error::new(ident.span(), format!("enum variant name {}", e)));
             }
             if v.fields.len() > 1 {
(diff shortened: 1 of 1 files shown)


brson:
I pushed one more commit that errors on enums with no variants. This case also currently emits syntax errors.

soroban-sdk-macros/src/derive_enum.rs
```diff
@@ -21,6 +21,12 @@ pub fn derive_type_enum(
     let mut errors = Vec::<Error>::new();
 
     let variants = &data.variants;
+    if variants.is_empty() {
+        errors.push(Error::new(
+            enum_ident.span(),
+            format!("enum {} must have variants", enum_ident),
+        ));
+    }
```
leighmcculloch, line 29: Thanks for catching this.

leighmcculloch (approved):
👏🏻 Thanks for fixing these cases where errors were lacking.

PR #839, opened 2023-01-26
Allow same named types and functions

### LINKED ISSUE #828: Cannot have a contract with a type and method of the same name
For example, if the getter method in https://github.com/stellar/soroban-examples/blob/main/custom_types/src/lib.rs were named `state`, you'd get an error. This seems like a common pattern people would hit (especially with the 10char limit for method names).

leighmcculloch:
### What
Change the names of the static variables that contain the spec XDR to include a type or function qualifier for the kind of entry the spec contains.

### Why
To allow same named types and functions in contracts. When a type and function have the same name today they result in  two static variables that have the same names. Adding a qualifier to what they are to those variables names will stop them from colliding.

Close #828


sisuresh (approved):
Nice!

PR #849, opened 2023-02-02
Remove build-optimized makefile target

leighmcculloch:
### What
Remove build-optimized makefile target.

### Why
We aren't using the method of optimization that it employs in everyday development, and it complicates the Makefile.


graydon (approved):
Yeah this is probably for the best at this point; we can fish it out of git history if we ever want to do it again but I think it's not a big win anymore.

PR #850, opened 2023-02-02
Support variants with multiple fields in UDTs

brson:
### What

Add support for UDT enums with variants containing up to 12 fields.

Add more test cases.

### Why

UDT enum variants are currently limited to a single field, which seems arbitrary and surprising.

### Known limitations

This does not add support for struct-like variants.

The spec for tuple-like variants is now a ScpSpecUdtUnionCaseV0 of ScSpecTypeTuple:

```rust
        ScSpecUdtUnionCaseV0 {
            name: name.try_into().unwrap_or_else(|_| StringM::default()),
            type_: Some(ScSpecTypeDef::Tuple(Box::new(ScSpecTypeTuple {
                value_types: field_types,
            }))),
        }
```

Please consider whether this is correct. I am not sure.

This patch requires updates to rs-stellar-xdr and rs-soroban-sdk to support conversions of 13-tuples:

- https://github.com/stellar/rs-soroban-env/pull/658
- https://github.com/stellar/rs-stellar-xdr/pull/230

This patch went through a lot of iteration and ended up making a bigger diff than necessary.

cc https://github.com/stellar/rs-soroban-sdk/issues/153

### DIFF
--- a/soroban-sdk-macros/src/derive_enum.rs
+++ b/soroban-sdk-macros/src/derive_enum.rs
@@ -5,8 +5,8 @@ use soroban_env_common::Symbol;
 use syn::{spanned::Spanned, Attribute, DataEnum, Error, Fields, Ident, Path};
 
 use stellar_xdr::{
-    ScSpecEntry, ScSpecTypeDef, ScSpecUdtUnionCaseTupleV0, ScSpecUdtUnionCaseV0,
-    ScSpecUdtUnionCaseVoidV0, ScSpecUdtUnionV0, StringM, WriteXdr,
+    Error as XdrError, ScSpecEntry, ScSpecTypeDef, ScSpecUdtUnionCaseTupleV0, ScSpecUdtUnionCaseV0,
+    ScSpecUdtUnionCaseVoidV0, ScSpecUdtUnionV0, StringM, VecM, WriteXdr,
 };
 
 use crate::{doc::docs_from_attrs, map_type::map_type};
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> The spec for tuple-like variants is now a ScpSpecUdtUnionCaseV0 of ScSpecTypeTuple:
> 
> ```rust
>         ScSpecUdtUnionCaseV0 {
>             name: name.try_into().unwrap_or_else(|_| StringM::default()),
>             type_: Some(ScSpecTypeDef::Tuple(Box::new(ScSpecTypeTuple {
>                 value_types: field_types,
>             }))),
>         }
> ```

I think this will result in code generated that doens't match the original code. For example:

```rust
type Enum enum {
    Variant(A1, A2)
}
```

Will result inthe following being code generated:

```rust
type Enum enum {
    Variant((A1, A2))
}
```

So I think we need to change the XDR so that the `type_` value can be a vec instead of a optional, so that it can contain one or more types.

leighmcculloch:
The smallest diff we could make to the spec to support it is:
```diff
diff --git a/Stellar-contract-spec.x b/Stellar-contract-spec.x
index 56c3adf..d26c169 100644
--- a/Stellar-contract-spec.x
+++ b/Stellar-contract-spec.x
@@ -139,7 +139,7 @@ struct SCSpecUDTUnionCaseV0
 {
     string doc<SC_SPEC_DOC_LIMIT>;
     string name<60>;
-    SCSpecTypeDef *type;
+    SCSpecTypeDef type<12>;
 };
 
 struct SCSpecUDTUnionV0
```

But maybe we should make that struct a union, and explicitly distinguish between unit and tuple variants.

leighmcculloch:
I'm making some changes to the spec xdr to support multiple tuples in the spec itself: https://github.com/stellar/stellar-xdr/pull/67.

brson:
Thanks for making the XDR changes for me.

I started upgrading rs-soroban-sdk to the latest XDR until I discovered the new `doc` fields, and the open questions about it in chat. Do you want me to hold off and let you deal with the SDK upgrade?

leighmcculloch:
Sorry about that, the docs changes are in #841 which I plan to finish and need on Monday. You could base your changes on that PR though if you wanted to move ahead.

brson:
Rebased.

soroban-sdk-macros/src/derive_enum.rs
```diff
@@ -67,75 +58,34 @@ pub fn derive_type_enum(
...
-                            Ok(t) => t,
-                            Err(e) => {
-                                errors.push(e);
-                                ScSpecTypeDef::I32
-                            }
-                        },
-                        ].try_into().unwrap()
-                    }
+            let has_fields = v.fields.iter().next().is_some();
+            if has_fields {
```
leighmcculloch, line 62: Something I notice is that because we're continuing to use the number of fields to determine if the type is a unit vs tuple variant, we're not distinguishing between an empty tuple and a unit variant.

```rust
enum E {
    A,
    B(),
}
```

Because there's no distinction between the two, code generated versions of the contract will diverge and become:
```rust
enum E {
    A,
    B,
}
```

We could address the divergence by determining if we should map to the void or tuple union case on whether `v.fields` is a `Fields::Unnamed` or a `Fields::Unit` instead of on the number of fields. (https://docs.rs/syn/0.12.15/syn/enum.Fields.html)

But this is also such an edge case, that arguably we should just treat them both the same. There's really no benefit in encoding `B()` differently, and I doubt any other programming language will render the difference in a meaningful way. To many Rust developers the fact that `A` and `B()` are different is also probably a surprise. Also, if we encode `B()` as a tuple variant it is less efficient in the spec and in the conversions, so there's no win there.

So I think this is good as is, keeping to using `v.fields` length.

Thoughts anyone else?

leighmcculloch (approved):

soroban-sdk-macros/src/derive_enum.rs
brson: I forgot that empty tuple variants were possible - I would never think to write them!

That they won't round trip when re-rendered as rust seems like it could become a surprising problem in ways I can't imagine yet.

We could hedge and make empty-tuple variants an error for now, figure it out later. Hopefully nobody is using them already.

leighmcculloch: It shocked me when empty tuple variants weren't the same as unit variants. I think there's a variety of ways they are a surprise and a foot gun. So doing as you suggest and erroring for now feels like a good plan, possibly even long term.

leighmcculloch: Opened an issue for adding an error: https://github.com/stellar/rs-soroban-sdk/issues/861.

PR #856, opened 2023-02-07
Re-export all xdr

paulbellamy:
### What

Re-export all the xdr from this library.

### Why

So we don't have to add types every time xdr changes. We already forgot ScAddress this time.

### Known limitations

I'm guessing at the rust syntax. Is this the best way?


leighmcculloch (approved):

leighmcculloch:
> Is this the best way?

This is the way.

soroban-sdk/src/xdr.rs
```diff
@@ -7,38 +7,6 @@ pub use super::env::xdr::HostFunction;
 // XDR generic types and traits.
 #[cfg(not(target_family = "wasm"))]
 pub use super::env::xdr::ReadXdrIter;
```
leighmcculloch, line 9: You can delete the couple lines about too.

leighmcculloch: You can delete both manual exports.

leighmcculloch (approved):

PR #862, opened 2023-02-11
Add docs in specs for types

leighmcculloch:
### What
Add docs in specs for types.

### Why
To further expand and fill out what we store docs for in specs.


leighmcculloch:
Todo:
- [ ] Tests.

soroban-sdk/src/tests/contract_docs_enum.rs
```diff
@@ -0,0 +1,46 @@
+use crate as soroban_sdk;
```
dmkozh, line 1: I'm not strictly against creating multiple test files, but just wonder if there is something preventing us from consolidating the tests into a single file.

soroban-sdk/src/tests/contract_docs_enum_int.rs
```diff
@@ -0,0 +1,38 @@
+use crate as soroban_sdk;
+use soroban_sdk::contracttype;
+use stellar_xdr::{ReadXdr, ScSpecEntry, ScSpecUdtEnumCaseV0, ScSpecUdtEnumV0};
+
+/// E has variants A and B.
+#[contracttype]
+#[derive(Copy, Clone)]
+pub enum E {
+    /// A is a.
```
dmkozh, line 9: Can we cover a multi-line comment for enum variant somewhere?

dmkozh (approved):

soroban-sdk/src/tests/contract_docs_enum.rs
leighmcculloch: Nothing prevents it. Anything we can do in a file we can do inside a mod inside one file. I'll try that before merging to see if it's too overwhelming.

leighmcculloch: Consolidated the tests in one big file: 3d6c7f63

soroban-sdk/src/tests/contract_docs_enum_int.rs
leighmcculloch: Added some multiline cases in fce2c958.

PR #869, opened 2023-02-17
Expose reset budget functions to testutils

Smephite:
### What

This PR exposes the various reset budget functions to the testutils.

### Why

The ability to change the budget to arbitrary values was [just recently](https://github.com/stellar/rs-soroban-env/pull/677) added to the soroban-env.  
This PR also aims to allow this during the testing stage of a contract, as suggested here:
https://discord.com/channels/897514728459468821/1076187600668340384/1076187600668340384

### Known limitations

Previously the `reset` method in the test utils set the budget limits to unlimited.  
As this PR introduces the `reset_unlimited` functionality, the `reset` function was changed to only reset to the default budget limits.  
This change might be breaking in rare circumstances.

### DIFF
--- a/soroban-sdk/src/testutils.rs
+++ b/soroban-sdk/src/testutils.rs
@@ -87,7 +87,26 @@ pub mod budget {
         /// Reset the budget.
         pub fn reset(&mut self) {
             self.0.reset_default();
+        }
+
+        pub fn reset_unlimited(&mut self) {
             self.0.reset_unlimited();
+        }
+
+        pub fn reset_limits(&mut self, cpu: u64, mem: u64) {
+            self.0.reset_limits(cpu, mem);
+        }
+
+        pub fn reset_unlimited_cpu(&mut self) {
+            self.0.reset_unlimited_cpu();
+        }
+
+        pub fn reset_unlimited_mem(&mut self) {
+            self.0.reset_unlimited_mem();
+        }
+
+        pub fn reset_inputs(&mut self) {
+            self.0.reset_inputs();
         }
 
         /// Returns the CPU instruction cost.
--- a/soroban-sdk/src/testutils.rs
+++ b/soroban-sdk/src/testutils.rs
@@ -85,7 +85,7 @@ pub mod budget {
         }
 
         /// Reset the budget.
-        pub fn reset(&mut self) {
+        pub fn reset_default(&mut self) {
             self.0.reset_default();
         }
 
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/testutils.rs
```diff
@@ -87,9 +87,28 @@ pub mod budget {
         /// Reset the budget.
         pub fn reset(&mut self) {
             self.0.reset_default();
+        }
```
jayz22, line 90: Maybe change the `reset` to `reset_default` to be consistent.

jayz22 (commented):
Overall looks good. My only comment is most of these methods are for convenience only. The only one you'll probably need is just `reset_limits` and these other ones are just for convenience (e.g writing one less line). I suggest to only keep what's essential to minimize the impact once some of those methods are cleaned away eventually.

soroban-sdk/src/testutils.rs
```diff
@@ -87,9 +87,28 @@ pub mod budget {
...
+            self.0.reset_unlimited_cpu();
+        }
+
+        pub fn reset_unlimited_mem(&mut self) {
+            self.0.reset_unlimited_mem();
+        }
+
+        pub fn reset_inputs(&mut self) {
+            self.0.reset_inputs();
+        }
```
jayz22, line 110: I don't know if you'll need the `reset_input` function in any case. Once you call the reset budget it automatically resets inputs.

Smephite: I'd argue that it can be useful in a scenario where you have to call multiple functions in one or more contracts for setup but only want to measure one.  
e.g., in FCA00C, we have the game engine `init` and a user's `solve` but only want to measure the performance for the user's contract. Therefore we use the tests as usual and reset the inputs between the two calls.

Smephite:
> Overall looks good. My main comment is most of these methods are for convenience only. The only one you'll probably need is just `reset_limits` and these other ones are just for convenience (e.g writing one less line). I suggest only keep what's essential to minimize the impact once some of those methods are cleaned away eventually.

good comment; I was only mirroring what the env already exposes. Which methods should be kept?  
 I'd propose `reset_unlimited`, `reset_default` and `reset_limits`.

Smephite:
> Overall looks good. My main comment is most of these methods are for convenience only. The only one you'll probably need is just `reset_limits` and these other ones are just for convenience (e.g writing one less line). I suggest to only keep what's essential to minimize the impact once some of those methods are cleaned away eventually.

Good comment; I was only mirroring the exposed methods of the env.  
What methods would you sugest to keep?  
I would propose `reset_default`, `reset_limits`, `reset_unlimited` (and `reset_inputs`)

soroban-sdk/src/testutils.rs
Smephite: One could make the point for rust calling `reset_limits` when ever one would call `reset_inputs` but I it feels more clean with an own function imo

jayz22: Sounds good.

jayz22 (approved):
Looks good, thanks for updating!

jayz22:
@leighmcculloch to merge.

graydon (approved):

PR #870, opened 2023-02-21
feat(Ledger Snapshot): add contract methods

willemneal:
### What

Adding methods used in the CLI upstream

### Why

Better to have these as instance methods.

### Known limitations

[TODO or N/A]


soroban-ledger-snapshot/src/lib.rs
```diff
@@ -118,6 +124,61 @@ impl LedgerSnapshot {
             }
         }
     }
+
+    pub fn add_contract_to_entries(&mut self, contract_id: [u8; 32], wasm_hash: [u8; 32]) {
```
leighmcculloch, line 128: Could we add comments to both new functions?

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -118,6 +124,61 @@ impl LedgerSnapshot {
...
+            data: LedgerEntryData::ContractCode(ContractCodeEntry {
+                code: contract.try_into()?,
+                ext: ExtensionPoint::V0,
+                hash: hash.clone(),
+            }),
+            ext: LedgerEntryExt::V0,
+        };
+        for (k, e) in entries.iter_mut() {
+            if **k == code_key {
+                **e = code_entry;
```
leighmcculloch, line 175: Could you add happy path tests for both new functions, given they have some interesting logic around sha2-ing data?

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -118,6 +124,61 @@ impl LedgerSnapshot {
...
+            }),
+            ext: LedgerEntryExt::V0,
+        };
+        for (k, e) in entries.iter_mut() {
+            if **k == contract_key {
+                **e = contract_entry;
+                return;
+            }
+        }
+        entries.push((Box::new(contract_key), Box::new(contract_entry)));
```
leighmcculloch, line 153: Would it be reasonable to use the `update_entries` function above to do the write? It contains logic already for identifying an existing key and replacing it.

leighmcculloch (commented):
Three small asks, otherwise this looks good to me.

leighmcculloch:
@willemneal This is pretty stale. Do you think it's salvageable? I'm not clear on if it's still needed or relevant.

PR #871, opened 2023-02-21
feat: initial work to generate TS from XDR

willemneal:
### What

Generate a TS file from XDR to allow for building TS interfaces from contracts.

### Why

Allow Dapps to create custom interfaces for the contracts they interact with.  This will be integrated with the CLI to be part of workflow.

### Known limitations

[TODO or N/A]


soroban-spec/src/gen/ts/boilerplate.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(non_snake_case)]
+use heck::ToLowerCamelCase;
 use heck::ToShoutySnakeCase;
```
willemneal, line 3:
```suggestion
use heck::{ToShoutySnakeCase, ToLowerCamelCase};
```

soroban-spec/src/gen/ts/boilerplate.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(non_snake_case)]
+use heck::ToLowerCamelCase;
```
willemneal, line 2:
```suggestion
```

soroban-spec/root/README.md
```diff
@@ -0,0 +1,55 @@
+# abundance-token JS
```
willemneal, line 1: Are these files supposed to be included?

chadoh: I don't know how they got here; no they are not

soroban-spec/src/gen/ts/project_template/src/invoke.ts
```diff
@@ -67,8 +69,25 @@ export async function invoke({ method, args = [], fee = 100 }: InvokeArgs): Prom
     .build()
 
   const simulated = await Server.simulateTransaction(tx)
+  
+  const auths = simulated.results?.[0]?.auth
+
+  // is it possible for `auths` to be present but empty? Probably not, but let's be safe.
+  if (auths && auths.length > 0) {
```
willemneal, line 76:
```suggestion
  if (auths?.length > 0) {
```

You can do this, no?

chadoh: yes, thanks, pushed

soroban-spec/src/gen/ts/project_template/src/invoke.ts
```diff
@@ -67,8 +69,25 @@ export async function invoke({ method, args = [], fee = 100 }: InvokeArgs): Prom
...
+      throw new NotImplementedError("Multiple auths not yet supported")
+    }
+
+    const auth = SorobanClient.xdr.ContractAuth.fromXDR(auths[0], 'base64')
+
+    if (auth.addressWithNonce() !== undefined) {
+      throw new NotImplementedError(
+        `This transaction needs to be signed by ${
+          auth.addressWithNonce()
+        }; how do we do that?`
```
leighmcculloch, line 87: 👍🏻

leighmcculloch:
I've changed the target/base branch of this PR to `release/v0.8.5`, since we hope to release this forking from the last release, v0.8.4. This PR needs rebasing onto `release/v0.8.5`, and any changes not intended to be part of the PR that are in its history need discarding.

leighmcculloch:
Note that there was a substantial rename of Status to Error on main, that will not be included in v0.8.5. Unfortunately this change will need to ship without that, which will mean when this change is applied to `main`, it'll need rewriting. We'll essentially need to write a significant number of lines of this PR twice.

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,339 @@
...
+    match entry {
+        Entry::Function {
+            doc,
+            name,
+            inputs,
+            outputs,
+        } => {
+            let sign_me = doc
+                .contains("@signme")
+                .then_some("sign: true, ")
```
leighmcculloch, line 89: Is the plan to keep this? @chadoh If I understood some things you were saying today, it is no longer needed since preflight is used to determine when signing is required.

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,339 @@
...
+                .unwrap_or_default();
+            let return_type = if outputs.is_empty() {
+                "".to_owned()
+            } else if outputs.len() == 1 {
+                format!(": Promise<{}>", type_to_ts(&outputs[0]))
+            } else {
+                format!(
+                    ": Promise<[{}]>>",
+                    outputs.iter().map(type_to_ts).join(", ")
+                )
```
leighmcculloch, line 112: It's worth noting that `outputs` in `ScSpecFunctionV0` ([ref](https://github.com/stellar/rs-stellar-xdr/blob/a2f370c930bb94f7bc0c9ea0426ecef3700b90a9/src/next/generated.rs#L5751-L5756)) is a `VecM<ScSpecTypeDef, 1>`. The `1` means the maximum length is guaranteed to be `1`. So this else block can never be triggered. 

I guess this arises here because the way we're generating TS is by following a path of `ScSpecEntry` -> `json::Entry` -> `TS`, and in the conversion to `json::Entry` we lose information.

If the `ScSpecFunctionV0` was available here, you could do the following code to guarantee the maximum length of 1 and to get a compiler error if that ever changed:
```rust
let outputs: &VecM<ScSpecTypeDef, 1> = &sc_spec_function_v0.outputs;
```

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,339 @@
...
+
+pub fn generate_from_wasm(wasm: &[u8]) -> Result<String, FromWasmError> {
+    let spec = from_wasm(wasm)?;
+    let json = generate(&spec);
+    Ok(json)
+}
+
+pub fn generate(spec: &[ScSpecEntry]) -> String {
+    let collected: Vec<_> = spec.iter().map(Entry::from).collect();
+    collected.iter().map(entry_to_ts).join("\n")
```
leighmcculloch, line 62: I'm surprised the way we generate TS is by converting from `ScSpecEntry` to `json::Entry` to TS. The JSON format outputs a contract interface as JSON, not a contract client, or even usable types by a contract. It serves a different purpose. I wonder if this will come back to bite us, as we may want to change the format of the JSON interface outputted without changing the output of the TS client. `ScSpecEntry` canonically describes the contract interface and contains the most information. Given it's available I expected we'd convert directly from it.

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,339 @@
...
+            if name == "Error" {
+                let cases = cases
+                    .iter()
+                    .map(|c| format!("{{error:\"{}\"}}", c.doc))
+                    .join(",\n  ");
+                return format!(
+                    r#"const Errors = [ 
+  {cases}
+]"#
+                );
```
leighmcculloch, line 197: What's the `Error` special case for? Contract errors should be separated into a different type `ScSpecErrorEnum`, and shouldn't show up as `ScSpecEnum`.

soroban-spec/src/gen/ts/boilerplate.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    #[test]
+    fn test_project_dir_location() {
+        let temp_dir = TempDir::new().unwrap();
+        let _: Project = init(&temp_dir).unwrap();
+    }
+
+    #[test]
+    fn test_project_dir_location_non_temp() {
```
leighmcculloch, line 137: Writing into a non-tmp location in a test is a little scary. Do we need this? It shouldn't matter where the writing happens, writing is writing.

soroban-spec/src/gen/ts/boilerplate.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    use super::*;
+
+    const EXAMPLE_WASM: &[u8] =
+        include_bytes!("../../../../target/wasm32-unknown-unknown/release/test_contract_data.wasm");
+
+    fn init(root: impl AsRef<Path>) -> std::io::Result<Project> {
+        let spec = crate::read::from_wasm(EXAMPLE_WASM).unwrap();
+        let p: Project = root.as_ref().to_path_buf().try_into()?;
+        p.init(
```
leighmcculloch, line 119: Could we have an assertion on the full contents of the directory that gets generated? Presumedly by including a test fixture which is the generated directory and comparing the two? It will be invaluable for being able to visualize changes in the generated code when PRs are opened.

leighmcculloch (commented):
Few questions, some asks. Some concern about building the conversion on-top of the json converter, which motivation for is unclear.

Ultimately defer to @paulbellamy and co as I'm not well versed in TypeScript.

soroban-spec/src/gen/ts/boilerplate.rs
leighmcculloch: (I should go back and add something similar to the json generator too.)

soroban-spec/src/gen/ts.rs
willemneal: JSON seemed close enough for converting to TS types. What other info is lost?

Personally I think we should write a XDR -> JSON schema parser. This way the generated schema is self describing and provide an additional validation method and make use of tools that generate a form from json schema.

soroban-spec/src/gen/ts.rs
willemneal: Removing this!

soroban-spec/src/gen/ts.rs
willemneal: Initially to get something done I copied from the JSON stuff and liked the smaller surface area of the JSON type. However, as you point out it's a bad because it loses information. This to me suggests taking the time to generate to actual json schema so that these type constraints such as this one would be maintained. NEAR and CosmWasm use JSON schema, however with no agreed upon contract schema they all differ in their shape.

Issues like this are another reason that I still think producing a Wasm which can handle the JSON <-> XDR this also removes all dependencies from the generated code.

soroban-spec/src/gen/ts/boilerplate.rs
willemneal: Fair enough!

soroban-spec/src/gen/ts/boilerplate.rs
willemneal: Done!

willemneal:
@leighmcculloch I added a diff test and a new make command to build the snapshot directory. Not sure how to get around the deny issues.

soroban-spec/src/gen/ts.rs
leighmcculloch: The canonical form of contract interfaces is XDR, for better or worse, so I think if we have the interface in that form, which we do here, we should generate from the canonical form.

Independent of this we do need JSON <-> XDR, and we have it in the Rust XDR lib. I think there's value in seeing if we can exactly replicate the Rust XDR libs JSON <-> XDR in the js-stellar-base, but that's a big project and don't think we should make any of this work contingent on it.

leighmcculloch: > we should generate from the canonical form

We can merge the PR as it is done now, but I think we should treat this as tech debt and circle back on it so there isn't that dependency / conversion path.

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,351 @@
...
+            throw e;
+        }}
+    }}"#
+                );
+            }
+            let args = (!inputs.is_empty())
+                .then(|| format!("args: [{args}], "))
+                .unwrap_or_default();
+            format!(
+                r#"{ts_doc}export async function {name}({input}){return_type} {{
```
paulbellamy, line 159: Might be able to clean up a bit of code above by always returning a Promise, but for no result: `Promise<void>`

soroban-spec/src/gen/ts.rs
```diff
@@ -0,0 +1,351 @@
...
+  {fields}
+}}
+
+function {name}ToXDR({arg_name}?: {name}): xdr.ScVal {{
+    if (!{arg_name}) {{
+        return {void};
+    }}
+    let arr = [
+        {encoded_fields}
+        ];
```
paulbellamy, line 184:
```suggestion
    ];
```

soroban-spec/src/gen/ts/project_template/src/invoke.ts
```diff
@@ -0,0 +1,147 @@
...
+  let exponentialFactor = 1.5
+
+  while ((Date.now() < waitUntil) && getTransactionResponse.status === "NOT_FOUND") {
+    // Wait a beat
+    await new Promise(resolve => setTimeout(resolve, waitTime))
+    /// Exponential backoff
+    waitTime = waitTime * exponentialFactor;
+    // See if the transaction is complete
+    getTransactionResponse = await Server.getTransaction(sendTransactionResponse.hash)
+  }
```
paulbellamy, line 138: This will actually wait about 11 seconds.. Because of the exponent, but whatever. 😂

soroban-spec/fixtures/ts/src/convert.ts
```diff
@@ -0,0 +1,129 @@
...
+    let b = 0;
+    b |= value.high;
+    b <<= 8;
+    b |= value.low;
+    return b;
+}
+
+export function scvalToString(value: xdr.ScVal): string | undefined {
+    return value.bytes().toString();
+}
```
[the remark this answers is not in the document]
paulbellamy: This seems unused? or.. it should be? Because what is it trying to do?

paulbellamy (commented):
Well, I commented on the wrong (fixture) files, not the src.. But same comments on the src files...

Seems like the ts fixtures are getting a diff when testing anyway?

I agree with @leighmcculloch that it would be better to generate from the ScSpecEntries instead of the json, as it's a more accurate data-source with fewer lossy conversions.

But overall, LGTM. Just a few fixes and cleanups.

soroban-spec/src/gen/ts/project_template/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "dependencies": {
+        "@stellar/freighter-api": "1.4.0",
+        "buffer": "6.0.3",
+        "soroban-client": "0.7.2"
+    },
+    "scripts": {
+        "build": "tsc"
+    },
+    "type": "module",
+    "main": "out/index.js",
```
paulbellamy, line 13: Oh, just noticed. I feel like the more usual js name for this dir is `dist` instead of `out`.

soroban-spec/src/gen/ts/project_template/src/invoke.ts
```diff
@@ -0,0 +1,147 @@
...
+export class NotImplementedError extends Error { }
+
+/**
+ * Invoke a method on the INSERT_CONTRACT_NAME_HERE contract.
+ *
+ * Uses Freighter to determine the current user and sign the transaction.
+ *
+ * @param {string} obj.method - The method to invoke.
+ * @param {any[]} obj.args - The arguments to pass to the method.
+ * @param {boolean} obj.sign - Whether to sign the transaction with Freighter.
```
paulbellamy, line 52: Since this method is just doing a `simulateTransaction` it doesn't actually need to sign anything. And this arg seems unused anyway.

It would be nice if the method name and/or docs indicated it was not going to actually send a real live transaction to the network.

soroban-spec/fixtures/ts/src/convert.ts
```diff
@@ -0,0 +1,129 @@
...
+            return Address.fromScVal(val).toString() as T;
+        case xdr.ScValType.scvString():
+            return val.str().toString() as T;
+        case xdr.ScValType.scvSymbol():
+            return val.sym().toString as T;
+        case xdr.ScValType.scvBytes():
+            return val.bytes() as T;
+        case xdr.ScValType.scvVec(): {
+            type Element = ElementType<T>;
+            return val.vec().map(scValToJs<Element>) as T
```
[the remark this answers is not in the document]
willemneal: Good point.  Should we add prettier that to the generated code too?

soroban-spec/src/gen/ts/project_template/src/invoke.ts
paulbellamy: Wait.... it sometimes sends the txn? what?

soroban-spec/fixtures/ts/src/convert.ts
```diff
@@ -0,0 +1,129 @@
...
+            type Key = KeyType<T>;
+            type Value = ValueType<T>;
+            let res = new Map<Key, Value>();
+            val.map().forEach((e) => {
+                let key = scValToJs<Key>(e.key());
+                let value = scValToJs<Value>(e.val());
+                res.set(key as Key, value as Value);
+            });
+
+            return res as T
```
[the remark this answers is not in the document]
willemneal: TS is mad because without the casting it can't be sure that the specific type returned is `T`. e.g.
```ts
val.str().toString() as T
```
Here we either needed to assert that T is a string or cast it.

This is again an issue with doing the XDR -> JSON here. For one this we lack the ScTypeDef types which include the user defined types. So this type isn't even correct here because ScVal treats maps and structs the same.

Now that we are generating the exact package I think it would be worth returning to using rust for this parsing. Then we just need to ship the XDR.

This prevents having to have parity of JSON generators/parsers.

soroban-spec/src/gen/ts/project_template/src/invoke.ts
willemneal: Yeah we need to fix it to check the footprint instead of the auth.

soroban-spec/fixtures/ts/src/convert.ts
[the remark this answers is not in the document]
paulbellamy: eugh. that sucks. but fine for now.

soroban-spec/src/gen/ts/project_template/tsconfig.json
```diff
@@ -0,0 +1,98 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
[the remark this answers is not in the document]
paulbellamy: Oh, huh. TIL. Yeah sounds fine.

soroban-spec/fixtures/ts/src/convert.ts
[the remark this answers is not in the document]
willemneal: Unused!

soroban-spec/src/gen/ts.rs
willemneal: Good idea!

soroban-spec/src/gen/ts/project_template/src/invoke.ts
paulbellamy: Oh, is this the thing where you're trying to auto-infer if it is a read call or a write call? That will never work correctly. I'd strongly argue for a more viem style [read](https://viem.sh/docs/ethers-migration.html#reading-from-contracts)/[write](https://viem.sh/docs/ethers-migration.html#writing-to-contracts) interface where the dev specifically says which they want to do.

soroban-spec/src/gen/typescript/project_template/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface Result<T, E = Error_> {
+    unwrap(): T,
+    unrap_err(): E,
+    isOk(): boolean,
+    isErr(): boolean,
+};
+
+export class Ok<T> implements Result<T> {
+    constructor(readonly value: T) { }
+    unrap_err(): Error_ {
```
paulbellamy, line 34: Camel casing? and typo
```suggestion
    unwrapErr(): Error_ {
```

soroban-spec/src/gen/typescript/project_template/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+if (typeof window !== 'undefined') {
+    //@ts-ignore Buffer exists
+    window.Buffer = window.Buffer || Buffer;
+}
+
+const regex = /ContractError\((\d+)\)/;
+
+function get_error(err: string): Err<Error_> | undefined {
```
paulbellamy, line 75:
```suggestion
function getError(err: string): Err<Error_> | undefined {
```

soroban-spec/src/gen/typescript/project_template/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    isErr(): boolean {
+        return !this.isOk()
+    }
+}
+
+export class Err<T> implements Result<T> {
+    constructor(readonly error: Error_) { }
+    unrap_err(): Error_ {
```
paulbellamy, line 52:
```suggestion
    unwrapErr(): Error_ {
```

soroban-spec/fixtures/ts/src/convert.ts
```diff
@@ -0,0 +1,109 @@
...
+        case undefined: {
+            return BigInt(0);
+        }
+        case xdr.ScValType.scvU64(): {
+            const { high, low } = scval.u64();
+            return bufToBigint(new Uint32Array([high, low]));
+        }
+        case xdr.ScValType.scvI64(): {
+            const { high, low } = scval.i64();
+            return bufToBigint(new Int32Array([high, low]));
```
Shaptic, line 16: this isn't quite right: low is uint32, high is int32

soroban-spec/fixtures/ts/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export async function put({key, val}: {key: string, val: string}): Promise<void> {
+    let invokeArgs: InvokeArgs = {method: 'put', args: [((i) => xdr.ScVal.scvSymbol(i))(key),
+        ((i) => xdr.ScVal.scvSymbol(i))(val)], };
+    
+    // @ts-ignore Type does exist
+    const response = await invoke(invokeArgs);
+    return ;
+    
+
```
Shaptic, line 96: nitpick: weird whitespace, needs a formatter pass

soroban-spec/fixtures/ts/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return undefined;
+}
+
+export async function put({key, val}: {key: string, val: string}): Promise<void> {
+    let invokeArgs: InvokeArgs = {method: 'put', args: [((i) => xdr.ScVal.scvSymbol(i))(key),
+        ((i) => xdr.ScVal.scvSymbol(i))(val)], };
+    
+    // @ts-ignore Type does exist
+    const response = await invoke(invokeArgs);
```
Shaptic, line 93: why not return the `invoke` promise?

soroban-spec/fixtures/ts/src/index.ts
Shaptic: oh these are test fixtures :facepalm:

soroban-spec/src/gen/typescript/project_template/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+export type i256 = bigint;
+export type Address = string;
+export type Option<T> = T | undefined;
+
+/// Error interface containing the error message
+export interface Error_ { message: string };
+
+export interface Result<T, E = Error_> {
+    unwrap(): T,
+    unrap_err(): E,
```
willemneal, line 27:
```suggestion
    unwrapErr(): E,
```

CONTRIBUTING.md
```diff
@@ -64,3 +64,8 @@ Build and test on changes:
 ```
 make watch
 ```
+
+Build the snapshot for testing project generation:
+```
+make build-snapshot
+```
```
Shaptic, line 71: It's not immediately obvious to me how to do anything besides testing the fixtures - can we get some docs in the README on how to use this?

Shaptic:
I'm getting an error running the test suite:

```
$ cargo test --package soroban-spec --lib -- gen::typescript::boilerplate::test::build_package --exact --nocapture --ignored                                                  141 ↵
   Compiling soroban-spec v0.8.5 (/home/george/stellar/rs-soroban-sdk/soroban-spec)
    Finished test [unoptimized + debuginfo] target(s) in 1.94s
     Running unittests src/lib.rs (target/debug/deps/soroban_spec-f14147b742883a90)

running 1 test
About to write
thread 'gen::typescript::boilerplate::test::build_package' panicked at 'called `Result::unwrap()` on an `Err` value: Parse(NonZeroPadding)', soroban-spec/src/gen/typescript/boilerplate.rs:118:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test gen::typescript::boilerplate::test::build_package ... FAILED

failures:

failures:
    gen::typescript::boilerplate::test::build_package

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p soroban-spec --lib`
```

is there a prep step I'm missing? also this should probably be run as part of CI

soroban-spec/fixtures/ts/src/convert.ts
willemneal: Hmmm well I don't have an obvious solution to this.  I think the whole convert piece will be changed in the next release.

CONTRIBUTING.md
willemneal: Sure, and currently the only test is a diff on the fixture.

soroban-spec/fixtures/ts/src/convert.ts
Shaptic: Yeah def, just wanted to call this out for when stellar/js-stellar-base#620 gets in

soroban-spec/fixtures/ts/src/index.ts
willemneal: Sometimes the response returns a result type. So this allows wrapping the value with `Ok`.

willemneal:
@Shaptic Do you mean you get this when you do `make test`? You should only need to rebuild the fixture when the template code changes in the `src/typescript/project_template`.

Shaptic:
:facepalm: doh! After that, the test passes, but the `make test` itself failed:

```
failures:
---- gen::typescript::boilerplate::test::test_project_dir_location stdout ----
thread 'gen::typescript::boilerplate::test::test_project_dir_location' panicked at 'assertion failed: `(left == right)`
  left: `12`,
 right: `1`: [DirEntry("/tmp/.tmph5sPEz"), DirEntry("/tmp/.tmph5sPEz/.gitignore"), DirEntry("/tmp/.tmph5sPEz/README.md"), DirEntry("/tmp/.tmph5sPEz/package.json"), DirEntry("/tmp/.tmph5sPEz/src"), DirEntry("/tmp/.tmph5sPEz/src/constants.ts"), DirEntry("/tmp/.tmph5sPEz/src/convert.ts"), DirEntry("/tmp/.tmph5sPEz/src/env.d.ts"), DirEntry("/tmp/.tmph5sPEz/src/index.ts"), DirEntry("/tmp/.tmph5sPEz/src/invoke.ts"), DirEntry("/tmp/.tmph5sPEz/src/server.ts"), DirEntry("/tmp/.tmph5sPEz/tsconfig.json")].len() != [DirEntry("./fixtures/ts")].len()', soroban-spec/src/gen/typescript/boilerplate.rs:162:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Another obvious mistake on my part?

willemneal:
@Shaptic It seems that the snapshot needed to be rebuilt. Just pushed

soroban-spec/fixtures/ts/src/index.ts
willemneal: Removed the extra spaces!

willemneal:
Any ideas on how to fix the [deny issue:](https://github.com/stellar/rs-soroban-sdk/actions/runs/5193038580/jobs/9363030577?pr=871) 
```
error[duplicate]: found 2 duplicate entries for crate 'aho-corasick'
  ┌─ /github/workspace/Cargo.lock:3:1
  │  
3 │ ╭ aho-corasick 0.7.20 registry+https://github.com/rust-lang/crates.io-index
4 │ │ aho-corasick 1.0.1 registry+https://github.com/rust-lang/crates.io-index
  │ ╰────────────────────────────────────────────────────────────────────────^ lock entries
  │  
  = aho-corasick v0.7.20
    └── globset v0.4.10
        └── ignore v0.4.20
            └── globwalk v0.8.1
                └── assert_fs v1.0.13
                    └── (dev) soroban-spec v0.8.5
                        ├── (dev) soroban-sdk v0.8.5
                        │   ├── soroban-token-sdk v0.8.5
                        │   ├── test_add_i128 v0.8.5
                        │   ├── test_add_u128 v0.8.5
                        │   ├── test_add_u64 v0.8.5
                        │   ├── (dev) test_alloc v0.8.5
                        │   ├── test_auth v0.8.5
                        │   ├── test_contract_data v0.8.5
                        │   ├── test_empty v0.8.5
                        │   ├── test_empty2 v0.8.5
                        │   ├── test_errors v0.8.5
                        │   ├── test_events v0.8.5
                        │   ├── test_import_contract v0.8.5
                        │   ├── test_invoke_contract v0.8.5
                        │   ├── test_logging v0.8.5
                        │   └── test_udt v0.8.5
                        └── soroban-sdk-macros v0.8.5
                            └── soroban-sdk v0.8.5 (*)
  = aho-corasick v1.0.1
    └── regex v1.8.1
        ├── globset v0.4.10
        │   └── ignore v0.4.20
        │       └── globwalk v0.8.1
        │           └── assert_fs v1.0.13
        │               └── (dev) soroban-spec v0.8.5
        │                   ├── (dev) soroban-sdk v0.8.5
        │                   │   ├── soroban-token-sdk v0.8.5
        │                   │   ├── test_add_i128 v0.8.5
        │                   │   ├── test_add_u128 v0.8.5
        │                   │   ├── test_add_u64 v0.8.5
        │                   │   ├── (dev) test_alloc v0.8.5
        │                   │   ├── test_auth v0.8.5
        │                   │   ├── test_contract_data v0.8.5
        │                   │   ├── test_empty v0.8.5
        │                   │   ├── test_empty2 v0.8.5
        │                   │   ├── test_errors v0.8.5
        │                   │   ├── test_events v0.8.5
        │                   │   ├── test_import_contract v0.8.5
        │                   │   ├── test_invoke_contract v0.8.5
        │                   │   ├── test_logging v0.8.5
        │                   │   └── test_udt v0.8.5
        │                   └── soroban-sdk-macros v0.8.5
        │                       └── soroban-sdk v0.8.5 (*)
        └── ignore v0.4.20 (*)
```

willemneal:
Hopefully fixed by switching to tempdir.

willemneal:
I updated the branch and it even shows it on GH, but it is not showing up here for me.

willemneal:
Just changed the target branch to itself and that seemed to fix it.

soroban-spec/fixtures/ts/src/invoke.ts
```diff
@@ -0,0 +1,159 @@
...
+    if (auths.length > 1) {
+      throw new NotImplementedError("Multiple auths not yet supported")
+    }
+
+    const auth = SorobanClient.xdr.ContractAuth.fromXDR(auths[0], 'base64')
+
+    if (auth.addressWithNonce() !== undefined) {
+      throw new NotImplementedError(
+        `This transaction needs to be signed by ${auth.addressWithNonce()
+        }; how do we do that?`
```
paulbellamy, line 80:
```suggestion
        }; Not yet supported`
```

soroban-spec/fixtures/ts/src/invoke.ts
```diff
@@ -0,0 +1,159 @@
...
+    networkPassphrase: NETWORK_PASSPHRASE,
+  })
+    .addOperation(contract.call(method, ...args))
+    .setTimeout(SorobanClient.TimeoutInfinite)
+    .build()
+
+  const simulated = await Server.simulateTransaction(tx)
+  const auths = simulated.results?.[0]?.auth
+  // is it possible for `auths` to be present but empty? Probably not, but let's be safe.
+  if (signAndSend) {
```
paulbellamy, line 70: Suggestion: We could clean this up to be a bit more readable, by inverting this:
```suggestion
  if (!signAndSend) {
    const { results } = simulated
    if (!results || results[0] === undefined) {
      if (simulated.error) {
        throw new Error(simulated.error as unknown as string)
      }
      throw new Error(`Invalid response from simulateTransaction:\n{simulated}`)
    }
    return results[0]
  }
  
  const auths = simulates.results?.[0]?.auth
  // .. The rest of signing & sending
```

paulbellamy (approved):
A couple optional formatting/text suggestions, but LGTM.

leighmcculloch (approved):

leighmcculloch:
Approved and merged, but I think we should plan a couple follow up tasks:

- [ ] This change needs opening against `main`. We won't commit a merge commit of `release/v0.8.5` into `main` because we don't use merge commits in this repository. Someone can open a merge as a PR then we'll squash and merge that commit into `main`. Or the below items could be visited and the final product merged into `main`.

- [ ] Break up the `soroban-spec` crate into `soroban-spec`, `soroban-spec-rust`, `soroban-spec-json`, `soroban-spec-typescript`. Why: Having typescript dependencies in the soroban-sdk repo is going to make development here pretty complex, and ultimately the typescript generation code needs a unique set of reviewers/owners who are different to those of the rest of the code here. We can house the `soroban-spec-typescript` crate in the `soroban-tools` repo.

- [ ] Remove the dependence on the JSON types so that conversion from XDR spec to typescript is `ScSpecEntry` -> TypeScript, and not `ScSpecEntry` -> JSON -> TypeScript. Why: For the reasons mentioned at https://github.com/stellar/rs-soroban-sdk/pull/871#discussion_r1213585760, https://github.com/stellar/rs-soroban-sdk/pull/871#discussion_r1213584250, and https://github.com/stellar/rs-soroban-sdk/pull/871#pullrequestreview-1462963564.

cc @paulbellamy @tsachiherman

PR #881, opened 2023-03-03
Rough first pass at generating typescript contract bindings

paulbellamy:
### What

A rough first pass at generating typescript bindings for the contract spec.

### Why

This approach is higher maintenance (as the js sdk etc changes), and we can't reuse the rust strval.rs code. But, it also avoids all the rust->wasm->js machinery.

With this, you can do:
```typescript
import SorobanClient from 'soroban-client';
const xdr = SorobanClient.xdr;

// Import our generated binding for the hello world contract
import { Contract } from './contract';

// And we get an easy helper which can parse native js types to build our xdr.Operation
const contract = new Contract(contractId);
const operation = contract.hello("world");

// The rest of the txn sending flow is the same as before.
const server = new SorobanClient.Server('http://localhost:8000/soroban/rpc', { allowHttp: true });
const source = await server.getAccount(accountId);
const txn = new SorobanClient.TransactionBuilder(source, {
  fee: 100,
  networkPassphrase: SorobanClient.Networks.STANDALONE,
  v1: true,
})
  .addOperation(operation)
  .setTimeout(SorobanClient.TimeoutInfinite)
  .build();
console.debug("Transaction:", txn.toEnvelope().toXDR('base64').toString());
const result = await server.simulateTransaction(txn);
console.debug(result);
```

### Known limitations

- Only a few primitive types implemented so far, not UDTs, etc...
- We should push the parser code into js-stellar-base, then emit something like `const val_xdr = xdr.ScVal.fromJSON(js_obj, spec)` and `xdr.ScVal.toJSON(...)` for each.
  - Pro: Would simplify the rust parser emission
  - Con: Would spread the parser logic across two repos.
  - Pro: Would allow js apps to use the parsers as well.
- Future: Should also figure out how to generate parsers for the operation result values.

### DIFF
--- a/soroban-spec/src/gen/typescript.rs
+++ b/soroban-spec/src/gen/typescript.rs
@@ -51,6 +51,7 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
     let header = r#"import * as SorobanClient from "soroban-client";
 let xdr = SorobanClient.xdr;
 
+// TODO: Move all the non-trivial conversions to js-stellar-base and use them them from there.
 function bigintToI128(value: bigint): SorobanClient.xdr.ScVal {
   const buf = bigintToBuf(value);
   if (buf.length > 16) {
@@ -194,7 +195,6 @@ fn generate_method(method: &ScSpecFunctionV0, _spec: &[ScSpecEntry]) -> String {
         ));
     }
 
-    let names = arg_names.join(", ");
     let types = arg_types.join(", ");
     let parsers = arg_parsers.join("\n");
 
--- a/soroban-spec/src/gen/typescript.rs
+++ b/soroban-spec/src/gen/typescript.rs
@@ -1,0 +1,148 @@
+pub mod types;
+use std::{fs, io};
+
+use sha2::{Digest, Sha256};
+use stellar_xdr::{ScSpecEntry, ScSpecFunctionV0};
+
+use crate::read::{from_wasm, FromWasmError};
+use types::{generate_type_ident, generate_type_parser};
+
+#[derive(thiserror::Error, Debug)]
+pub enum GenerateFromFileError {
+    #[error("reading file: {0}")]
+    Io(io::Error),
+    #[error("sha256 does not match, expected: {expected}")]
+    VerifySha256 { expected: String },
+    #[error("parsing contract spec: {0}")]
+    Parse(stellar_xdr::Error),
+    #[error("getting contract spec: {0}")]
+    GetSpec(FromWasmError),
+}
+
+pub fn generate_from_file(
+    file: &str,
+    verify_sha256: Option<&str>,
+) -> Result<String, GenerateFromFileError> {
+    // Read file.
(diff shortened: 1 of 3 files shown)


soroban-spec/src/gen/typescript.rs
```diff
@@ -0,0 +1,245 @@
...
+    let spec = from_wasm(wasm)?;
+    let code = generate(&spec);
+    Ok(code)
+}
+
+pub fn generate(spec: &[ScSpecEntry]) -> String {
+    let header = r#"import * as SorobanClient from "soroban-client";
+let xdr = SorobanClient.xdr;
+
+// TODO: Move all the non-trivial conversions to js-stellar-base and use them them from there.
```
leighmcculloch, line 54: +1 As much as possible we can move utilities into a lib sgtm.

When we do that we need to think about how to version those things. A low effort approach to start with might be to include a comment at the top that says it is dependent on a specific version of the lib.

soroban-spec/src/gen/typescript.rs
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  // left-pad with zeros up to 16 bytes
+  let padded = Buffer.alloc(16);
+  buf.copy(padded, padded.length-buf.length);
+
+  if (value < BigInt(0)) {
+    // Set the top bit
+    padded[0] |= 0x80;
+  }
```
leighmcculloch, line 73: ❓ Why does the top bit get cleared and then set again during the padding process?

soroban-spec/src/gen/typescript.rs
```diff
@@ -0,0 +1,245 @@
...
+    Number(bigintFromBytes(false, ...padded.slice(12, 16))),
+    Number(bigintFromBytes(false, ...padded.slice(8, 12)))
+  );
+
+  return xdr.ScVal.scvObject(xdr.ScObject.scoU128(new xdr.Int128Parts({lo, hi})));
+}
+
+function bigintFromBytes(signed: boolean, ...bytes: (string | number | bigint)[]): bigint {
+    let sign = 1;
+    if (signed && bytes[0] === 0x80) {
```
leighmcculloch, line 111: ❗ If I'm understanding this correctly this is checking the top byte, not just the top bit. I think you need to `&` with `0x80`.

leighmcculloch (commented):
This looks like a good direction to me.

We may have similar testing issues that we have with [xdrgen](https://github.com/stellar/xdrgen), where in the repo the code is written we can't reliably test the outputs because the outputs are a different langauge to what we're building in this repo. There might be an argument for separating the `soroban-spec` crate into its own repo where we actually somehow do more elaborate testing of the generated outputs. I don't think we need to do this now though, just thinking out loud. Also, if the tests end up being simple enough, as they appear right now in this pull request, just doing string comparisons may be fine. @paulbellamy Do you have thoughts on this?

leighmcculloch:
What's the plan with this PR?

paulbellamy:
@willemneal from aha labs is working on a better version. This will be closed in favor of that.

Issue #840, opened 2023-01-26
Increase test coverage for `soroban-sdk`


jayz22:
### What problem does your feature solve?
Increase test coverage for `soroban-sdk`.
`soroban-sdk` provides building blocks for contracts. Increasing test coverage helps eliminating unexpected bugs (e.g. https://github.com/stellar/rs-soroban-sdk/pull/830 and https://github.com/stellar/rs-soroban-sdk/pull/827) before they occur. 

### What would you like to see?
Add more unit tests simliar to https://github.com/stellar/rs-soroban-sdk/pull/830 and https://github.com/stellar/rs-soroban-sdk/pull/827.
Currently most of the modules in `soroban-sdk` have limited / no unit tests. Yet, they are key building blocks for contracts which developers rely on. E.g. module `Vec`'s `push`, `pop`, `remove`, `get` are not covered by unit tests, similar problems exists for other modules such as `Map`, `Bytes`, `Set` etc. 
Too much of the sdk testing relies on integration tests or test contracts, which is limited in scope. Adding more unit tests helps eliminate unexpected bugs and reduces dependency on test contracts for bug finding.

jayz22:
@Ray-Escobar That would be great, thanks for helping!

Ray-Escobar:
Hi @jayz22 I have created various new test over some methods in Map, Bytes and Sets... I would like to continue onto other methods/modules but would first like to get some small feedback just to be sure everything is aligned, so will set up a PR soon. Does this sound good?

jayz22:
@Ray-Escobar yes that sounds good!

anupsdf:
Closing for now. Please re-open if this is important to have.

PR #900, opened 2023-03-30
Adds test for methods in vec, map and bytes (#840)

Ray-Escobar:
### What

Adds tests for the methods within the modules Vec, Map and Bytes.

Tests have been added for the following methods...

Bytes: `is_empty`, `first`, `first_unchecked`, `last`, `last_unchecked`, `get`, `get_unchecked`, `remove`, `remove_unchecked`, `pop`, `pop_unchecked`, `insert`, `slice`

Map: `from_array`, `contains_key`, `is_empty`, `get`, `get_unchecked`, `remove`, `remove_unchecked`  

Vec: `get`, `get_unchecked`, `remove`, `remove_unchecked`

### Why

These modules are the building blocks of soroban-sdk.

### Extra

I refactored some tests names in `Vec` so that they are pre-pended with the word `test`. It seems the majority of the tests follow this convention, but I also stopped short in case the renaming is not necessary. I do find its useful as an easy way to search files i.e. if you want to look for tests simply use the query "find_" on all your files.

### DIFF
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -1258,7 +1258,6 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
     fn test_first_unchecked() {
         let env = Env::default();
         let mut bin = bytes![&env, [1, 2, 3, 4]];
@@ -1266,8 +1265,12 @@ mod test {
         assert_eq!(bin.first_unchecked(), 1);
         bin.remove(0);
         assert_eq!(bin.first_unchecked(), 2);
+    }
 
-        // first on empty bytes should panic
+    #[test]
+    #[should_panic(expected = "HostError: Error(Object, IndexBounds)")]
+    fn test_first_unchecked_panics() {
+        let env = Env::default();
         let bin = bytes![&env];
         bin.first_unchecked();
     }
@@ -1287,7 +1290,6 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
     fn test_last_unchecked() {
         let env = Env::default();
         let mut bin = bytes![&env, [1, 2, 3, 4]];
@@ -1295,10 +1297,14 @@ mod test {
         assert_eq!(bin.last_unchecked(), 4);
         bin.remove(0);
         assert_eq!(bin.last_unchecked(), 3);
+    }
 
-        // first on empty bytes should panic
+    #[test]
+    #[should_panic(expected = "HostError: Error(Object, IndexBounds)")]
+    fn test_last_unchecked_panics() {
+        let env = Env::default();
         let bin = bytes![&env];
-        bin.first_unchecked();
+        bin.last_unchecked();
     }
 
     #[test]
@@ -1325,7 +1331,6 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
     fn test_get_unchecked() {
         let env = Env::default();
         let bin = bytes![&env, [0, 1, 5, 2, 8]];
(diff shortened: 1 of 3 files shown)


Ray-Escobar:
On a side note, there are some `println!` statements in lines 1144 and line 1165 in the tests for `bytes.rs`. Should I go ahead and remove them?

jayz22:
@Ray-Escobar thanks for the PR! I will take a look at the tests soon. 
As for print statements, we normally don't have them (either active or commented out) in our code base, so I would suggest removing them.

Ray-Escobar:
@jayz22 thanks! I am very open with feedback so be as critical as you need.

soroban-sdk/src/bytes.rs
```diff
@@ -1233,4 +1233,240 @@ mod test {
...
+        let env = Env::default();
+        let mut bin = bytes![&env, [1, 2, 3, 4]];
+
+        assert_eq!(bin.first_unchecked(), 1);
+        bin.remove(0);
+        assert_eq!(bin.first_unchecked(), 2);
+
+        // first on empty bytes should panic
+        let bin = bytes![&env];
+        bin.first_unchecked();
```
leighmcculloch, line 1272: This test is somewhat brittle because if any part of the test panics it'll pass, even though the only case we expect to panic is the last line. It would be ideal if we could replace the `#[should_panic]` with wrapping the `bin.first_unchecked` with a `std::panic::catch_unwind`. You will probably need to assert unwind safe because of the `Env`, so you'll probably need something like:

```suggestion
    #[should_panic]
    fn test_first_unchecked() {
        let env = Env::default();
        let mut bin = bytes![&env, [1, 2, 3, 4]];

        assert_eq!(bin.first_unchecked(), 1);
        bin.remove(0);
        assert_eq!(bin.first_unchecked(), 2);

        // first on empty bytes should panic
        let bin = bytes![&env];
        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
            bin.first_unchecked();
        }));
        assert_eq!(result, Err(...));
```

leighmcculloch (commented):
Thanks for adding tests!

If possible I'd like us to change how we assert the panics so that we're asserting on the specific function that we expect to panic. See inline comment.

soroban-sdk/src/bytes.rs
Ray-Escobar: Hi I am back, was traveling around so didn't have time to sit down. In terms of changes I changed the panic tests to catch the panic in a result variable (as you have shown here). Agree that it can be brittle as something else could trigger the panic. As for the test, I am unsure how I should test the type of panic (if necessary). I am already catching the panic and asserting that its indeed an error, as in this example:

```
fn test_first_unchecked() {
        let env = Env::default();
        let mut bin = bytes![&env, [1, 2, 3, 4]];
        assert_eq!(bin.first_unchecked(), 1);
        bin.remove(0);
        assert_eq!(bin.first_unchecked(), 2);

        // first on empty bytes should panic
        let bin = bytes![&env];
        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
            bin.first_unchecked();
        }));

        assert!(result.is_err())
    }
```

But can't find the correct way to test the type of panic. I was trying to follow how the Rust language does it ( https://github.com/rust-lang/rust/blob/a1bad57fa59a8069a6ebb05cd6a2ae73c88b2e98/src/test/run-pass/panic-uninitialized-zeroed.rs ), but this method is not working well at the point string equality is being tested.

I can commit what I have at the moment, if its evaluated as unnecessary to match the panic string message, I already changed all panic tests so as to follow the format I have posted above.

Ray-Escobar: I also deleted the redundant println messages found in the bytes file, as mentioned above.

leighmcculloch: I agree, it's pretty gnarly. I managed to get it to work by it was too big:
```rust
assert_eq!(
    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| bin.first_unchecked()))
        .map_err(|err| err
            .downcast_ref::<String>()
            .map(|s| { s.lines().into_iter().next().unwrap().to_owned() })),
    Err(Some("HostError: Error(Object, IndexBounds)".to_owned()))
);
```

In ef6c634348513525fc626d8cfbeec9c5ad31add9 I've tuned the should_panic use a bit so that the panic tests are smaller and focused on the one call.

leighmcculloch (approved):

PR #911, opened 2023-04-12
Set a random hidden source account for test Envs

leighmcculloch:
### What
Set a random source account for all test Envs that developers use in tests, that is hidden from the developer.

### Why
Make sure that the `Env` is configured more similar to how it will be for real outside tests. So that we can write code in Env more consistently assuming the source account will be set safely without causing errors in users of `Env`.

We used to always set a source account for the `Env` because in a real environment the source account is always set. I think it was removed as part of the auth-next work, although maybe unnecessarily. In the past we had run into some situations where it was surprising or unexpected for there to be no source account and errors arose. Today I ran into a situation where a function panicked because there was no source account.

It seems otherwise harmless to set a source account since developers have no visibility into it and it has no functional impact, so this change feels rather safe.

In the future we may want to make source accounts settable by a developer again, but there seems no need for it today and so a random one will do the job in just making sure the `Env` is configured in a way that is more similar to how it will be for real.

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -483,6 +483,9 @@ impl Env {
         let budget = internal::budget::Budget::default();
         let env_impl = internal::EnvImpl::with_storage_and_budget(storage, budget.clone());
         env_impl.switch_to_recording_auth();
+        env_impl.set_source_account(xdr::AccountId(xdr::PublicKey::PublicKeyTypeEd25519(
+            xdr::Uint256(random()),
+        )));
         let env = Env {
             env_impl,
             snapshot: None,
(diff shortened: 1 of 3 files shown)


dmkozh:
Just curious, which function has panicked?

I removed this to make sure we don't have dependencies on the source account; I don't think it is problematic to set in general, but it's a bit concerning that something actually still depends on it.

dmkozh (approved):

leighmcculloch:
> which function has panicked?

While I was playing around with `AuthorizationManager`, I removed the "recording" line, and saw panicks about source account not being set. Arguably I'm making modifications in SDK and not at the user level, but on another note it was really easy for me to trigger the panick, which speaks to an assumption being made that isn't enforced.

> I removed this to make sure we don't have dependencies on the source account

I think we probably need something in Env then that ensures source account is never used, or removed at that level?

PR #915, opened 2023-04-17
Remove soroban-token-spec

leighmcculloch:
### What
Remove soroban-token-spec and move some of its tests into the sdk.

### Why
It is no longer required as #912 duplicated all the functionality provided by it directly in the `soroban_sdk::token` module. Some of the tests should have been moved or copied in #912, but I missed it then, so bundling into this.


sisuresh (approved):
LGTM

PR #925, opened 2023-04-27
Make crate path configurable in contractimpl

leighmcculloch:
### What
Make crate path configurable in contractimpl

### Why
So it is possible to use the contractimpl macro in places where the crate path needs to be modified.

Other macros already support configuring crate path, so this brings the contractimpl macro up to date.

### DIFF
--- a/soroban-sdk-macros/src/derive_fn.rs
+++ b/soroban-sdk-macros/src/derive_fn.rs
@@ -5,11 +5,12 @@ use syn::{
     punctuated::Punctuated,
     spanned::Spanned,
     token::{Colon, Comma},
-    Attribute, Error, FnArg, Ident, Pat, PatIdent, PatType, Type, TypePath,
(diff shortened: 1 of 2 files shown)


sisuresh (approved):
LGTM! This change would allow you to use the macros in `soroban-sdk` right? I'm curious if there are other use cases to be aware of.

leighmcculloch:
@sisuresh I could have been more specific and detailed on the use cases, there are several.

1. As your suggested, it allows the macro to be used inside the SDK, such as in the testicles to provide stock contacted for use in tests. 

2. Other SDKs that would like to embed the soroban SDK such as the one @willemneal is working on would be able wrap the macro in another macro and use a modified crate path to reference the reexported soroban SDK, without users of that wrapping SDK needing to import the soroban SDK themselves. 

My immediate use cases is the first.

Issue #917, opened 2023-04-20
Auth Next: Support enforced auth and real auth in tests


leighmcculloch:
Independent of #875 which is discussing a convenient test specific interface for working with auth, we need to allow support for using auth in its rawest form in the SDK tests, where enforcing is enabled and the same tools are available as will be available outside of tests.

While it's critical that we provide a simple interface for mocking, stubbing, or testing against auth without actually performing auth (the topic of #875), we also need to provide a way for developer to play with custom accounts in enforcing mode, where the actual contracts execute, and where developers can explore much like a playground the full scope of how auth works on a real network.

A goal of the test framework in the SDK is that to the largest extent possible it operates just like a contract will experience on the real network. In many ways it is a playground that lets a user discover and learn about Soroban, and auth is a big component of that.

Adding this capability should interact very little with whatever we plan to do in #875 because running in enforcing mode should just be a matter of not enabling the mocks/stubs/conveniences that #875 provides.

dmkozh:
While I'm not strictly opposed to this in concept, there is a few things to keep in mind:

- the enforcing mode requires passing the `ContractAuth` payloads to the host. these payloads have to have the call trees of `require_auth`
- it also requires the correct accounting of nonces (though this might be simplified with temp storage nonces, https://github.com/stellar/rs-soroban-env/issues/655
- it requires the ledger state for the account, classic or custom, depending on what we want to cover here. the ledger state for the accounts would also result in additional ledger state required for SAC
- it requires the proper signatures

Keeping all these requirements in mind, I think the best way to cover this is to do that via sending transactions to the core, i.e. that would be an integration test (as in 'integration between dapp and stellar', not 'integration between contracts'). If we strive to emulate the real behavior and use the state external to the control invocation (such as the auth payload/sigantures/nonces), we should probably also take the footprints and resource consumption into account as well. This way the developers will be able to actually test the wallet interactions, auth payload building, footprint building etc.

I'm not sure about building this into the SDK directly though. It would still need to emulate a big chunk of the logic (unless we want the users to write boilerplate for the sake of boilerplate), while not providing any additional coverage. E.g. if we have a custom, SDK-implemented builder of the auth payloads, then it will be only usable by the SDK tests for the SDK tests. It doesn't help any real use case (as in reality XDR will have to be used) and it doesn't cover anything (because auth payloads are managed outside of the SDK).

PR #926, opened 2023-04-27
Default to enforced auth in tests

leighmcculloch:
### What

Default to enforced auth in tests.

Also renamed 'record' to 'mock' so clarity that when in use, these functions are mocking auth.

Also as a side-effect the auths() fn that returns a list of seen auths during an invocation now works for enforced auth too.

### Example

```rust
#[contractimpl]
impl Contract {
    pub fn doit(a: Address) -> u64 {
        a.require_auth();
        2
    }
}
```
```rust
#[test]
fn test_with_mock_auth() {
    let e = Env::default();
    
    e.mock_all_auths()

    let contract_id = e.register_contract(None, Contract);

    let a = Address::random(&e);

    let r = client.doit(&a);

    assert_eq!(r, 2);
}
```

### Why

The test framework doesn't allow the developer to use auth like it will behave in production. As much as possible a principle of the test framework is that it behaves like production. While we have and are improving functions for mocking auth such that testing with auth is convenient, the test framework should also be able to behave like production.

Dependent on https://github.com/stellar/rs-soroban-env/pull/787

Close https://github.com/stellar/rs-soroban-sdk/issues/917

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -761,9 +761,9 @@ impl Env {
     /// When mocking is enabled, if the [`Address`] being authorized is the
     /// address of an account, the account does not need to exist.
     ///
-    /// To disable mocking, see [`set_auth`].
+    /// To disable mocking, see [`set_auths`].
     ///
-    /// To access a recording of mocked auths, see [`mocked_auths`].
+    /// To access a list of auths that have occurred, see [`auths`].
     ///
     /// It is not currently possible to mock a subset of auths.
     ///
@@ -799,8 +799,8 @@ impl Env {
         self.env_impl.switch_to_recording_auth();
     }
 
-    /// Returns a list of calls to [`require_auth`] or [`require_auth_for_args`] that
-    /// were mocked during the last contract invocation.
+    /// Returns a list of authorizations that were seen during the last contract
+    /// invocation.
     ///
     /// Use this in tests to verify that the expected authorizations with the
     /// expected arguments are required.
@@ -846,7 +846,7 @@ impl Env {
     ///     let address = Address::random(&env);
     ///     client.transfer(&address, &1000_i128);
     ///     assert_eq!(
-    ///         env.mocked_auths(),
+    ///         env.auths(),
     ///         std::vec![(
     ///             address.clone(),
     ///             client.contract_id.clone(),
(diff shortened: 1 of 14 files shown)


soroban-sdk/src/env.rs
```diff
@@ -750,20 +811,10 @@ impl Env {
...
-    /// to coincide with the actual top-level contract invocation. For example,
-    /// if contract A doesn't use `require_auth` and then it calls contract B
-    /// that uses `require_auth`, then `verify_top_authorization` will return
-    /// `true` when verifying the contract B's `require_auth` call, but it will
-    /// return `false` if contract A makes a `require_auth` call.
-    ///
-    /// It is possible for a single address to be present multiple times in the
-    /// output, as long as there are multiple disjoint call trees for that
-    /// address.
+    /// [`require_auth`] calls. Repeated calls to [`require_auth`] in the same
```
dmkozh, line 814: > Repeated calls to [`require_auth`] in the same tree of contract invocations will appear only once in the vector

This is not correct. First, trees are relative to the address, so for multiple addresses multiple entries will be present for the same contract invocation. Second, it's actually possible to call `reqiure_auth` for the same address multiple times in the same call - that's fine as long as the arguments are different.

soroban-sdk/src/env.rs
```diff
@@ -737,8 +736,70 @@ impl Env {
...
+    ///     let client = HelloContractClient::new(&env, &contract_id);
+    ///     let addr = Address::random(&env);
+    ///     client.hello(&addr);
+    /// }
+    /// ```
+    pub fn mock_all_auths(&self) {
+        self.env_impl.switch_to_recording_auth();
+    }
+
+    /// Returns a list of authorizations that were seen during the last contract
```
dmkozh, line 801: These are still 'top' authorizations, but the comment doesn't really refer to that. While the previous definition probably had some room to improve the clarity, it contained information about what the 'top' means. That's important as e.g. internal token transfer calls won't be visible here and this is probably the most frequent scenario for the auth trees.

soroban-sdk/src/env.rs
leighmcculloch: Fixed in d66eb844 and 53266a03. Thanks!

soroban-sdk/src/env.rs
leighmcculloch: Thanks for identifying this gap in the docs. I'd like to address the gap by removing top-level auths from the SDK for this function, to simplify the concepts a developer has to interact with. Opened https://github.com/stellar/rs-soroban-env/pull/787 and updated this pull request to use it instead. This PR is blocked on that env change.

Issue #875, opened 2023-02-24
Auth Next: Revisit testing interface


dmkozh:
Current testing interface might be a bit confusing due to the fact that we provide the auth records for the most recent contract call. We could signify that by chaining the auth check with the contract invocation, for example:

- `client.expect_auth(...).xfer(...)` - specify expected auth entries before invocation. This could be a bit confusing though in case if a contract method fails for a different reason
- `env.record_auth(|| client.xfer(..))` - this explicitly specifies the intent of verifying auth and provides more clear errors (as `assert_eq` should be used on to of this). The downside is that the contract call should happen in the closure.

tomerweller:
For some added context: 
1. When invoking a contract in local-testing mode all `require_auth()` checks automatically succeed and 
2. The successful authorization of the last invocation can be accessed through `env.recorded_top_authorizations()` 

Examples: [Here](https://github.com/stellar/soroban-examples/blob/v0.6.0/auth/src/test.rs#L16:L31) is how this is used in the official auth example. And [here](https://github.com/tomerweller/tinynft/blob/master/src/test.rs#L24:L26) is a more minimal version from my own tinynft implementation

@dmkozh, your proposal seems to tackle the idiosyncratic nature of handling 2 but I think that 1 is also a problem: the idea that all auth checks succeed without providing explicit auth is unintuitive. 

invoker-auth (auth next's predecessor) required explicit authorization using `client.with_source_account()` ([see previous auth example](https://github.com/stellar/soroban-examples/blob/71c07f0f40f4b06100df6971df0d03d003ce9877/auth/src/test.rs#LL16C37-L16C37)), which I found to be more intuitive. 


### Alternative proposal 

For auth next I propose that authorizations won't automatically succeed and will be provided explicitly, similar to how invoker auth used to work. Practically this could looks something like these three versions:

``` rust
//authorization is forced to succeed  
assert_eq!(client.with_auth_success().increment(&user_2, &4), 5);

//authorization is explicitly provided, checks only pass if there's a match
let user1_auth = std::vec![(
    user_1.clone(),
    contract_id.clone(),
    symbol!("increment"),
    (user_1.clone(), 5_u32).into_val(&env)
)];
assert_eq!(client.with_auth(user1_auth), increment(&user_1, &5), 7);

//authorization is checked against a list of Addresses (for brevity)  
assert_eq!(client.with_auth_accounts([&user_1]), increment(&user_1, &5), 7);
```

[...]

PR #929, opened 2023-04-28
Add granular mocking of auth in tests

### LINKED ISSUE #917: Auth Next: Support enforced auth and real auth in tests

leighmcculloch:
### What

Add granular mocking of auth in tests, as a capability that is available to developers in addition to the existing mock-everything approach.

### Example

```rust
#[contractimpl]
impl Contract {
    pub fn doit(a: Address) -> u64 {
        a.require_auth();
        2
    }
}
```
```rust
#[test]
fn test_with_mock_auth() {
    let e = Env::default();

    let contract_id = e.register_contract(None, Contract);

    let a = Address::random(&e);

    let r = client
        .mock_auths(&[MockAuth {
            address: &a,
            nonce: 0,
            invoke: &MockAuthInvoke {
                contract: &contract_id,
                fn_name: "add",
                args: (&a,).into_val(&e),
                sub_invokes: &[],
            },
        }])
        .doit(&a);

    assert_eq!(r, 2);
}
```

### Why

The test framework doesn't allow the developer to use auth like it will behave in production. As much as possible a principle of the test framework is that it behaves like production. While we have and are improving functions for mocking auth such that testing with auth is convenient, the test framework should also be able to behave like production.

Also, we should provide a way to conveniently mock auth, but in a way that still causes the developer to understand how auth works. This change presents auth to the user in the same way they must engage auth when invoking contracts for real, by requiring them to define what call tree is authorized for an address. However, the actual production of signatures is not required, eliminating the repetitive boilerplate code we've seen in the past of generating ed25519 keys, signing blobs, etc.

Including the call trees is important so that developers understand that on Soroban auth is auth of a call tree, and not just an individual contract. One downside of the existing "top-level only" recording / spying in tests is that the concept of call trees is invisible to developers.

In regards to #875 this does not implement everything one might find convenient in a mocking framework. I don't think we need everything, at least not everything on day one, and what is included in this change is a tradeoff, providing basic mocking functionality at the benefit of being able to do so with no modifications to how auth works in the `rs-soroban-env` repo. In the future it would be great to build the features out and to make it more feature rich as needed, but doing so will probably require modifications to `rs-soroban-env`.

Close #917 #875 

Dependent on https://github.com/stellar/rs-soroban-env/pull/780


soroban-sdk-macros/src/derive_client.rs
```diff
@@ -118,10 +124,16 @@ pub fn derive_client(crate_path: &Path, ty: &str, name: &str, fns: &[syn_ext::Fn
         pub struct #client_ident<'a> {
             pub env: #crate_path::Env,
             pub contract_id: #crate_path::BytesN<32>,
+            #[doc(hidden)]
```
leighmcculloch, line 127: Slipping this doc hidden in so that in so that in IDEs these generated non-public fields don't display in auto-complete for developers. Even though they aren't public, because this code gets generated inside a developer's module, the developer can access the private fields and them showing up in auto-completes is very annoying.

soroban-sdk/src/testutils/mock_auth.rs
```diff
@@ -0,0 +1,69 @@
...
+#[contractimpl(crate_path = "crate")]
+impl MockAuthContract {
+    #[allow(non_snake_case)]
+    pub fn __check_auth(_signature_payload: RawVal, _signatures: RawVal, _auth_context: RawVal) {}
+}
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct MockAuth<'a> {
+    pub address: &'a Address,
+    pub nonce: u64,
```
dmkozh, line 17: Can we get rid of the nonce here? It's not a relevant detail and it can be easily loaded from the ledger using `read_nonce` host fn. This would not let you use `From` conversions, but I think that's fine.

soroban-sdk/src/env.rs
```diff
@@ -683,14 +689,43 @@ impl Env {
...
 
-        let token_id = self.env_impl.invoke_functions(vec![create]).unwrap()[0]
+        let token_id: BytesN<32> = self.env_impl.invoke_functions(vec![create]).unwrap()[0]
             .try_into_val(self)
             .unwrap();
+
+        let source_account = self.env_impl.source_account();
+        self.env_impl.set_source_account(issuer_id);
+
+        // Recording auth is used with set_admin because account auth is
```
dmkozh, line 699: The comment doesn't correspond to what happens here - we seem to be using enforcing mode with invoker auth.

leighmcculloch: Thanks for catching that. I used recording auth then switched to using the actually issuer since the auth was simple and convenient to construct.

soroban-sdk/src/testutils/mock_auth.rs
leighmcculloch: Interesting. It's not obvious to me that it's an irrelevant detail, or always predictable. I'd like to experiment with that idea in a follow up change that tries out a series of scenarios to make sure it's predictable and not relevant, after merging this pull request.

soroban-sdk/src/env.rs
dmkozh: Just a reminder, I still don't see the update.

soroban-sdk/src/testutils/mock_auth.rs
dmkozh: It's not relevant because the contract has no impact on it. Adding nonce here is basically covering the host logic, which I really want to avoid for the contract tests. The only valid nonce value currently is the one provided by `read_nonce` utilty.

I also don't want to expose this because in the following release we hopefully should be able to migrate to random temp nonces (https://github.com/stellar/rs-soroban-env/issues/655), which would make this even less relevant than it is now.

If it's really hard to change and if we don't expect much use of this during the release iteration, we probably could leave this for a bit, but I'd really like to avoid exposing this if possible.

soroban-sdk/src/env.rs
leighmcculloch: Thanks. Fixed in c17bc60.

dmkozh (approved):

soroban-sdk/src/testutils/mock_auth.rs
leighmcculloch: Issue to track addressing this: https://github.com/stellar/rs-soroban-sdk/issues/935#issue-1696489931

PR #931, opened 2023-05-02
initial version of cargo-deny

graydon:
This is ready now. Should prevent further crate duplication.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,21 @@ jobs:
     - run: rustup update
     - run: cargo fmt --all --check
 
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: EmbarkStudios/cargo-deny-action@e0a440755b184aa50374330fa75cca0f84fcb59a
+      with:
+        command: check ${{ matrix.checks }}
+
   rust-analyzer-compat:
     runs-on: ubuntu-latest
     steps:
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -22,15 +22,15 @@ use derive_spec_fn::derive_fn_spec;
 use derive_struct::derive_type_struct;
 use derive_struct_tuple::derive_type_struct_tuple;
 
-use darling::FromMeta;
+use darling::{ast::NestedMeta, FromMeta};
 use proc_macro::TokenStream;
 use proc_macro2::{Literal, Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
 use sha2::{Digest, Sha256};
 use std::fs;
 use syn::{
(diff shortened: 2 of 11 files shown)


.github/workflows/rust.yml
```diff
@@ -25,6 +25,20 @@ jobs:
...
     - run: cargo fmt --all --check
 
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories bans licenses sources
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
leighmcculloch, line 35: Is the goal to run this job four times in parallel, one time for each of the words listed here? If so, they need to be an array. Right now it is running one job with the `checks` being one value, this long value.

```suggestion
    strategy:
      matrix:
        checks: ["advisories", "bans", "licenses", "sources"]
    # Prevent sudden announcement of a new advisory from failing ci:
    continue-on-error: ${{ matrix.checks == 'advisories' }}
```

```suggestion
    strategy:
      matrix:
        checks:
        - advisories
        - bans
        - licenses
        - sources
    # Prevent sudden announcement of a new advisory from failing ci:
    continue-on-error: ${{ matrix.checks == 'advisories' }}
```

graydon: Oh, huh. So .. the example I copied this from had `advisories` in one matrix entry, and `bans licenses sources` in another. And I think the `continue-on-error` line there is designed with that organization in mind. I'll set it back to that (it works in all cases -- you can pass multiple checks on the same invocation -- but the idea with the 2-entry organization is to allow the `advisory` version not to block CI since it happens asynchronously, outside the user's control)

soroban-sdk-macros/src/lib.rs
```diff
@@ -50,7 +50,12 @@ struct ContractSpecArgs {
 
 #[proc_macro_attribute]
 pub fn contractspecfn(metadata: TokenStream, input: TokenStream) -> TokenStream {
-    let args = parse_macro_input!(metadata as AttributeArgs);
+    let args = match NestedMeta::parse_meta_list(metadata.into()) {
```
leighmcculloch, line 53: Why the switch from using syn's `parse_macro_input!` to using darling's `parse_meta_list`?

soroban-spec/Cargo.toml
```diff
@@ -14,17 +14,17 @@ rust-version = "1.69"
...
 thiserror = "1.0.32"
-syn = {version="1.0",features=["full"]}
+syn = {version="2.0",features=["full"]}
 quote = "1.0"
 proc-macro2 = "1.0"
 itertools = "0.10.3"
-darling = "0.14.1"
-sha2 = "0.10.2"
+darling = "0.20.0"
+sha2 = "0.9.9"
```
leighmcculloch, line 22: Why the downgrade of sha2?

soroban-sdk-macros/src/lib.rs
graydon: `AttributeArgs` went away -- internal changes in `syn`

soroban-spec/Cargo.toml
graydon: de-duplication -- dalek is on an older one. the differences are harmless. we'll move everything ahead once dalek finally publishes 2.0 (_very_ hopefully this summer before we go final, but hard to predict, they just came out of neglected-maintenance)

leighmcculloch (approved):

PR #940, opened 2023-05-14
Err reform

graydon:
This is the companion PR to https://github.com/stellar/rs-soroban-env/pull/806 (and https://github.com/stellar/rs-stellar-xdr/pull/258 and https://github.com/stellar/stellar-xdr/pull/92) overhauling error codes and renaming Status to Error.


soroban-sdk/src/logging.rs
```diff
@@ -61,16 +60,11 @@ use crate::{
...
 /// use soroban_sdk::testutils::Logger;
-///
-/// assert_eq!(
-///     env.logger().all(),
-///     std::vec![
-///         "a log entry: I32(5), Symbol(another)".to_string(),
-///     ],
-/// );
+/// let logentry = env.logger().all().last().unwrap().clone();
+/// assert!(logentry.contains("[\"a log entry\", 5, another]"));
```
leighmcculloch, line 67: The changes to this test seem unnecessary and result in a testing approach that panics when no life are outputted rather than a comparison failure which is easier to debug.

soroban-sdk/src/address.rs
```diff
@@ -129,15 +129,15 @@ impl TryFromVal<Env, &Address> for RawVal {
 #[cfg(not(target_family = "wasm"))]
 impl TryFrom<&Address> for ScVal {
     type Error = ConversionError;
-    fn try_from(v: &Address) -> Result<Self, Self::Error> {
+    fn try_from(v: &Address) -> Result<Self, ConversionError> {
```
leighmcculloch, line 132: What's the reason for all the changes from referencing the associated type to explicitly stating it? I don't see any functional change and so it seems like a matter of style. It looks like a large amount of the diff, so wondering why were going to that length.

tests/logging/src/lib.rs
```diff
@@ -45,17 +44,17 @@ mod test {
...
+            let pats = std::vec![
+                "\"none\"",
+                "\"none\"",
+                "[\"one:\", one]",
+                "[\"one:\", one]",
+                "[\"one and two:\", one, two]",
+                "[\"one and two:\", one, two]"
+            ];
+            for (msg, pat) in env.logger().all().iter().zip(pats.iter()) {
+                assert!(msg.contains(pat));
```
leighmcculloch, line 56: Same reason as above, why did the assertion change from a comparison, which is easy to debug, to a boolean assertion, which is harder to inspect?

If something about diagnostic errors is driving to this more limited partial testing I'd like to see how we can tweak it to allow for complete comparison of logs.

leighmcculloch (commented):
👏 Few questions inline.

soroban-sdk/src/logging.rs
graydon: `impl Display for HostEvent` produces a larger thing now -- a pretty but fairly long string -- that includes the contract ID, which due to https://github.com/stellar/rs-soroban-env/issues/810 isn't a stable string, so we can at this point only do a match on it.

We can revise this, possibly by introducing a special accessor / stringifier for just the data field of the event (eg. expose the pretty-printer for ScVal) but that'll take another round of changes to the env, where the pretty-printer is. I figured this was an acceptable stopping point for this PR -- the purpose of which is just to catch up with the env change -- but I'm happy to fine-tune the way strings look in followups.

soroban-sdk/src/address.rs
graydon: `Self::Error` is ambiguous between `TryFrom<_>::Error` and `ScVal::Error`, now that the latter is itself a defined symbol. Every change in the PR is just a fix to a compile error, nothing here is for style.

tests/logging/src/lib.rs
graydon: Same as above. Happy to tweak string formatting, but prefer to do it in a followup so that SDK can at least compile and run in the meantime.

Issue #868, opened 2023-02-17
Auth Next: Contract IDs use `Address` not `BytesN<32>`


dmkozh:
This probably should be done after/together with https://github.com/stellar/rs-soroban-env/issues/692 (address conversion host fns)

There are two ways to approach this:

- Host-side: change all the respective host fns to use `Address`
- SDK-side: leave the host as is and use the conversion functions from https://github.com/stellar/rs-soroban-env/issues/692 in SDK to convert from contract `Address` to `BytesN<32>`

leighmcculloch:
Cross posting from Discord: https://discord.com/channels/897514728459468821/1103075731715653672

+100 for us changing all uses of `BytesN<32>` and contract IDs to using `Address`.

If we end up succeeding at true account abstraction, in the end an Address that is a contract ID should be usable anywhere any Address is usable and the importance of contract vs account addresses should fade away even more.

I'd like to see these these places using `Address` instead of `BytesN<32>`:
 ● Contract clients must be instantiated with a BytesN<32>.
 ● Registering contracts in tests returns a BytesN<32>.
 ● Passing around contract IDs between contracts in our examples use BytesN<32>.
 ● Env call stack functions. Possibly it would even make sense to have the source account be a frame in the call stack 🤔.
 
I'm mainly in support of this because by proliferating BytesN<32> about in so many ways, the developer has to learn about and deal with both Address and BytesN<32> and their conversion. When coding myself I find myself doing conversions between them as a matter of necessity and their distinction isn't adding much value to the code I write – it feels like busy work.

It'd especially clean up the testing API. `register_contract` would return an Address, that could immediately be used as an Address to be authorized.

---

In terms of host-side vs sdk-side, I think we should do this host-side. It will result in smaller WASM for the common case where a contract does not need a `BytesN<32>` for the contract ID.

leighmcculloch:
Given that the host-side will be more work, we could do SDK-side earlier using the new conversion functions that were implemented, then update the SDK to be use the host-side when that change is made. Wdyt @dmkozh?

dmkozh:
> we could do SDK-side earlier using the new conversion functions that were implemented, then update the SDK to be use the host-side when that change is made

[...]

PR #947, opened 2023-05-17
Replace BytesN<32> for contract IDs with Address

leighmcculloch:
### What
Replace BytesN<32> for contract IDs with Address.

### Why
Exposing contract IDs as BytesN<32> is too low level and results in contracts binding themselves to an implementation detail. Contracts are deployed to an Address, and accounts are also an Address, and as much as possile contract developers shouldn't think about these differently, especially once folks start using contracts for authentication.

In all cases we currently use BytesN<32> for a contract ID we can trivially replace it with an Address.

The functions remain for building Addresses from BytesN purely for test capabilities, because in some edge case situations it will be necessary to define a contract ID in a test.

Close https://github.com/stellar/rs-soroban-sdk/issues/868

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -408,9 +408,12 @@ impl Env {
         T: TryFromVal<Env, RawVal>,
         E: TryFrom<Status>,
     {
+        let Some(contract_id) = contract_id.try_contract_id() else {
+            return Err(E::try_from(Status::from_status(xdr::ScStatus::UnknownError(xdr::ScUnknownErrorCode::General))));
+        };
         let rv = internal::Env::try_call(
             self,
-            contract_id.contract_id().to_object(),
+            contract_id.to_object(),
             func.to_val(),
             args.to_object(),
         )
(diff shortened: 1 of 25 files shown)


soroban-sdk/src/env.rs
```diff
@@ -408,7 +410,7 @@ impl Env {
     {
         let rv = internal::Env::try_call(
             self,
-            contract_id.to_object(),
+            contract_id.contract_id().to_object(),
```
leighmcculloch, line 413: TODO: Catch the error when Address is not a contract. This line is a problem. Try_invoke should never panic, but the call to contract_id() does panic now. I probably should have kept contract_id() as an Option<>. I'll change it back and massage the None into a general error of sorts.

leighmcculloch: Fixed in [c02d267](https://github.com/stellar/rs-soroban-sdk/pull/947/commits/c02d2671096a9b475cb6ec0b72faa0389622a961)

graydon (approved):
I have no real opinion on the ergonomics here, it looks fine to me!

PR #951, opened 2023-05-18
Keep Address::from_contract_id

leighmcculloch:
### What
Keep Address::from_contract_id for contracts to use.

### Why
There's an edge case where we still need to be able to construct an address from BytesN<32>, and that's with account contracts. The account contracts interface still exposes BytesN<32> unfortunately as that requires an environment change.

### DIFF
--- a/soroban-sdk/src/address.rs
+++ b/soroban-sdk/src/address.rs
@@ -236,7 +236,7 @@ impl Address {
     /// Prefer using the `Address` directly as input or output argument. Only
     /// use this in special cases, for example to get an Address of a freshly
     /// deployed contract.
-    pub(crate) fn from_contract_id(contract_id: &BytesN<32>) -> Self {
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/auth.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[contracttype(crate_path = "crate", export = false)]
+#[derive(Clone)]
+pub struct AuthContext {
+    pub contract: BytesN<32>,
+    pub fn_name: Symbol,
+    pub args: Vec<RawVal>,
+}
+
+impl AuthContext {
```
dmkozh, line 11: I've actually tried to expand 'auth' when reasonable. `check_auth` is an exception because the full version is `check_authentication_and_authorization`. I don't feel too strong here though.

dmkozh (approved):

Issue #268, opened 2022-07-21
Make contractimpl and derive_contract_function_set work with multiple traits


jonjove:
### What problem does your feature solve?

There are several areas of code that have introduced a restriction on the number of `impl` blocks that can appear for a type that has `contractimpl`'s.

1. The `derive_contract_function_set` will try to impl the `ContractFunctionSet` repeatedly if there are multiple `contractimpl` for the same type.

2. The `contractclient` attribute that `contractimpl` adds to the impl will try and create multiple contract client types that all have the same name.

### What would you like to see?

It should be possible to define multiple `impl` blocks both for the inherent type and for traits for a struct that is being used as a contract implementation.

### What alternatives are there?

Use different types for each impl, but this leads to an imperfect setup for testing, since only a single ContractFunctionSet can currently be registered for a contract ID. It's possible we could support the registration of multiple ContractFunctionSets.

leighmcculloch:
I had some discussions with folks in Rust Discord a couple times about the issue of storing and building state across multiple macro invocations. The linkme crate kept being suggested as a way to do it, https://github.com/dtolnay/linkme.

leighmcculloch:
We are unlikely to solve this problem in the near term, so I'm removing this from any near term iterations.

leighmcculloch:
We could support multiple traits inside a single proc-macro invocation such as:
```rust
contractimpl! {
    impl Contract {
    }

    impl Trait for Contract {
    }
}
```

PR #952, opened 2023-05-18
Add auth interface

leighmcculloch:
### What
Add auth interface.

### Why
So that contracts who wish to implement the `__check_auth` function to provide custom auth can do so by implementing a trait, which is more convenient than copy-pasting a specific function.

The reason we left this out previously was because it is not convenient for developers to implement multiple traits. However, there is a work around for this, and the work around is more convenient than getting the function right.


soroban-sdk/src/auth.rs
```diff
@@ -13,3 +15,21 @@ impl Context {
         Address::from_contract_id(&self.contract)
     }
 }
+
+/// Auth interface that a contract implements to support being used for custom
+/// auth.
+///
+/// Once a contract implements the interface, the address of the contract can be
+/// used for addresses that have [`Address::require_auth`] called.
+pub trait Interface<E>
```
dmkozh, line 24: Unlike `Context`, I feel like this could benefit from a more specific name (like `CustomAccountInterface` or something). `auth::Interface` doesn't make much sense to me.

soroban-sdk/src/auth.rs
```diff
@@ -13,3 +15,21 @@ impl Context {
...
+where
+    Status: for<'a> TryFrom<&'a E>,
+{
+    /// Check that the signatures and auth contexts are valid.
+    fn __check_auth(
+        env: Env,
+        signature_payload: BytesN<32>,
+        signatures: Vec<RawVal>,
```
dmkozh, line 32: One interesting caveat is that without the interface it's possible to do type auto-conversion via SDK and have something like `Vec<BytesN<64>>` or whatever appropriate UDT. Probably not a big deal though.

leighmcculloch: That's a good point. I changed the interface to have an associated type used for the signature type.

soroban-sdk/src/auth.rs
leighmcculloch: 👍🏻 Changed.

dmkozh (approved):

leighmcculloch:
Holding this because it's near impossible to use multiple contractimpls now. It has always been tricky, but seems like it has gotten worse. Using this is really dependent on https://github.com/stellar/rs-soroban-sdk/issues/268.

Issue #939, opened 2023-05-10
`contractimport!` emits compiler errors for contract types containing `RawVal`.


brson:
### What version are you using?

0.8.0

### What did you do?

I created a contract with a `#[contracttype]` containing `RawVal` and compiled it to wasm.
Then I tried to load it in a host environment with the `contractimport!` macro.

The contract looks like

```rust
#![no_std]

use soroban_sdk::{contractimpl, contracttype, log, Env, RawVal};

#[contracttype]
#[derive(Clone, Debug)]
pub enum FuzzInstruction {
    LogValue(RawVal),
}

pub struct FuzzContract;

#[contractimpl]
impl FuzzContract {
    pub fn run(env: Env, fuzz_instruction: FuzzInstruction) {
        match fuzz_instruction {
            FuzzInstruction::LogValue(rawval) => log!(&env, "log value: {}", rawval)
        }

    }
}
```

and the host code looks like (some of this fuzzing code is not valid on the master branch)

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;
use soroban_sdk::{Env, FromVal};
use fuzzcontract::FuzzInstruction;
use soroban_sdk::arbitrary::SorobanArbitrary;

mod fuzzcontract {
    soroban_sdk::contractimport!(file = "../target/wasm32-unknown-unknown/release/contract_for_fuzz.wasm");
}

fuzz_target!(|input: <FuzzInstruction as SorobanArbitrary>::Prototype| {
    let env = Env::default();
    
    let contract_id = env.register_contract_wasm(None, fuzzcontract::WASM);
    
    let client = fuzzcontract::Client::new(&env, &contract_id);

    let fuzz_instruction = FuzzInstruction::from_val(&env, &input);

    client.run(&fuzz_instruction)
});
```

### What did you expect to see?

When compiling the host crate, I expected the compilation to succeed.

### What did you see instead?

The host crate fails to compile with

```
error[E0277]: the trait bound `RawVal: Eq` is not satisfied
 --> fuzz_targets/fuzz_target_1.rs:9:5
  |
9 |     soroban_sdk::contractimport!(file = "../target/wasm32-unknown-unknown/release/contract_for_fuzz.wasm");
[...]

PR #957, opened 2023-05-20
Add fuzzing support with docs; RawVal comparison proptests

brson:
### What

Preliminary support for fuzzing Soroban contracts with [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz/). This patch is primarily concerned with introducing a pattern for implementing the [Arbitrary](https://docs.rs/arbitrary/latest/arbitrary/) trait, by which fuzz tests generate semi-random Rust values.

This is a new revision of previous pr https://github.com/stellar/rs-soroban-sdk/pull/878. Little has changed functionally since that pr.

This patch additionally uses the Arbitrary impls with [`proptest-arbitrary-interop` crate](https://github.com/graydon/proptest-arbitrary-interop) to add proptests that help ensure that:
RawVals and ScVals can be converted between each other and their their comparison functions are equivalent, which closes https://github.com/stellar/rs-soroban-env/issues/743.

This patch introduces the SorobanArbitrary trait which looks like this:

```rust
    pub trait SorobanArbitrary:
        TryFromVal<Env, Self::Prototype> + IntoVal<Env, RawVal> + TryFromVal<Env, RawVal>
    {
        type Prototype: for<'a> Arbitrary<'a>;
    }
```

Basically every type relevant to soroban contracts implements (or should) this trait, including i32, u32, i64, u64, i128, u128, (), bool, I256Val, U256Val, Error, Bytes, BytesN, Vec, Map, Address, Symbol, TimepointVal, DurationVal.

The `#[contracttype]` macro automatically derives an implementation, along with a type that implements `SorobanArbitrary::Prototype`.

In use the trait looks like

```rust
use soroban_sdk::{Address, Env, Vec};
use soroban_sdk::contracttype;
use soroban_sdk::arbitrary::{Arbitrary, SorobanArbitrary};
use std::vec::Vec as RustVec;

#[derive(Arbitrary, Debug)]
struct TestInput {
    deposit_amount: i128,
    claim_address: <Address as SorobanArbitrary>::Prototype,
    time_bound: <TimeBound as SorobanArbitrary>::Prototype,
}

#[contracttype]
pub struct TimeBound {
    pub kind: TimeBoundKind,
    pub timestamp: u64,
}

#[contracttype]
pub enum TimeBoundKind {
    Before,
    After,
}

fuzz_target!(|input: TestInput| {
    let env = Env::default();
    let claim_address: Address = input.claim_address.into_val(&env);
    let time_bound: TimeBound = input.time_bound.into_val(&env).
    // fuzz the program based on the input
});
```

A more complete example is at https://github.com/brson/rs-soroban-sdk/blob/val-fuzz/soroban-sdk/fuzz/fuzz_targets/fuzz_rawval_cmp.rs

### Why

This patch assumes it is desirable to fuzz Soroban contracts with cargo-fuzz.

Soroban reference types can only be constructed with an `Env`, but the `Arbitrary` trait constructs values from nothing but bytes. The `SorobanArbitrary` trait provides a pattern to bridge this gap, expecting fuzz tests to construct `SorobanArbitrary::Prototype` types, and then convert them to their final type with `FromVal`/`IntoVal`.

There are a lot of docs here and hopefully they explain what's going on well enough.

### fuzz_catch_panic

This patch also introduces a helper function, `fuzz_catch_panic`, which is built off of the `call_with_suppressed_panic_hook` function in soroban-env-host.

The `fuzz_target!` macro overrides the Rust panic hook to abort on panic, assuming all panics are bugs, but Soroban contracts fail by panicking, and I have found it desirable to fuzz failing contracts. `fuzz_catch_panic` temporarily prevents the fuzzer from aborting on panic.

### Known limitations

The introduction of SorobanArbitrary requires a bunch of extra documentation to explain why Soroban contracts can't just use the stock Arbitrary trait.

As an alternative to this approach, we could instead expect users to construct XDR types, not SorobanArbitrary::Prototype types, and convert those to RawVals. I have been assuming that contract authors should rather concern themselves with contract types, and not the serialization types; and presently there are a number of XDR types that have values which can't be converted to contract types. The SorobanArbitrary::Prototype approach does allow more flexibility in the generation of contract types, e.g. we can generate some adversarial types like invalid object references and bad tags.

Contracts must use `IntoVal` to create the final types, but these traits are under-constrained for this purpose and always require type hints:

```rust
fuzz_target!(|input: <Address as SorobanArbitrary>::Prototype| {
    let env = Env::default();
    let address: Address = input.into_val(&env);
    // fuzz the program based on the input
});
```

This is quite unfortunate because it means the real type must be named twice.

This patch introduces a new private module `arbitrary_extra` which simply defines a bunch of new conversions like

```rust
impl TryFromVal<Env, u32> for u32 {
    type Error = ConversionError;
    fn try_from_val(_env: &Env, v: &u32) -> Result<Self, Self::Error> {
        Ok(*v)
    }
}
```

These conversions are required by `SorobanArbitrary`, which is only defined when `cfg(feature = "testutils")`; the `arbitrary_extra` module defines these conversions for all cfgs to ensure type inference is always the same, but the impls are probably useless outside of the SorobanArbitrary prototype pattern.

Crates that use `#[contracttype]` need to define a "testutils" feature if they want to use Arbitrary.

This patch doesn't generate "adversarial" values, which might include:

- RawVals with bad tags
- objects that have invalid references
- objects that are reused multiple times
- deeply nested vecs and maps
- vecs and maps that contain heterogenous element types

The arbitrary module has unit tests, and these help especially ensure the macros for custom types are correct, but the tests are not exhaustive.

### DIFF
--- a/soroban-sdk/src/arbitrary.rs
+++ b/soroban-sdk/src/arbitrary.rs
@@ -167,7 +167,8 @@
 //! });
 //! ```
 
-#![cfg(feature = "testutils")]
+#![cfg(any(test, feature = "testutils"))]
+#![cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
 
 /// A reexport of the `arbitrary` crate.
 ///
--- a/soroban-sdk/src/arbitrary.rs
+++ b/soroban-sdk/src/arbitrary.rs
@@ -1,0 +1,1232 @@
+//! Support for fuzzing Soroban contracts with [`cargo-fuzz`].
+//!
+//! This module provides a pattern for generating Soroban contract types for the
+//! purpose of fuzzing Soroban contracts. It is focused on implementing the
+//! [`Arbitrary`] trait that `cargo-fuzz` relies on to feed fuzzers with
+//! generated Rust values.
+//!
+//! [`cargo-fuzz`]: https://github.com/rust-fuzz/cargo-fuzz/
+//! [`Arbitrary`]: ::arbitrary::Arbitrary
+//!
+//! This module
+//!
+//! - defines the [`SorobanArbitrary`] trait,
+//! - defines the [`fuzz_catch_panic`] helper,
+//! - reexports the [`arbitrary`] crate and the [`Arbitrary`] type.
+//!
+//! This module is only available when the "testutils" Cargo feature is defined.
+//!
+//!
+//! ## About `cargo-fuzz` and `Arbitrary`
+//!
+//! In its basic operation `cargo-fuzz` fuzz generates raw bytes and feeds them
+//! to a program-dependent fuzzer designed to exercise a program, in our case a
+//! Soroban contract.
+//!
+//! `cargo-fuzz` programs declare their entry points with a macro:
+//!
+//! ```
+//! # macro_rules! fuzz_target {
+//! #     (|$data:ident: $dty: ty| $body:block) => { };
+//! # }
+//! fuzz_target!(|input: &[u8]| {
+//!     // feed bytes to the program
+//! });
+//! ```
+//!
+//! More sophisticated fuzzers accept not bytes but Rust types:
+//!
+//! ```
+//! # use arbitrary::Arbitrary;
+//! # macro_rules! fuzz_target {
+//! #     (|$data:ident: $dty: ty| $body:block) => { };
+//! # }
+//! #[derive(Arbitrary, Debug)]
+//! struct FuzzDeposit {
+//!     deposit_amount: i128,
+//! }
+//!
+//! fuzz_target!(|input: FuzzDeposit| {
+//!     // fuzz the program based on the input
+//! });
+//! ```
+//!
+//! Types accepted as input to `fuzz_target` must implement the `Arbitrary` trait,
+//! which transforms bytes to Rust types.
+//!
+//!
+//! ## The `SorobanArbitrary` trait
+//!
+//! Soroban types are managed by the host environment, and so must be created
+//! from an [`Env`] value; `Arbitrary` values though must be created from
+//! nothing but bytes. The [`SorobanArbitrary`] trait, implemented for all
+//! Soroban contract types, exists to bridge this gap: it defines a _prototype_
+//! pattern whereby the `fuzz_target` macro creates prototype values that the
+//! fuzz program can convert to contract values with the standard soroban
+//! conversion traits, [`FromVal`] or [`IntoVal`].
+//!
+//! [`Env`]: crate::Env
+//! [`FromVal`]: crate::FromVal
+//! [`IntoVal`]: crate::IntoVal
+//!
+//! The types of prototypes are identified by the associated type,
+//! [`SorobanArbitrary::Prototype`]:
+//!
+//! ```
+//! # use soroban_sdk::arbitrary::Arbitrary;
+//! # use soroban_sdk::{TryFromVal, IntoVal, Val, Env};
+//! pub trait SorobanArbitrary:
+//!     TryFromVal<Env, Self::Prototype> + IntoVal<Env, Val> + TryFromVal<Env, Val>
+//! {
+//!     type Prototype: for <'a> Arbitrary<'a>;
+//! }
+//! ```
+//!
+//! Types that implement `SorobanArbitrary` include:
+//!
+//! - `i32`, `u32`, `i64`, `u64`, `i128`, `u128`, `I256`, `U256`, `()`, and `bool`,
+//! - [`Error`],
+//! - [`Bytes`], [`BytesN`], [`Vec`], [`Map`],
+//! - [`Address`], [`Symbol`],
+//! - [`Val`],
+//!
+//! [`I256`]: crate::I256
+//! [`U256`]: crate::U256
+//! [`Error`]: crate::Error
+//! [`Bytes`]: crate::Bytes
+//! [`BytesN`]: crate::BytesN
+//! [`Vec`]: crate::Vec
+//! [`Map`]: crate::Map
+//! [`Address`]: crate::Address
+//! [`Symbol`]: crate::Symbol
+//! [`Val`]: crate::Val
+//!
+//! All user-defined contract types, those with the [`contracttype`] attribute,
+//! automatically derive `SorobanArbitrary`. Note that `SorobanArbitrary` is
+//! only derived when the "testutils" Cargo feature is active. This implies
+//! that, in general, to make a Soroban contract fuzzable, the contract crate
+//! must define a "testutils" Cargo feature, that feature should turn on the
+//! "soroban-sdk/testutils" feature, and the fuzz test, which is its own crate,
+//! must turn that feature on.
+//!
+//! [`contracttype`]: crate::contracttype
+//!
+//!
+//! ## Example: take a Soroban `Vec` of `Address` as fuzzer input
+//!
+//! ```
+//! # macro_rules! fuzz_target {
+//! #     (|$data:ident: $dty: ty| $body:block) => { };
+//! # }
+//! use soroban_sdk::{Address, Env, Vec};
+//! use soroban_sdk::arbitrary::SorobanArbitrary;
+//!
+//! fuzz_target!(|input: <Vec<Address> as SorobanArbitrary>::Prototype| {
+//!     let env = Env::default();
+//!     let addresses: Vec<Address> = input.into_val(&env);
+//!     // fuzz the program based on the input
+//! });
+//! ```
+//!
+//!
+//! ## Example: take a custom contract type as fuzzer input
+//!
+//! ```
+//! # macro_rules! fuzz_target {
+//! #     (|$data:ident: $dty: ty| $body:block) => { };
+//! # }
+//! use soroban_sdk::{Address, Env, Vec};
+//! use soroban_sdk::contracttype;
+//! use soroban_sdk::arbitrary::{Arbitrary, SorobanArbitrary};
+//! use std::vec::Vec as RustVec;
+//!
+//! #[derive(Arbitrary, Debug)]
+//! struct TestInput {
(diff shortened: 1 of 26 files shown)


brson:
The cargo-deny CI error is about duplicate crates. I'll investigate another time.

leighmcculloch (commented):
Wow this is awesome. I've started reviewing but haven't finished yet and will have a play with it. The integration with proptest will be helpful.

I'm also particularly interested in some of those invalid RawVal cases, especially Vec (and Map) types where the contents aren't all the same type, because that's a valid Vec, for things like tuples or argument lists.

brson:
I looked into the cargo-deny errors, and pushed two commits to resolve them, but they may not be acceptable.

The first commit just upgrades proptest, which resolves the quick-error dupe.

The second adds rand 0.8 and its dependencies to the deny.toml ban whitelist, allowing dupes of rand, rand_core, rand_chacha, and getrandom. The problem here seems to be that all revisions of ed25519-dalek are on rand 0.7, while all revisions of proptest are on rand 0.8. I added comments to deny.toml indicating that proptest is only a dev-dependency so the dupes won't cause contract bloat, but it's possible that by being on a whitelist rand dupes could sneak into contracts in the future.

I'm still poking at other solutions to the duplicated rand crate, but am not super optimistic.

brson:
I found something I need to change here: I have implemented Arbitrary for U256Val, but I should be using U256 instead.

brson:
I rebased this and made the following changes:

- Implemented SorobanArbitrary for I256/U256 instead of I256Val, U256Val
- Removed the implementations for DurationVal and TimepointVal. These are not user-facing types. I am assuming there are two types yet to be implemented that wrap these and/or convert to them.
- Added some test cases.

I also discovered that `Error` cannot be stored in user-defined types, and posted a fix for that here: https://github.com/stellar/rs-soroban-env/pull/846

brson:
It's also worth noting that `RawVal` cannot be stored in UDTs: https://github.com/stellar/rs-soroban-sdk/issues/939

brson:
It looks like this is the issue for Timepoint and Duration: https://github.com/stellar/rs-soroban-env/issues/724

Also indicates that the standard Rust String will be expected to be usable in contracts in the future, so will need implementations for Soroban Arbitrary.

It's worth noting that this SorobanArbitrary is definitely a maintenance burden, as every new type that is expected to be used in contracts needs to be given explicit compatibility; and every type that might be stored in a UDT needs an implementation, and probably tests verifying they work.

brson:
Rebased.

leighmcculloch (commented):
👏🏻 Thanks @brson! This looks great.

There's a couple minor things with the testutils features I've commented inline, and I'll push fixes for because they're trivial to fix.

I've been playing around with it with a simple contract, and run into a couple issues. What I'm going to do is push a contract into this PR with some comments in the contract with the issue.

soroban-sdk/src/arbitrary.rs
```diff
@@ -0,0 +1,1232 @@
...
+//!
+//! fuzz_target!(|input: TestInput| {
+//!     let env = Env::default();
+//!     let claim_address: Address = input.claim_address.into_val(&env);
+//!     let time_bound: TimeBound = input.time_bound.into_val(&env);
+//!     // fuzz the program based on the input
+//! });
+//! ```
+
+#![cfg(feature = "testutils")]
```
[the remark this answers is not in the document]
leighmcculloch: Fixed in 54800e3.

soroban-sdk/src/tests/proptest_scval_cmp.rs
```diff
@@ -0,0 +1,123 @@
+//! Check that Val and ScVal can be converted between each other,
+//! and that their comparison functions are equivalent.
+
+#![cfg(feature = "testutils")]
```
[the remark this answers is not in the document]
leighmcculloch: Fixed in 54800e3.

tests/fuzz/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+edition = "2021"
+publish = false
+rust-version = "1.70"
+
+[lib]
+# Adding rlib is required so that the test_fuzz crate can be imported as a
+# library into the fuzzing crate. However, having multiple crate types is a
+# problem, it'll cause LTO optimizations to be disabled for the cdylib build.
+# TODO: Figure out what to do about this.
+crate-type = ["cdylib", "rlib"]
```
leighmcculloch, line 15: This problem is back. ☝🏻 cc @graydon

tests/fuzz/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,103 @@
...
+// |
+// 3  + use soroban_sdk::Error;
+// |
+// 3  + use soroban_sdk::testutils::ed25519::Error;
+// |
+// and 4 other candidates
+// help: if you import `Error`, refer to it directly
+// |
+// 18 | #[derive(Arbitrary, Debug)]
+// |
```
leighmcculloch, line 70: I need some help with this compile error ☝🏻

leighmcculloch (commented):
I pushed an example that uses the fuzzer in 22dd445. I think it'd be helpful for us to have a complete example in the test vectors of this repo. We've tried to keep the test vectors few and wherever possible keep any tests inside the sdk crate, but given that fuzzing is an out-of-crate experience, I think a test vector is suitable.

I've hit some compile errors I'm not sure what to do with though. See inline comment.

Also note the example I pushed isn't a particularly good one, I was just trying to get it working before doing anything meaningful.

tests/fuzz/Cargo.toml
leighmcculloch: Some more details is here:
- https://github.com/rust-lang/cargo/issues/4611

tests/fuzz/fuzz/fuzz_targets/fuzz_target_1.rs
leighmcculloch: I was able to resolve all my errors by importing `arbitrary = "1.3.0"` manually. So I think the re-export we have in the soroban-sdk isn't sufficient.

leighmcculloch:
I pushed some more changes so that the sdk doesn't ever generate testutils dependent code if it isn't dependent on testutils.

leighmcculloch:
I think there's three things left we need to figure a story for:

- [ ] The arbitrary re-export wasn't sufficient for me in the `test_fuzz` example's fuzzing target code. What do we need to do to fix that?

- [ ] What do we do about the rustc / cargo multiple crate-type issue with LTO disappearing from the cdylib type? Do we recommend folks build their contract in an rlib, and re-export their contract in a cdylib that imports the rlib? Feel pretty heavy. Tell people to temporarily add rlib, but to remove it when finished fuzzing? 😕 

- [ ] Any objection to renaming the module from arbitrary to fuzz, or should we keep the name arbitrary?

graydon:
Pushed a fix for the arbitrary re-export issue. It's not beautiful, but it works. The Arbitrary macro doesn't path-qualify the code it generates, so it assumes there'll be a module `arbitrary` in its expansion environment. Yay (lack of) hygiene.

graydon:
The solution to the multiple crate-types situation [suggested by cargo devs](https://github.com/rust-lang/cargo/issues/12260#issuecomment-1589624151) is to build our wasms with `cargo rustc --crate-type cdylib`, overriding whatever's written in `Cargo.toml`. This does work (it produces the LTO build) and I think it might be enough of a workaround to just .. include in the docs somewhere, maybe set as more of a default pattern in examples (i.e. remove the `crate-type` specifications from `Cargo.toml` and change the `Makefile` that builds wasms to call `cargo rustc --crate-type cdylib` explicitly). WDYT?

leighmcculloch:
I think that's reasonable. We can make that the default in docs for how to build .wasm files, and contracts can be rlib otherwise. We can keep the builds in this repo as is for the convenience of the workspace, and update the soroban-docs to show a setup without a crate-type in the toml, and using that new command.

leighmcculloch (approved):

graydon:
@leighmcculloch as far as renaming it `fuzz` I don't think that's appropriate -- the same `arbitrary` machinery here is being used for fuzzing and proptesting (which I'm actually slightly more excited about than fuzzing, tbh!)

graydon (approved):
The only thing that seems a little odd / unmotivated to me is the specialized typed vec<foo> and map<foo,foo> types rather than just vec<val> and map<val,val> but it's not a big deal and I can imagine them plausibly being useful at some point, no need to delay further on that basis.

Overall this is great stuff! Thanks so much.

tests/fuzz/fuzz/fuzz_targets/fuzz_target_1.rs
brson: The `Arbitrary` derive expects the `arbitrary` crate/module to be in scope. The `soroban_sdk::arbitrary` module reexports the `arbitrary` crate for this purpose, so

```
use soroban_sdk::arbitrary::arbitrary;
```

should fix this.

soroban-sdk/src/lib.rs
```diff
@@ -686,4 +686,7 @@ pub mod xdr;
 
 pub mod testutils;
 
+pub mod arbitrary;
```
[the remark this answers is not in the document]
brson: sgtm

brson:
> The only thing that seems a little odd / unmotivated to me is the specialized typed vec and map<foo,foo> types rather than just vec and map<val,val> but it's not a big deal and I can imagine them plausibly being useful at some point, no need to delay further on that basis.

If this is referring to the giant list of types that might be generated for an arbitrary Val, then I agree it's not ideal. Just getting something working that covers a bunch of cases. It can be improved in the future. The current Val generator can be considered an implementation detail.

PR #960, opened 2023-05-30
Remove stellar-xdr as dep in soroban-sdk

leighmcculloch:
### What
Remove stellar-xdr as dep in soroban-sdk.

### Why
It isn't needed. The stellar-xdr is re-exported in the imported env.

The only places the SDK repo needs to import the XDR independent of an env is for contract specs. We should keep it that way, because when we need to experiment with new contract specs it's helpful to be able to update the version of the stellar-xdr lib for that purpose ahead of the version that is used by the env crate which is only possible if we're using the stellar-xdr lib for a disjoint purpose to the xdr in the sdk and from the env.

This also encourages others to do the best thing. Most of the uses are in test contracts. If folks look at the test contracts and follow their patterns, we're hopefully directing them to do things we want them to do in the real world. Using the re-exported xdr helps prevent xdr misalignment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,7 +869,6 @@ dependencies = [
  "soroban-sdk-macros",
  "soroban-spec",
  "stellar-strkey",
- "stellar-xdr",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


leighmcculloch:
Closing, as I forgot that the whole reason we use a separate dev dependency for stellar-xdr is for writing tests against the spec.

Issue #379, opened 2022-07-29
Re-evaluate Option<Result<...>> trade-offs


jonjove:
```[tasklist]
### Tasks
- [x] Storage https://github.com/stellar/rs-soroban-sdk/pull/961
- [x] Set #979
- [x] Vec https://github.com/stellar/rs-soroban-sdk/pull/974
- [x] Map #980 
- [ ] Bytes
```

leighmcculloch:
We can evaluate this from multiple fronts:

- Secure Coding – Does this meaningfully help developers write more secure code. Is the code more likely to do what the developer wants and expects.

- Ergonomics – How convenient is the code.

– Code Size – Does this trade-off too significantly on the code size generated. Is the code size noise, or significant?

– Instruction Count – Does this trade-off too significantly on the instructions executed.

I'm suspicious of the pattern as well, but I was also surprised to find that in a decent amount of the code I've updated adding it, the code already had safety checks and so the addition of the `Option<>` just moved logic around, and it didn't introduce new branches.

graydon:
I'd lean (ergonomics-wise) towards making the normal cases return `Result<>` as they have to and adding `_try` cases that return `Result<Option<T>,E>` when it's less of an "error" and more of a partial function (eg. `Map::try_get`)

leighmcculloch:
@graydon Are you suggesting lifting the Option::None into an error in the Result? This is what we had before what we had now and it felt cumbersome to maintain.

Or do you mean switch around the names of the functions, so the _unchecked function becomes the main one, and the main one becomes the try_ one?

graydon:
@leighmcculloch the latter option -- we make the main path one that panics if surprised, and we give users a try path for handling the unusual cases if they care. We got feedback recently that this is a bit of an ergonomic sore spot for users. Do you feel strongly either way?

leighmcculloch:
I think that sounds fine. If doing that, I think we should remove the _unchecked variations, which often are just confusing.

Instead of littering the type with 2x of every function, what if we have two types, `Vec` and `TryVec`. The former would just wrap the latter. `TryVec` would essentially be what we have. `Vec` would wrap it and call `.unwrap_optimized` on any `Result` value.

Would there be a performance downside to doing it that way?

We could hide `TryVec` in the docs for now, to see if anyone asks for this behavior, but to not advertise it. This would be better imo over making it private, since if someone shows up an needs it, private means there's no option until a future release to use it.

[...]

PR #961, opened 2023-05-30
Remove Result from Storage::get API

leighmcculloch:
### What
Remove Result from Storage::get API, and remove Storage::get_unchecked.

### Why
A contract is the only entity capable of writing data for a key to storage. Therefore a correctly written contract should always be able to accurate predict the type of the data written to storage.

While functions processing user data need to be able to handle unexpected data, storage functions do not because the data is predictable.

Related to https://github.com/stellar/rs-soroban-sdk/issues/379#issuecomment-1569273478

### DIFF
--- a/soroban-sdk/src/lib.rs
+++ b/soroban-sdk/src/lib.rs
@@ -395,8 +395,7 @@ pub use soroban_sdk_macros::contractmeta;
 ///     pub fn get_state(env: Env) -> State {
 ///         env.storage()
 ///             .get(&Symbol::short("STATE"))
-///             .unwrap_or_else(|| Ok(State::default())) // If no value set, assume 0.
-///             .unwrap() // Panic if the value of COUNTER is not a State.
+///             .unwrap_or_else(|| State::default()) // If no value set, assume 0.
 ///     }
 /// }
 ///
@@ -470,7 +469,6 @@ pub use soroban_sdk_macros::contractmeta;
 ///     pub fn get(env: Env) -> Option<Color> {
 ///         env.storage()
 ///             .get(&Symbol::short("COLOR"))
-///             .map(Result::unwrap) // Panic if the value of COLOR is not a Color.
 ///     }
 /// }
 ///
--- a/soroban-sdk/src/storage.rs
+++ b/soroban-sdk/src/storage.rs
@@ -3,7 +3,7 @@ use core::fmt::Debug;
 
 use crate::{
     env::internal::{self, RawVal},
-    unwrap::UnwrapInfallible,
+    unwrap::{UnwrapInfallible, UnwrapOptimized},
     Env, IntoVal, TryFromVal,
 };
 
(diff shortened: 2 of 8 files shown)


graydon:
Generally ok with this direction (it's an improvement!) but I still feel like this could maybe go a bit further: make `get` return just `T` and make `try_get()` return `Option<T>` (or `Option<Result<T,Err>>` since it's possible other errors besides type-incorrectness happen?) -- I'm just thinking that if a contract is a state machine, the simple path users will want to code often is just "fail if an invalid transition is attempted" including things like "get a nonexisting key". IOW it seems likely to me that 90% of `get(k)` calls will be `get(k).unwrap()` in which case maybe we should just make that what `get(k)` does?

leighmcculloch:
I don't think we should wrap `Result` and `Option` up together in a variant. A user will likely want `T` or `Option<T>`, but unlikely to want `Option<Result<T>>`. Wrapping them up together in a try variation will mean anyone wanting the more common `Option<T>` will be forced to also deal with the Result, which is unfortunate.

We could have three variants of functions. We had that actually at one point. But it was confusing.

In this specific case `Result` is so unimportant we can omit it all together. However, the pattern we use here I'd like to also be extendable to `Vec`. Result will be more important in `Vec`. If we have a try variant that returns only Option here, it's inconsistent to have it to also return `Result` in `Vec`, and in `Vec` I don't think we should have three variants for the same reason as above.

Returning `Option<T>` feels like a reasonable compromise that extends well to `Vec`. Some contracts I've seen are using `unwrap_or` with `Option<T>` which makes me think it's reasonable to return `Option<T>`. In contrast to `Result<T>` which we've learnt no one cares about.

The change to `Vec` would result in `get` also being `Option<T>` and a try variant returning `Option<Result<T>>`.

Wdyt?

leighmcculloch:
> I don't think i've ever used .get without .unwrap tbh unless i used .get_unchecked

@silence48 Are you advocating for us keeping `.get_unchecked` as well?

This PR was also proposing removing the function, but it sounds like it's been valuable to you?

If we keep it, the Storage interface will be:

```
pub fn get<K, V>(&self, key: &K) -> Option<V>
pub fn get_unchecked<K, V>(&self, key: &K) -> V
```

And the Vec interface would following the same pattern become:

```
pub fn get<T>(&self, i: u32) -> Option<V>
pub fn get_unchecked<T>(&self, i: u32) -> V
pub fn try_get<T>(&self, i: u32) -> Option<Result<V, _>>
```

Is having both unchecked and try versions too much? In Vec we can reduce the API surface area by breaking the try functions into a separate type.


```
impl Vec<T> {
    pub fn get<T>(&self, i: u32) -> Option<V>
    pub fn get_unchecked<T>(&self, i: u32) -> V
}

impl TryVec<T> {
    pub fn get<T>(&self, i: u32) -> Option<Result<V, _>>
    pub fn get_unchecked<T>(&self, i: u32) -> Result<V, _>
}
```

silence48:
> > I don't think i've ever used .get without .unwrap tbh unless i used .get_unchecked
> 
> @silence48 Are you advocating for us keeping `.get_unchecked` as well?
> 
> This PR was also proposing removing the function, but it sounds like it's been valuable to you?
> 
> If we keep it, the Storage interface will be:
> 
> ```rust
> pub fn get<K, V>(&self, key: &K) -> Option<V>
> pub fn get_unchecked<K, V>(&self, key: &K) -> V
> ```
> 
> And the Vec interface would following the same pattern become:
> 
> ```rust
> pub fn get<T>(&self, i: u32) -> Option<V>
> pub fn get_unchecked<T>(&self, i: u32) -> V
> pub fn try_get<T>(&self, i: u32) -> Option<Result<V, _>>
> ```
> 
> Is having both unchecked and try versions too much? In Vec we can reduce the API surface area by breaking the try functions into a separate type.
> 
> ```rust
> impl Vec<T> {
>     pub fn get<T>(&self, i: u32) -> Option<V>
>     pub fn get_unchecked<T>(&self, i: u32) -> V
> }
> 
> impl TryVec<T> {
>     pub fn get<T>(&self, i: u32) -> Option<Result<V, _>>
>     pub fn get_unchecked<T>(&self, i: u32) -> Result<V, _>
> }
> ```
Sorry for my hurried comment earlier.

I wouldn't say I'm advocating for it at all, if there is a better way to do it I think we should go that route.  I don't think there's any real valid reason we would need to use get _unchecked because basically iirc the reason i had used that was because I couldn't get an option to work in a struct.  

It seems to be that by keeping get_unchecked we aren't limiting anything it doesn't seem like, but a contract should not need it and by leaving it in, it gives developers a way to do less safe things without any type checking.  The more I think about it the less we need it, and this .get implementation looks good.

leighmcculloch:
Removing it, the interface becomes:

```rust
pub fn get<K, V>(&self, key: &K) -> Option<V>
```

And the Vec interface would following the same pattern become:

```rust
pub fn get<T>(&self, i: u32) -> Option<V>
pub fn try_get<T>(&self, i: u32) -> Option<Result<V, _>>
```

or

```rust
impl Vec<T> {
    pub fn get<T>(&self, i: u32) -> Option<V>
}

impl TryVec<T> {
    pub fn get<T>(&self, i: u32) -> Option<Result<V, _>>
}
```

silence48 (approved):
This is an improvement over the current implementation I agree.  the reason i brought up get unchecked is this will create a breaking change in some cases, but i think it's alright because the benefits outweigh the few moments it will take to change implementations.

wildework:
I think we shouldn't overload the standard interface with getters. This looks good instead.

```rust
impl TryVec<T> {
    pub fn get<T>(&self, i: u32) -> Option<Result<V, _>>
}
```

While the standard interface is left with a single `get`.

```rust
pub fn get<K, V>(&self, key: &K) -> Option<V>
```

graydon (approved):

Issue #966, opened 2023-06-06
Token SDK could use "in-place" documentation


ElliotFriend:
### What problem does your feature solve?

A developer without a background in EVM or smart contracts in general can be easily confused by the existing interface documentation for the SAC (i.e., `from`, `spender`, `allowance`, etc. may not have much context or meaning to them).

Additionally, between the `soroban-docs` site, and the SDK documentation there are different levels of information, when a "unified" set of descriptions could be useful in both places. Having this information "originate" in the source code and then copied into the documentation elsewhere seems like the ideal way to designate the SDK repo as the "canonical" source for it.

### What would you like to see?

Following a brief (old-ish) conversation in Slack, the Token interface code could use some "doc comments" to make it feel a bit more informative. We don't have to dive into the nitty-gritty and detailed descriptions of vocabulary (that's probably more fitting for the docs site, anyway). But a fuller description of the arguments each function expects could greatly improve usability.

### What alternatives are there?

Not exactly sure. Here are some guesses?

- Keep the existing documentation fragmented between docs-site, source code, etc.
- Have the interface documentation live solely in the SDK?
- Have the interface documentation live solely in the docs-site?

tomerweller:
I think that documentation in-place in the sdk would be great, but is not sufficient. We also need:
 
1. A technical token Interface SEP. We haven't really started creating Soroban SEPs but we need to start and none is more important than the token one. This should probably look like the [official erc20 doc](https://eips.ethereum.org/EIPS/eip-20) adapted to Soroban and the SEP format. 
2. A more user friendly doc that explains all the token interface concepts. something like [this](https://ethereum.org/en/developers/docs/standards/tokens/erc-20/)

ElliotFriend:
> documentation in-place in the sdk would be great, but is not sufficient

As an initial effort, I copied some of what lives in the [Token Interface] page in the docs-site, with expanded definitions and explanations of the arguments and stuff. The PR is here: #967 

> 1. A technical token Interface SEP. We haven't really started creating Soroban SEPs but we need to start and none is more important than the token one. This should probably look like the [official erc20 doc](https://eips.ethereum.org/EIPS/eip-20) adapted to Soroban and the SEP format.

[...]

PR #967, opened 2023-06-06
Token doc comments

ElliotFriend:
### What

Re: #966

Includes brief "doc comments" in the token interface to assist with creating a "canonical" set of descriptions and interface documentation that can be copied/applied in the docs-site, etc.

### Why

We have an interface description on the docs-site: https://soroban.stellar.org/docs/reference/interfaces/token-interface. This is a great step toward seeing this interface well understood by devs, but leaves a lot to be desired (terms are a bit unclear, and assumptions are made about the reader's understanding of them).

There currently is no documentation embedded into the token interface's source code. Having these doc-comments can help with IDE code completion, and educate devs who don't have a background in EVM or smart contract development. Additionally, this repo seems like the ideal place for this to "originate" and be disseminated/adapted to other contexts.

### Known limitations

I tried following the commands in `CONTRIBUTING.md` following these changes. `make fmt`, `make doc`, and `make build` all complete successfully, and I can preview the documentation changes I've made locally. However, `make test` fails at `test::token_spec::test_spec_xdr` and `test::token_spec::test_spec_xdr`. Here's the relevant output:

```bash
thread 'tests::token_spec::test_spec_xdr' panicked at 'index out of bounds: the len is 1108 but the index is 1108', soroban-sdk/src/token.rs:270:17
thread 'tests::token_spec::test_spec_xdr_len' panicked at 'assertion failed: `(left == right)`
  left: `1108`,
 right: `5900`', soroban-sdk/src/tests/token_spec.rs:13:5
```

I feel a bit out-of-my-depth with that error, though. I'm not sure I know enough about the testing setup to proceed with troubleshooting. Does anybody have a suggestion about what might be going wrong?

_Edit:_ I figured out the test panic, and "fixed it," but I'm not sure I fixed it the right way lol


ElliotFriend:
I figured out at least _how_ to get the tests passing. fac01a2 changes the `SPEC_XDR_LEN` value to 5900 (because that's what the error message told me). I'm far from knowing precisely what that value does, or _why_ that problem presented itself. Moreover, I don't know if pulling a number from the error message is the right way to determine the value, or if there's a better, more "proper" way. at least the tests passed, though... ¯\_(ツ)_/¯

leighmcculloch (approved):

leighmcculloch:
@ElliotFriend Thanks!

> I figured out at least _how_ to get the tests passing. [fac01a2](https://github.com/stellar/rs-soroban-sdk/commit/fac01a2544cb097c17153139de0961fd82ec6a01) changes the `SPEC_XDR_LEN` value to 5900 (because that's what the error message told me). I'm far from knowing precisely what that value does, or _why_ that problem presented itself.

The size of the array has to be known at compile time due to its use in a const function, and having to be a fixed size array. The compiler can't determine it itself, but we have those tests so that if we set the wrong value we can find out what the right value is easily.

PR #974, opened 2023-06-14
Make Result<> optional in Vec fns

### LINKED ISSUE #379: Re-evaluate Option<Result<...>> trade-offs

leighmcculloch:
### What
Add try_ variations of each function in the Vec type that continue to have the Result<> and remove the Result<> from existing functions. Also does same for the iterator.

#### `get`

Instead of `get` returning `Option<Result<T, _>>`, it now returns `Option<T>`.

There also exists `get_unchecked` that returns `T`.

There also exists `try_get` that returns `Result<Option<T>, _>`.

There also exists `try_get_unchecked` that returns `Result<T, _>`.

The first two functions, `get` and `get_unchecked`, are consistent with `std::vec::Vec`'s functions of the same name, so they will be easily found, and easily recognized by Rust developers.

The second two functions, `try_` variations, are used internally by the SDK in several places, and will likely only be used by advanced users, or where we explicitly say you should use it to protect against a certain vulnerability. The docs have been updated to provide some recommendations on those places.

#### `iter`

Instead of `iter` returning an iterator that returns `Option<Result<T, _>`, it now returns `Option<T>`.

There also exists `try_iter` that returns `Option<Result<T, _>>`.

### Why
When developers are accessing Vec's with trusted data they created themselves or have prior validated, they don't need to keep handling the case that the Vec might contain elements of the wrong type.

We've seen plenty of evidence that developers are calling .unwrap() on the result, so we should take it out of the main path developers are following.

The `Result` and `Option` were swapped in the `get` case because it is common for developers to handle errors first. This was feedback provided by @graydon and @willemneal on the existing API that exposes the `Option` ahead of the error.

The `Result` and `Option` were not swapped in the `iter` case because the iterator interface requires an `Option` first because it signals whether there are anymore items to attempt to decode.

Close #379

### Known Limitations

The surface area of the Vec API is expanding, although most of it is just variations on-top of one concept. Some functions have four variations now.

This change does not address inefficiencies within the Vec type that we separately plan to improve, such as the repeated calls to `.len()` internally, and the iterator using the `.slice()` function. These are both things we will address separately.


soroban-sdk/src/vec.rs
```diff
@@ -380,21 +380,26 @@ where
...
+            Ok(None)
         }
     }
 
     #[inline(always)]
-    pub fn get_unchecked(&self, i: u32) -> Result<T, T::Error>
-    where
-        T::Error: Debug,
-    {
+    pub fn get_unchecked(&self, i: u32) -> T {
```
leighmcculloch, line 397: I've kept the _unchecked variations. I think it's dubious to have four variations of a function, but the _unchecked variation seems pretty helpful when you know you're within bounds.

graydon (approved):

PR #976, opened 2023-06-14
fix(ts): properly encode u/i128 numbers

willemneal:
### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/soroban-spec/fixtures/ts/src/convert.ts
+++ b/soroban-spec/fixtures/ts/src/convert.ts
@@ -133,14 +133,14 @@ export function addressToScVal(addr: string): xdr.ScVal {
 
 export function i128ToScVal(i: bigint): xdr.ScVal {
     return xdr.ScVal.scvI128(new xdr.Int128Parts({
-        lo: xdr.Uint64.fromString((Number(i & BigInt(0xFFFFFFFFFFFFFFFFn))).toString()),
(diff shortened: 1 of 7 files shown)


willemneal:
@leighmcculloch Can you make a new release branch for this PR to target.

leighmcculloch:
I've updated this pull to be based on #977.

soroban-spec/fixtures/ts/src/convert.ts
```diff
@@ -130,3 +130,17 @@ export function addressToScVal(addr: string): xdr.ScVal {
     let addrObj = Address.fromString(addr);
     return addrObj.toScVal();
 }
+
+export function i128ToScVal(i: bigint): xdr.ScVal {
+    return xdr.ScVal.scvI128(new xdr.Int128Parts({
+        lo: xdr.Uint64.fromString((Number(i & BigInt(0xFFFFFFFFFFFFFFFFn))).toString()),
```
leighmcculloch, line 136: Can we use `Number` for a 64-bit value? My understanding was no, because Number is max 53-bit. cc @paulbellamy @Shaptic

willemneal: Good point, I'll change it.

Shaptic: No you cannot, but pls keep an eye out for when https://github.com/stellar/js-stellar-base/pull/620 lands! (Soon:tm:)

PR #978, opened 2023-06-14
fix(ts-gen): use strict type-checking

chadoh:
### What

- the dapp-example uses strict type-checking
- it's probably a good idea to use it in these libs, too
- the `convert.ts` logic currently fails in strict mode; this fixes it

### Why

- make it easier to copy the `convert.ts` logic into other apps
- if other apps use strict type checking and also type-check the
  generated libraries (on purpose or by accident), it jush Just Work

### Known limitations

N/A

### DIFF
--- a/soroban-spec/src/gen/typescript/project_template/tsconfig.json
+++ b/soroban-spec/src/gen/typescript/project_template/tsconfig.json
@@ -69,7 +69,7 @@
     // "preserveSymlinks": true,                         /* Disable resolving symlinks to their realpath. This correlates to the same flag in node. */
     // "forceConsistentCasingInFileNames": true, /* Ensure that casing is correct in imports. */
     /* Type Checking */
-    // "strict": true, /* Enable all strict type-checking options. */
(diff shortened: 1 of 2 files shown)


chadoh:
Nevermind; this will take more work and we should wait until the TS lives in a separate repo.

soroban-spec/src/gen/typescript/project_template/tsconfig.json
```diff
@@ -95,4 +95,4 @@
   "include": [
     "src/*"
   ]
-}
\ No newline at end of file
+}
```
leighmcculloch, line 98: @chadoh @willemneal I've noticed you folks are flipping file endings. I think one of you, maybe @willemneal, might have new-lines at end of files turned off. Would be great to fix that in your editor to avoid permadiffs.

chadoh: Thanks for the note; will check

PR #981, opened 2023-06-15
fix(TS): Prevents error when SignAndSend is true and there is no auth array

willemneal:
### What

Currently if a method is to be signed the code assumed that there was an auth array with one element

### Why

### Known limitations

[TODO or N/A]


willemneal:
@leighmcculloch Sorry should have mentioned it!

leighmcculloch (approved):

PR #982, opened 2023-06-15
fix(ts): fail on ts build errors

chadoh:
### What

Report `tsc` errors when building the JS client.

### Why

Previously, if the `tsc` step failed, the `build` script continued and failed with a confusing error further down.

### Known limitations

N/A


chadoh:
@leighmcculloch this needs to target a `release/0.8.7` branch.

leighmcculloch:
Done. Can we test as much end to end before cutting another release? Let's try to nut out the errors before hand, if possible.

PR #984, opened 2023-06-15
SDK updates for recent host changes.

dmkozh:
### What

- Support deep subcontract authorization
- Update deployer interface to work on behalf of any address.rs
- Use host fn for uploading Wasm instead of a test-only fn

### Why

Supporting new host functionality

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/auth.rs
+++ b/soroban-sdk/src/auth.rs
@@ -1,15 +1,64 @@
 use crate::{contracttype, Address, BytesN, RawVal, Symbol, Vec};
 
-#[contracttype(crate_path = "crate", export = false)]
+/// Context of a single authorized call peformed by an address.
+///
+/// Custom account contracts that implement `__check_auth` special function
+/// receive a list of `Context` values corresponding to all the calls that
+/// need to be authorized.
 #[derive(Clone)]
-pub struct Context {
-    pub contract: BytesN<32>,
+#[contracttype(crate_path = "crate", export = false)]
+pub enum Context {
+    Contract(ContractContext),
+    CreateContractHostFn(CreateContractHostFnContext),
+}
+
+/// Authorization context of a single contract call.
+///
+/// This struct corresponds to a `require_auth_for_args` call for an address
+/// from `contract` function with `fn_name` name and `args` arguments.
+#[derive(Clone)]
+#[contracttype(crate_path = "crate", export = false)]
+pub struct ContractContext {
+    pub contract: Address,
     pub fn_name: Symbol,
     pub args: Vec<RawVal>,
 }
 
-impl Context {
-    pub fn contract(&self) -> Address {
-        Address::from_contract_id(&self.contract)
-    }
+/// Authorization context for `create_contract` host function that creates a
+/// new contract on behalf of authorizer address.
+#[derive(Clone)]
+#[contracttype(crate_path = "crate", export = false)]
+pub struct CreateContractHostFnContext {
+    pub executable: ContractExecutable,
+    pub salt: BytesN<32>,
+}
+
+/// Contract executable used for creating a new contract and used in
+/// `CreateContractHostFnContext`.
+#[derive(Clone)]
+#[contracttype(crate_path = "crate", export = false)]
+pub enum ContractExecutable {
+    Wasm(BytesN<32>),
+}
+
+/// A node in the tree of authorizations performed on behalf of the current
+/// contract as invoker of the contracts deeper in the call stack.
+///
+/// This is used as an argument of `authorize_as_curr_contract` host function.
+///
+/// This tree corresponds `require_auth[_for_args]` calls on behalf of the
+/// current contract.
+#[derive(Clone)]
+#[contracttype(crate_path = "crate", export = false)]
+pub enum InvokerContractAuthEntry {
+    Contract(SubContractInvocation),
+    CreateContractHostFn(CreateContractHostFnContext),
+}
+
+/// Value of contract node in InvokerContractAuthEntry tree.
+#[derive(Clone)]
+#[contracttype(crate_path = "crate", export = false)]
+pub struct SubContractInvocation {
+    pub context: ContractContext,
+    pub sub_invocations: Vec<InvokerContractAuthEntry>,
 }
--- a/soroban-sdk/src/deploy.rs
+++ b/soroban-sdk/src/deploy.rs
@@ -20,8 +20,8 @@
 //! # impl Contract {
 //! #     pub fn f(env: Env, wasm_hash: BytesN<32>) {
 //! #         let salt = [0u8; 32];
-//! let deployer = env.deployer().with_current_contract(&salt);
-//! let contract_address = deployer.deploy(&wasm_hash);
+//! let deployer = env.deployer().with_current_contract();
+//! let contract_address = deployer.deploy(&salt, &wasm_hash);
 //! #     }
 //! # }
 //! #
@@ -30,14 +30,17 @@
 //! #     let env = Env::default();
 //! #     let contract_address = env.register_contract(None, Contract);
 //! #     // Install the contract code before deploying its instance.
-//! #     let wasm_hash = env.install_contract_wasm(&[0u8; 100]);
+//! #     let mock_wasm = [0u8; 100];
+//! #     let wasm_hash = env.deployer().upload_contract_wasm(mock_wasm.as_slice());
 //! #     ContractClient::new(&env, &contract_address).f(&wasm_hash);
 //! # }
 //! # #[cfg(not(feature = "testutils"))]
 //! # fn main() { }
 //! ```
 
-use crate::{env::internal::Env as _, unwrap::UnwrapInfallible, Address, BytesN, Env, IntoVal};
+use crate::{
+    env::internal::Env as _, unwrap::UnwrapInfallible, Address, Bytes, BytesN, Env, IntoVal,
+};
 
 /// Deployer provides access to deploying contracts.
 pub struct Deployer {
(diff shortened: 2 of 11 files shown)


soroban-sdk/src/deploy.rs
```diff
@@ -54,85 +57,76 @@ impl Deployer {
...
-    salt: BytesN<32>,
+    address: Address,
 }
 
-impl DeployerWithCurrentContract {
-    /// Return the ID of the contract defined by the deployer.
-    #[doc(hidden)]
-    pub fn id(&self) -> Address {
-        todo!()
-    }
```
leighmcculloch, line 98: I don't think we need to remove this function, we're just about to fill this method in because @jayz22 just finished the environment changes to support it. The intent for why salt lived in the type represents a deployer which deploys to a specific address, not a variety of addresses. Could we keep salt there and forego the refactor?

soroban-sdk/src/env.rs
```diff
@@ -419,6 +420,26 @@ impl Env {
...
+    ///
+    /// For example, if the contract A calls contract B, contract
+    /// B calls contract C and contract C calls `A.require_auth()`, then an
+    /// entry corresponding to C call has to be passed in `auth_entries`. It
+    /// doesn't matter if contract B called `require_auth` or not. If contract A
+    /// calls contract B again, then `authorize_as_curr_contract` has to be
+    /// called again with the respective entries.
+    ///
+    ///
+    pub fn authorize_as_curr_contract(&self, auth_entries: Vec<InvokerContractAuthEntry>) {
```
leighmcculloch, line 438: Let's avoid the term `curr`.

I think we can simply name this `authorize`, because the only time a contract would authorize anything is when the contract is the one doing the authorization?

```suggestion
    pub fn authorize(&self, auth_entries: Vec<InvokerContractAuthEntry>) {
```

Alternatively, if we would like to be more specific, I think we can say:

```suggestion
    pub fn authorize_as_self(&self, auth_entries: Vec<InvokerContractAuthEntry>) {
```

soroban-sdk/src/deploy.rs
```diff
@@ -54,85 +57,76 @@ impl Deployer {
...
-    ) -> DeployerWithCurrentContract {
-        let env = self.env();
-        DeployerWithCurrentContract {
-            env: env.clone(),
-            salt: salt.into_val(env),
+    /// from the current contract.
+    pub fn with_current_contract(&self) -> DeployerWithAddress {
+        DeployerWithAddress {
+            env: self.env.clone(),
+            address: self.env.current_contract_address(),
```
leighmcculloch, line 64: Nice, I like how we're just using the same type 👏🏻 .

soroban-sdk/src/deploy.rs
```diff
@@ -54,85 +57,76 @@ impl Deployer {
...
+            address,
         }
     }
+
+    /// Deploys a new instance of Stellar Asset Contract corresponding to
+    /// provided serialized asset.
+    ///
+    /// `serialized_asset` is the Stellar `Asset` XDR serialized to bytes. Refer
+    /// to `stellar-xdr` create for the exact definition.
+    pub fn deploy_stellar_asset(&self, serialized_asset: impl IntoVal<Env, Bytes>) -> Address {
```
leighmcculloch, line 84: This function should be inside a Deployer, and have a corresponding `id` method that can give you the ID of the contract that would be created.

soroban-sdk/src/env.rs
```diff
@@ -428,8 +449,7 @@ impl Env {
     /// Replaces the executable of the current contract with the provided Wasm.
     ///
     /// The Wasm blob identified by the `wasm_hash` has to be already present
-    /// on-chain (the upload happens via `INSTALL_CONTRACT_CODE` host function
-    /// or via `install_contract_wasm` test function in unit tests).
+    /// on-chain (uploaded via `upload_contract_wasm` host function).
```
leighmcculloch, line 452: The developer doesn't interact with host functions, they interact with SDK functions. We should refer to them instead.
```suggestion
    /// on-chain (uploaded via `[Deployer::upload_contract_wasm]` host function).
```

soroban-sdk/src/deploy.rs
```diff
@@ -54,85 +57,76 @@ impl Deployer {
...
+        self.env
+            .create_asset_contract(serialized_asset.into_val(&self.env).to_object())
+            .unwrap_infallible()
+            .into_val(&self.env)
+    }
+
+    /// Upload the contract Wasm code to the network.
+    ///
+    /// Returns the hash of the uploaded Wasm that can be then used for
+    /// the contract deployment.
```
leighmcculloch, line 94: Can you retain the example we used to have on the `install_contract_wasm` function?

soroban-sdk/src/tests/contractimport.rs
```diff
@@ -69,7 +69,7 @@ fn test_reregister_wasm() {
 
     // Register a contract with code that will fail, to ensure this code isn't
     // the code that gets activated when invoked.
-    let add_contract_id = e.register_contract_wasm(None, &[]);
+    let add_contract_id = e.register_contract_wasm(None, []);
```
leighmcculloch, line 72: Nice.

soroban-sdk/src/deploy.rs
dmkozh: Done, though in general doctests here are weird (only the module as a whole is covered and none of the `with_` functions were). We could improve coverage as a followup though.

leighmcculloch: I'd love to get doc tests onto most functions. It's a bit scattered, some types have them, some functions. It may make sense for this example to be moved into the deploy modules doctests examples instead. Happy to figure that out later though, you don't have to address it here, would be good just to keep it for moving later.

soroban-sdk/src/auth.rs
```diff
@@ -1,15 +1,64 @@
 use crate::{contracttype, Address, BytesN, RawVal, Symbol, Vec};
 
+/// Context of a single authorized call peformed by an address.
+///
+/// Custom account contracts that implement `__check_auth` special function
+/// receive a list of `Context` values corresponding to all the calls that
+/// need to be authorized.
+#[derive(Clone)]
 #[contracttype(crate_path = "crate", export = false)]
+pub enum Context {
```
leighmcculloch, line 10: I thought I saw a comment about it being a bit strange we are using contracttype for built-in stuff. I actually think this is good. It forces us to ensure the contracttype is efficient, and reduces the number of bespoke host functions we need.

soroban-sdk/src/deploy.rs
```diff
@@ -53,61 +56,102 @@ impl Deployer {
...
+            address,
+            salt: salt.into_val(&self.env),
+        }
+    }
+
+    /// Get a deployer that deploys an instance of Stellar Asset Contract
+    /// corresponding to the provided serialized asset.
+    ///
+    /// `serialized_asset` is the Stellar `Asset` XDR serialized to bytes. Refer
+    /// to `stellar-xdr` create for the exact definition.
```
leighmcculloch, line 92: The type is in this crate, we can refer to it here.
```suggestion
    /// `serialized_asset` is the Stellar `Asset` XDR serialized to bytes. Refer
    /// to `[soroban_sdk::xdr::Asset]`
```

soroban-sdk/src/deploy.rs
```diff
@@ -53,61 +56,102 @@ impl Deployer {
...
+    ///
+    /// #[test]
+    /// fn test() {
+    /// # }
+    /// # fn main() {
+    ///     let env = Env::default();
+    ///     env.deployer().upload_contract_wasm(WASM);
+    /// }
+    /// ```
+    pub fn upload_contract_wasm(&self, contract_wasm: impl IntoVal<Env, Bytes>) -> BytesN<32> {
```
leighmcculloch, line 121: @dmkozh I thought for some reason you had advocated previously for upload being outside the Deployer type, because it isn't technically "deployment" in the contract instance case.

It may be a little odd that `update_current_contract_wasm` lives on `Env`, and yet `upload` lives in Deployer.

Should we move update here too?

soroban-sdk/src/deploy.rs
```diff
@@ -53,61 +56,102 @@ impl Deployer {
...
     salt: BytesN<32>,
 }
 
-impl DeployerWithCurrentContract {
-    /// Return the ID of the contract defined by the deployer.
+impl DeployerWithAddress {
+    /// Return the address of the contract defined by the deployer.
     #[doc(hidden)]
-    pub fn id(&self) -> Address {
+    pub fn contract_address(&self) -> Address {
```
leighmcculloch, line 140: Probably sufficient? +1 for Address over ID.
```suggestion
    pub fn address(&self) -> Address {
```
I guess both contract_address and address are pretty ambiguous given we have the address of the deployer 😕.

Maybe:


```suggestion
    /// Returns what the address of the contract will be once deployed.
    pub fn deployed_address(&self) -> Address {
```

soroban-sdk/src/auth.rs
dmkozh: Yeah, I can see both pros and cons to this, but ultimately it makes things simpler and allows for adding more complex host fns or more complex built-in contracts.

soroban-sdk/src/deploy.rs
dmkozh: I don't really have any strong opinion on that. Previously I didn't want to allow uploading wasm from contracts, but then I figured that makes sense for uploading external wasm (and for the sake of feature parity with operation). For the update I guess I just wanted to get something working quickly. Move to deployer for now.

leighmcculloch (approved):

PR #991, opened 2023-06-16
Fix a few clippy warnings

leighmcculloch:
### What
Fix a few warnings from clippy about redundant code.

### Why
So that we write necessary code. Code with extra borrows, etc can be misleading to a reader.

### DIFF
--- a/soroban-ledger-snapshot/src/lib.rs
+++ b/soroban-ledger-snapshot/src/lib.rs
@@ -70,7 +70,7 @@ impl LedgerSnapshot {
             protocol_version: self.protocol_version,
             sequence_number: self.sequence_number,
             timestamp: self.timestamp,
-            network_id: self.network_id.clone(),
+            network_id: self.network_id,
             base_reserve: self.base_reserve,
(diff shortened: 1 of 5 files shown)


soroban-sdk/src/bytes.rs
```diff
@@ -1384,7 +1384,7 @@ mod test {
         let env = Env::default();
         let mut bin = bytes![&env, [1, 2, 3, 4]];
 
-        assert_eq!(bin.remove_unchecked(2), ());
+        bin.remove_unchecked(2);
```
leighmcculloch, line 1387: This assertion is always true.

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -70,7 +70,7 @@ impl LedgerSnapshot {
             protocol_version: self.protocol_version,
             sequence_number: self.sequence_number,
             timestamp: self.timestamp,
-            network_id: self.network_id.clone(),
+            network_id: self.network_id,
```
leighmcculloch, line 73: Network ID can be copied, it's an array.

sisuresh (approved):

PR #993, opened 2023-06-17
Adapt to RawVal -> Val renaming

graydon:
Just adapting the SDK to https://github.com/stellar/rs-soroban-env/pull/862

(Also re-resolving Cargo.lock since we have to for k256 and it's not a bad time to give it a general bump)


leighmcculloch (approved):

leighmcculloch:
Looks like it is complaining about MPL-2.0.

PR #998, opened 2023-06-21
Add _many fns for collections.

### LINKED ISSUE #997: Add get_many and similar bulk operations to collections
In response to the discussion at:
- https://github.com/stellar/rs-soroban-env/issues/696

We should make it easier for folks using the SDK to perform bulk read / write operations on the collection types, `Vec` and `Map`. Bulk complete read/writes are already possible on the `Bytes` type, but we should probably align all collections to be similar and add partial bulk read/writes there too.

The idea is to follow the existing pattern in the Rust stdlib for bulk reads, by providing functions like `get_many` that return an array containing a fixed number of items.

leighmcculloch:
### What
Add _many fns for collections, such as `get_many_unchecked` to parallel `get_unchecked`.

### Why
Close #997

### Status

This change is a preliminary example of how we might implement #997.

### DIFF
--- a/soroban-sdk/src/vec.rs
+++ b/soroban-sdk/src/vec.rs
@@ -423,6 +423,24 @@ where
         T::try_from_val(env, &val)
     }
 
+    /// Returns the items in the range starting at the position.
+    ///
+    /// ### Panics
+    ///
+    /// If the position and length is out-of-bounds.
+    ///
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Closing to continue discussion for the right interface on #997.

PR #999, opened 2023-06-21
Change Vec iter to not create many Vecs

### LINKED ISSUE #189: Type: Vec: Change iterators to not create new collection during iteration
Depending on the outcome of this conversation: https://discord.com/channels/897514728459468821/935999748840751125/994771043602407454

Related to https://github.com/stellar/rs-stellar-contract-sdk/issues/191 https://github.com/stellar/rs-stellar-contract-sdk/issues/189

leighmcculloch:
### What
Change Vec iter to not create a new Vec on each step of the iteration.

### Why

The Vec iterator currently makes a host copy of the vec on every step of the iteration. This is because it uses the slice function to get a new vec with the first or last item removed. Even though it results in optimized guest code because the iterator is only a single u64 handle, those optimizations are outsized by the calls to the host that repeatedly duplicate the vec.

Close #189

### Alternatives

There might be some advantage in having the slice function not copy the vec on the host side. The host could lazily reuse the vec backing array for read, pop, and slice operations. I don't think the guest savings would be worth the host complexity though. Additionally every step of an iteration would get two host fn calls instead of one, so in the end it is difficult to know if it would actually be better.

#### Code

```rust
let env = Env::default();

let vec = vec![&env, 0, 1, 2, 3, 4];

let mut iter = vec.iter();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();
_ = iter.next();

env.budget().print()
```

#### Code Size

```diff
@@ -1 +1 @@
--rwxr-xr-x  1 leighmcculloch  staff  1210 Jun 20 22:25 test_iter.wasm
+-rwxr-xr-x  1 leighmcculloch  staff  934 Jun 20 22:30 test_iter.wasm
```

#### Budget

```diff
@@ -1,15 +1,15 @@
 =======================================================
-Cpu limit: 40000000; used: 35996
-Mem limit: 52428800; used: 504
+Cpu limit: 40000000; used: 6835
+Mem limit: 52428800; used: 104
 =======================================================
 CostType                 cpu_insns      mem_bytes      
 WasmInsnExec             0              0              
 WasmMemAlloc             0              0              
-HostMemAlloc             28200          504            
-HostMemCpy               276            0              
+HostMemAlloc             4700           104            
+HostMemCpy               46             0              
 HostMemCmp               0              0              
 InvokeHostFunction       0              0              
-VisitObject              475            0              
+VisitObject              114            0              
 ValXdrConv               0              0              
 ValSer                   0              0              
 ValDeser                 0              0              
@@ -23,7 +23,7 @@ VmMemRead                0              0
 VmMemWrite               0              0              
 VmInstantiation          0              0              
 InvokeVmFunction         0              0              
-ChargeBudget             7020           0              
+ChargeBudget             1950           0              
 ComputeKeccak256Hash     0              0              
 ComputeEcdsaSecp256k1Key 0              0              
 ComputeEcdsaSecp256k1Sig 0              0              
```

### DIFF
--- a/soroban-sdk/src/vec.rs
+++ b/soroban-sdk/src/vec.rs
@@ -895,15 +895,15 @@ where
 #[derive(Clone)]
 pub struct VecTryIter<T> {
     vec: Vec<T>,
-    start: u32,
-    len: u32,
+    start: u32, // inclusive
+    end: u32,   // exclusive
 }
 
 impl<T> VecTryIter<T> {
     fn new(vec: Vec<T>) -> Self {
         Self {
             start: 0,
-            len: vec.len(),
+            end: vec.len(),
             vec,
         }
     }
@@ -920,17 +920,17 @@ where
     type Item = Result<T, T::Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.len == 0 {
-            None
-        } else {
+        if self.start < self.end {
             let val = self.vec.try_get_unchecked(self.start);
             self.start += 1;
             Some(val)
+        } else {
+            None
         }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let len = self.len as usize;
+        let len = (self.end - self.start) as usize;
         (len, Some(len))
     }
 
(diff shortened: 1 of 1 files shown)


graydon (approved):
Bravo!

PR #1005, opened 2023-06-22
Rename logger to logs

leighmcculloch:
### What
Rename env.logger() to env.logs().

### Why
The logger function is inconsistently named compared to the other functions for accessing resources. E.g. `events`, `storage`, `ledger`. I keep looking for logs and having to remember it is logger. The type is used not just for the act of logging, but also the act accessing logs in tests.

Note that the deploy functionality is under `deployer`, but it's not one that as easily renamed to `deploy` or `deploys`.

The rename is done backwards compatible with a deprecated notice, so folks will get a meaningful warning message and do not need to go hunting for the new function.

### DIFF
--- a/soroban-sdk/src/logs.rs
+++ b/soroban-sdk/src/logs.rs
@@ -62,8 +62,8 @@ use crate::{env::internal::EnvBase, Env, Val};
 /// let value = 5;
 /// log!(&env, "a log entry", value, Symbol::short("another"));
 ///
-/// use soroban_sdk::testutils::Logger;
-/// let logentry = env.logger().all().last().unwrap().clone();
+/// use soroban_sdk::testutils::Logs;
+/// let logentry = env.logs().all().last().unwrap().clone();
 /// assert!(logentry.contains("[\"a log entry\", 5, another]"));
 /// # }
 /// ```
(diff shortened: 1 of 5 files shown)


soroban-sdk/src/logs.rs
```diff
@@ -62,21 +62,21 @@ use crate::{env::internal::EnvBase, Env, Val};
...
     ($env:expr, $fmt:literal $(,)?) => {
         if cfg!(debug_assertions) {
-            $env.logger().log($fmt, &[]);
+            $env.logs().log($fmt, &[]);
         }
     };
     ($env:expr, $fmt:literal, $($args:expr),* $(,)?) => {
         if cfg!(debug_assertions) {
-            $env.logger().log($fmt, &[
+            $env.logs().log($fmt, &[
```
dmkozh, line 79: I'm not a huge fan of `logs().log` honestly; either `logger().log()` or `logs().add()` looks more natural to me (i.e. either 'actor.action' or 'data_structure.mutator`). Though I suppose this is rarely used directly, so I don't fee too strongly. WDYT?

leighmcculloch: Maybe `logs().write()`? Otherwise .add sgtm.

leighmcculloch: Actually, add is better, as the interface makes it more like it's a collection.

dmkozh (approved):

Issue #965, opened 2023-06-02
Split up Token Interface


leighmcculloch:
I'd like to propose that we split up the token interface into a set of small interfaces that tokens can choose to implement.

```rust
pub trait Interface {
    fn name(env: Env) -> Bytes;
    fn symbol(env: Env) -> Bytes;
    fn decimals(env: Env) -> u32;
    fn balance(env: Env, id: Address) -> i128;
    fn spendable_balance(env: Env, id: Address) -> i128;
    fn transfer(env: Env, from: Address, to: Address, amount: i128);
    fn transfer_from(env: Env, spender: Address, from: Address, to: Address, amount: i128);
    fn authorized(env: Env, id: Address) -> bool;
    fn allowance(env: Env, from: Address, spender: Address) -> i128;
    fn increase_allowance(env: Env, from: Address, spender: Address, amount: i128);
    fn decrease_allowance(env: Env, from: Address, spender: Address, amount: i128);
    fn burn(env: Env, from: Address, amount: i128);
    fn burn_from(env: Env, spender: Address, from: Address, amount: i128);
}
```

```rust
pub trait AdminInterface {
    fn admin(env: Env) -> Option<Address>;
    fn set_admin(env: Env, new_admin: Address);
    fn mint(env: Env, to: Address, amount: i128);
    fn clawback(env: Env, from: Address, amount: i128);
    fn set_authorized(env: Env, id: Address, authorize: bool);
}
```

This issue was first posted and discussed on Discord at:
- https://discord.com/channels/897514728459468821/1114325793926037605

leighmcculloch:
We will have a challenge implementing this in the SDK, because the SDK currently doesn't support very well a contract implementing multiple traits. We've really gotta deal with that, but if we do, the above would be pretty smooth.

There's an issue open for dealing with it:
- #268 

The path I suggest to implementing the above would be:


```[tasklist]
### Tasks
- [ ] Break up the interface per above. (So we can start referring to these separately, and start SEPifying them.)
- [ ] Join them back together into a super interface temporarily (to retain existing behavior)
- [ ] Get out the 💪🏻 and fix the multi-trait support: #268
- [ ] Remove the super interface. 
```

PR #1006, opened 2023-06-23
Add #[contract] macro

### LINKED ISSUE #268: Make contractimpl and derive_contract_function_set work with multiple traits

leighmcculloch:
### What
Add `#[contract]` macro that generates a contract client for the contract, and make #[contractimpl] only add functions to the existing contract client.

### Why
To support multiple `#[contractimpl]` per contract.

For contracts that have multiple impls, it is not ideal if there are multiple clients for that one contract. Ideally clients appear in the same way everywhere, and anyone importing the contract will see a single flat client.

Since macros cannot coordinate, there's no way for the first contractimpl to output the client struct and others to skip it. Today what happens is a compile error is shown because there are multiple clients generated.

Unblocks:
- #952 
- #965 
- And general better interoperability through trait sharing.

Close #268

### DIFF
--- a/soroban-sdk/src/lib.rs
+++ b/soroban-sdk/src/lib.rs
@@ -126,7 +126,7 @@ pub use bytes_lit::bytesmin as __bytes_lit_bytesmin;
 /// Defining an error and capturing errors using the `try_` variant.
 ///
 /// ```
-/// use soroban_sdk::{contracterror, contractimpl, Env};
+/// use soroban_sdk::{contract, contracterror, contractimpl, Env};
 ///
 /// #[contracterror]
 /// #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
@@ -136,6 +136,7 @@ pub use bytes_lit::bytesmin as __bytes_lit_bytesmin;
 ///     AnotherError = 2,
 /// }
 ///
+/// #[contract]
 /// pub struct Contract;
 ///
 /// #[contractimpl]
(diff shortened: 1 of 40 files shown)


graydon (approved):

soroban-sdk/src/lib.rs
```diff
@@ -259,6 +262,52 @@ pub use soroban_sdk_macros::contracterror;
...
 
+/// Marks a type as being the type that contract functions are attached for.
+///
+/// Use `#[contractimpl]` on impl blocks of this type to make those functions
+/// contract functions.
+///
+/// Note that a crate only ever exports a single contract. While there can be
+/// multiple types in a crate with `#[contract]`, when built as a wasm file and
+/// deployed the combination of all contract functions and all contracts within
+/// a crate will be seen as a single contract.
```
jayz22, line 273: This is interesting, and a little counter-intuitive. Do we currently have an example of a single crate with multiple `#[contract]`s built into a single contract?

PR #1012, opened 2023-06-23
Split up the token interface trait

### LINKED ISSUE #965: Split up Token Interface

leighmcculloch:
### What
Separate admin functions in the token interface out to a separate trait.

### Why
The primary standard interface of tokens won't include the admin functions. Although many tokens on Soroban will have the admin functions because all Stellar Asset Contracts will have them.

We still want it to be convenient for folks to follow that same admin pattern if its a fit, but contracts shouldn't have to implement them if their follow a different pattern for admin operations.

This issue was discussed [on Discord](https://discord.com/channels/897514728459468821/1114325793926037605), and some folks suggested allowances should also be broken into a separate trait. However this seemed more controversial and there were also opinions that they shouldn't be separated, so I've kept them in for now.

Note that the Spec value in the SDK will continue to hold the entire token interface. It's not particularly useful to split that up, as the only user of it is the soroban-cli when interacting with Stellar Asset Contracts. I've hidden it from docs since it isn't routinely useful to most.

Close #965

### DIFF
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -21,6 +21,7 @@ use crate::{contractclient, contractspecfn, Address, Env, String};
 /// Spec contains the contract spec of Token contracts, including the general
 /// interface, as well as the admin interface, such as the Stellar Asset
 /// Contract.
+#[doc(hidden)]
 pub struct Spec;
 
 /// Interface for Token contracts, such as the Stellar Asset Contract.
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -18,8 +18,9 @@ use crate::{contractclient, contractspecfn, Address, Env, String};
 // 2. The implementations have been replaced with a panic.
 // 3. &Host type usage are replaced with Env
 
-/// Spec contains the contract spec of Token contracts, such as the Stellar
-/// Asset Contract.
+/// Spec contains the contract spec of Token contracts, including the general
+/// interface, as well as the admin interface, such as the Stellar Asset
+/// Contract.
 pub struct Spec;
 
 /// Interface for Token contracts, such as the Stellar Asset Contract.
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/token.rs
```diff
@@ -156,6 +158,46 @@ pub trait Interface {
...
+    ///
+    /// If the contract has not yet been initialized.
+    fn symbol(env: Env) -> String;
+}
+
+/// Interface for admin capabilities for Token contracts, such as the Stellar
+/// Asset Contract.
+#[contractspecfn(name = "Spec", export = false)]
+#[contractclient(crate_path = "crate", name = "AdminClient")]
+pub trait AdminInterface {
```
dmkozh, line 187: Shall we consider removing this completely? I'm not sure if there is a lot of value in standardizing the admin interface TBH, at least not necessarily at SDK level (also stuff like clawback may be better off as optional). I'm ok with this PR for now, just a consideration.

dmkozh (approved):

soroban-sdk/src/token.rs
leighmcculloch: I'm also wondering what to do about this. The admin interface is really only here because we need a client in the SDK for calling the stellar asset contract's admin functions.

Instead of having the admin interface here, maybe we move it to a `stellar_asset` module? We've also been trying to keep stellar specific concepts out of the SDK though.

leighmcculloch: I agree this is somethign we should continue to think about and increment on.

dmkozh: Yeah, I think there were several discussions where we agreed that it's ok to provide admin functionality just for SAC. So the move would probably make things more clear.

leighmcculloch: Opened issue to continue discussion: https://github.com/stellar/rs-soroban-sdk/issues/1014

PR #1013, opened 2023-06-23
Add some basic U256 and I256 doc comments

leighmcculloch:
### What
Add some basic U256 and I256 doc comments.

### Why
The types have zero comments which looks pretty odd in the docs. We can improve these over time, but this bare minimum should help.


jayz22:
Thanks @leighmcculloch , I will add more in https://github.com/stellar/rs-soroban-sdk/pull/1000

graydon (approved):

PR #1016, opened 2023-06-23
Add sdk version to contract meta automatically

### LINKED ISSUE #735: Embed Build and Version into contracts
Months ago we talked about the possibility of embedding meta information, such as the version of SDK and Auth SDK into contracts. This issue is to debate that.

The main requirement that drove the initial intent of this was so that it would help us debug problems in contracts as we release new SDKS, new Envs, etc. We already get visibility into this to a degree with the env interface version that is embedded into built contracts in the 'env meta' entries.

So there's two time frames of uses:
- Before a pubnet release we are likely to have many futurenets which are incompatible with each other, and a fast evolving SDK that accompanies them. This seems very useful as we've generally found debugging what a contract has used or been built with as difficult.
- After pubnet it may be useful, but unclear how much.

Because of that second point if we implement this we should at a point close to pubnet launch consider removing this feature. It may have limited utility past that point.

This isn't urgent, but if we agree on implementing something like this, I'd like to get this into a near future release so we don't have too many contracts out there that we can't help people debug their version and state of why it isn't working.

cc @graydon

leighmcculloch:
### What
Add rust version and sdk version to contract meta automatically.

Rustc version is stored in meta under the key `rsver`.

SDK version is stored in meta under the key `rssdkver`.

```
Section `contractmetav0` (100 bytes):
Length: 100 (0x64) bytes
0000:   00 00 00 00  00 00 00 05  72 73 76 65  72 00 00 00   ........rsver...
0010:   00 00 00 06  31 2e 37 30  2e 30 00 00  00 00 00 00   ....1.70.0......
0020:   00 00 00 08  72 73 73 64  6b 76 65 72  00 00 00 34   ....rssdkver...4
0030:   30 2e 38 2e  34 23 32 62  33 30 65 34  37 33 36 36   0.8.4#2b30e47366
0040:   39 66 61 63  38 30 33 31  38 38 66 61  63 30 32 63   9fac803188fac02c
0050:   62 63 66 61  36 30 39 63  31 39 64 63  63 36 2d 64   bcfa609c19dcc6-d
0060:   69 72 74 79                                          irty
```

### Why
The rustc version is a primitive for future supporting reproducible builds.

The SDK version is for debugging situations where incompatibilities arise across different versions, and for better understanding them.

Close #735


graydon (approved):

soroban-sdk/src/tests.rs
```diff
@@ -14,10 +14,8 @@ mod contract_store;
 mod contract_udt_enum;
 mod contract_udt_struct;
 mod contract_udt_struct_tuple;
-mod contractfile_with_sha256;
 mod contractimport;
 mod contractimport_with_error;
-mod contractimport_with_sha256;
```
leighmcculloch, line 20: I removed these two tests because frankly they're a pain to keep updated. The sha256 is always changing. With this pull request they're impossible to keep updated because the repo builds the git commit into the wasm, therefore the sha256 cannot be known.

PR #1017, opened 2023-06-24
Add instance storage wrapper

dmkozh:
### What

Add instance storage wrapper

### Why

Instance storage is introduced in host to make it cheaper to store data bound to the contract instance.

### Known limitations

N/A


soroban-sdk/src/storage.rs
```diff
@@ -76,18 +76,59 @@ impl Storage {
         Storage { env: env.clone() }
     }
 
+    /// Storage for data that can stay in the ledger forever until deleted.
+    ///
+    /// Persistent entries might be 'evicted' from the ledger if they run out
+    /// of the rent balance. However, evicted entries can be restored and also
+    /// they cannot be recreated. This means that semantically these entries
+    /// behave 'as if' they were stored in the ledger forever.
```
leighmcculloch, line 84: Can we use the term expired only? Eviction is a bit of an implementation detail. A contract really only deals with live and expired, and contracts don't see the difference between expired and evicted. Suggest:
```suggestion
    /// Persistent entries might expire and be removed from the ledger if they run out
    /// of the rent balance. However, expired entries can be restored and
    /// they cannot be recreated. This means these entries
    /// behave 'as if' they were stored in the ledger forever.
```

soroban-sdk/src/storage.rs
```diff
@@ -76,18 +76,59 @@ impl Storage {
...
     pub fn persistent(&self) -> Persistent {
         Persistent {
             storage: self.clone(),
         }
     }
 
+    /// Storage for data that may stay in ledger only for a limited amount of
+    /// time.
+    ///
+    /// This is cheaper than `persistent()` storage.
```
leighmcculloch, line 97:
```suggestion
    /// Temporary storage are cheaper than Persistent storage.
```

soroban-sdk/src/storage.rs
```diff
@@ -76,18 +76,59 @@ impl Storage {
...
+    ///
+    /// Temporary entries will be removed from the ledger after their lifetime
+    /// ends. Removed entries can be created again, potentially with different
+    /// values.
+    ///
+    /// This should be used for data that needs to only exist for a limited
+    /// period of time, such as oracle data, claimable balances, offerc etc.
```
leighmcculloch, line 104:
```suggestion
    /// period of time, such as oracle data, claimable balances, offer, etc.
```

soroban-sdk/src/storage.rs
```diff
@@ -76,18 +76,59 @@ impl Storage {
...
             storage: self.clone(),
         }
     }
 
+    /// Storage for a **small amount** of persistent data associated with
+    /// the current contract's instance.
+    ///
+    /// All the `instance` entries are stored in a single `persistent()` ledger
+    /// entry. That makes it cheap to use this, but also limits the total
+    /// size of the stored data.
```
leighmcculloch, line 116: We're discussing implementation details. We should discuss attributes / properties of instance storage instead: loaded automatically and doesn't appear in footprints, increases the cost of every invocation of the contract whether the data is used or not.

We discuss limits here, but don't define any concrete limits. If we discuss the attributes / properties ☝🏻 it might be clearer if we follow with, "therefore due to this qualities of instance data, it is really only suitable for small data."

leighmcculloch (approved):
Couple doc comments. Otherwise lgtm.

soroban-sdk/src/storage.rs
dmkozh: Ok, I just thought it's the other way around and 'evicted' si _the_ term.

soroban-sdk/src/storage.rs
dmkozh: I've tried to rephrase this a bit. There is a lot of information to convey; I'm still not sure how to structure this in the most efficient fashion.

soroban-sdk/src/storage.rs
```diff
@@ -76,18 +76,69 @@ impl Storage {
...
+    /// Instance storage is tightly coupled with the contract instance: it will
+    /// be loaded from the ledger every time the contract instance itself is 
+    /// loaded. It also won't appear in the ledger footprint. *All*
+    /// the data stored in the instance storage is read from ledger every time
+    /// the contract is used and it doesn't matter whether contract uses the
+    /// storage or not.
+    /// 
+    /// This has the same lifetime properties as Persistent storage, i.e.
+    /// the data semantically stays in the ledger forever and can be
+    /// evicted/restored.
```
leighmcculloch, line 126:
```suggestion
    /// expired/restored.
```

leighmcculloch (approved):

soroban-sdk/src/storage.rs
leighmcculloch: I think this is really good. There's a lot to communicate, I think the comment does a good job of picking and choosing.

leighmcculloch:
Seems to have a compile error.

dmkozh:
> Seems to have a compile error.

Oh, sure, this is blocked by env pr https://github.com/stellar/rs-soroban-env/pull/886

PR #1018, opened 2023-06-24
Remove unreachable strings from wasm file

### LINKED ISSUE #1001: Remove unreachable strings in compiled wasm
@mootz12 reported that they're seeing strings in the compiled wasm, and on inspection it appears the strings are caused by the `unreachable!` macro.

We should be able to replace the `unreachable!` macro for wasm builds so that instead the wasm32 unreachable is called instead.

Discussion at https://discord.com/channels/897514728459468821/1121133500179951756/1121155657756848280

leighmcculloch:
### What
Use wasm32 unreachable in wasm builds instead of the unreachable macro.

### Why
The unreachable macro injects a panic string into the wasm, which we don't use and isn't useful since the wasm will just abort.

Close #1001 

### Perf Diffs

The result is smaller wasm files:

```diff
--rwxr-xr-x  1 leighmcculloch  staff  741 Jun 23 20:12 test_errors.wasm
+-rwxr-xr-x  1 leighmcculloch  staff  646 Jun 23 20:17 test_errors.wasm
```

And the string is no longer in the data section of the wasm file:

```diff
-  (data (;0;) (i32.const 1048576) "internal error: entered unreachable codepersistednot implementedhello"))
+  (data (;0;) (i32.const 1048576) "persistedhello"))
```


soroban-sdk/src/env.rs
```diff
@@ -259,9 +259,9 @@ impl Env {
     ///
     /// Equivalent to `panic!`, but with an error value instead of a string.
     #[doc(hidden)]
+    #[inline(always)]
     pub fn panic_with_error(&self, error: impl Into<internal::Error>) {
         _ = internal::Env::fail_with_error(self, error.into());
-        unreachable!()
```
leighmcculloch, line 264: As far as I can tell this unreachable is unnecessary. The reason we benefit from unreachable in the macro is so that we let the compiler know the code path will stop wherever the macro is used.

graydon: Mm. I would actually think the content of the macro should move here and this function should be typed as `!` rather than `()`, but maybe there's some context where you want it to be typed this way? I don't really know why there's a macro at all tbh.

soroban-sdk/src/lib.rs
```diff
@@ -641,6 +641,9 @@ pub use soroban_sdk_macros::contractfile;
 macro_rules! panic_with_error {
     ($env:expr, $error:expr) => {{
         $env.panic_with_error($error);
+        #[cfg(target_family = "wasm")]
+        core::arch::wasm32::unreachable();
+        #[cfg(not(target_family = "wasm"))]
         unreachable!();
```
graydon, line 647: $env.panic_with_error($error)

graydon, line 647:
```suggestion
        $env.panic_with_error($error)
```

soroban-sdk/src/env.rs
```diff
@@ -259,9 +259,9 @@ impl Env {
     ///
     /// Equivalent to `panic!`, but with an error value instead of a string.
     #[doc(hidden)]
+    #[inline(always)]
     pub fn panic_with_error(&self, error: impl Into<internal::Error>) {
         _ = internal::Env::fail_with_error(self, error.into());
```
graydon, line 264:
```suggestion
    pub fn panic_with_error(&self, error: impl Into<internal::Error>) -> ! {
        _ = internal::Env::fail_with_error(self, error.into());
        #[cfg(target_family = "wasm")]
        core::arch::wasm32::unreachable()
        #[cfg(not(target_family = "wasm"))]
        unreachable!()
```

soroban-sdk/src/env.rs
leighmcculloch: Done in bfcc121. Thank you!

graydon (approved):

PR #1021, opened 2023-06-24
Derive standard traits for derived SorobanArbitrary::Prototype types.

brson:
### What

In `#[contracttype]` derive Clone, Eq, PartialEq, Ord, PartialOrd for generated SorobanArbitrary::Prototype types.

### Why

All the handwritten types used as SorobanArbitrary::Prototype implement these traits, and while writing fuzzing docs I found a case where I needed to clone one of these generated types.

### Known limitations

na


leighmcculloch (approved):

leighmcculloch:
Thanks @brson! 👏

PR #1023, opened 2023-06-27
Expose PRNG functions

### LINKED ISSUE #969: surface PRNG functionality in SDK
There are some new host functions that should surface in the SDK now.

masonforest:
### What

Expose `prng_reseed`,  `u64_in_inclusive_range` and `vec_shuffle` in the SDK.

### Why

Exposing the PRNG functions in the SDK allows users to utilize randomness provided by the host in their smart contracts.

### Known limitations

N/A

Close #969

-----

I added tests for `sha256` and `verify_sig_ed25519` while I was in here. Let me know if those changes should be moved to a separate PR. I considered adding a PRNG module separate from the `crypto` module. I decided to keep them under `crypto` because there are only three functions. Also this is also similar `getRandomValues` is a function implemented within the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) in browsers.


soroban-sdk/src/crypto.rs
```diff
@@ -22,6 +22,28 @@ impl Crypto {
         unsafe { BytesN::unchecked_new(env.clone(), bin) }
     }
 
+    // Reseeds the pseudorandom number generator (PRNG) with the provided `seed` value.
+    pub fn prng_reseed(&self, seed: &Bytes) {
+        let env = self.env();
```
jayz22, line 27: I would add `env.check_same_env(seed.env())` to make sure these two are from the same `env`.

soroban-sdk/src/crypto.rs
```diff
@@ -22,6 +22,28 @@ impl Crypto {
...
+    // Returns a random u64 in the range between `lower` and `upper` inclusive.
+    pub fn u64_in_inclusive_range(&self, lower: u64, upper: u64) -> u64 {
+        let env = self.env();
+        internal::Env::prng_u64_in_inclusive_range(env, lower.into(), upper.into())
+            .unwrap_infallible()
+            .into()
+    }
+
+    // Shuffles a given vector v using the Fisher-Yates algorithm.
+    pub fn vec_shuffle(&self, v: VecObject) -> VecObject {
```
jayz22, line 40: I would pass in `v: Vec` instead, and call `env.check_same_env(v.env())` to make sure they from the same env.

soroban-sdk/src/crypto.rs
jayz22: I notice this check is not present in other functions but they are nice (and also inexpensive) to have .

jayz22 (commented):
Thanks for contributing @masonforest ! 
Just a couple of suggestions on env sameness check. 
Overall it looks great and thanks for adding the additional tests.

soroban-sdk/src/crypto.rs
jayz22: (See updated comment above on `check_same_env`)

masonforest: Ah that makes sense. Done 👍 -> [Pass a Vec to vec_shuffle instead of a VecObject](https://github.com/stellar/rs-soroban-sdk/pull/1023/commits/749973904a74135856edc45110f708df58eec2d0)

soroban-sdk/src/crypto.rs
masonforest: Cool, I added those checks [here](https://github.com/stellar/rs-soroban-sdk/pull/1023/commits/05e9e686fa33670189a4394669798f106caa4dea) if we want them. Easy to revert if not.

jayz22 (approved):
Thanks!

soroban-sdk/src/crypto.rs
graydon: I don't feel strongly about these existing or not; as @jayz22 says they are not meaningful in production code, only testing, and it's likely that if you have two hosts in a test you've got bigger problems. but at the same time it's harmless to add them.

graydon (dismissed):
Sorry I'm late to the game on this one. The code of this is I think probably fine (I don't see anything problematic) but it's _essential_ that the PRNG functions _not_ live in a module called `crypto`, and that they instead be very clearly marked and documented as _not_ providing cryptographically strong randomness of the sort cryptographic applications need.  The `P` in `PRNG` and the term `pseudo` should be prominent, ideally every function should retain `prng_...` as a prefix, and documentation should repeat in big blinking letters that this is non-cryptographic, non-random, pseudorandom generation based on strictly public information.

 In other words, the analogy you're making to WebCrypto's `getRandomValues` function -- which provides access to an OS's cryptographically strong random number source -- is _dangerously wrong_ and we should be extremely clear to users to ensure they _never_ make that mistake -- if they do they will fatally undermine their code.

I am happy to write a block comment describing the potential contexts where these functions are appropriate, but they are _very_ inappropriate for cryptographic uses.

masonforest:
Thank you for the feedback @graydon and thank you for all you've done for the open source community.

> it's essential that the PRNG functions not live in a module called crypto

I moved these functions to a new module called `Prng` which is more appropriate and added the `prng` prefix to all of the functions.

> I am happy to write a block comment describing the potential contexts where these functions are appropriate, but they are very inappropriate for cryptographic uses.

Could you post that comment here? Or feel free to add it yourself before merge. If there is somewhere this is documented which I can synthesize into a block comment I'd be happy to do that as well. Otherwise, I think I've done all I'm able to do here.

soroban-sdk/src/prng.rs
```diff
@@ -0,0 +1,48 @@
+//! Prng contains functions for pseudo-random functions.
+use crate::{
+    env::internal, unwrap::UnwrapInfallible, Bytes, Env, IntoVal, TryIntoVal, Val,
+    Vec,
+};
+
+/// Prng provides access to pseudo-random  functions.
```
leighmcculloch, line 7:
```suggestion
/// Prng provides access to functions that operate a pesudo-random number generator.
```

soroban-sdk/src/prng.rs
```diff
@@ -0,0 +1,48 @@
...
+    // Returns a random u64 in the range between `lower` and `upper` inclusive.
+    pub fn prng_u64_in_inclusive_range(&self, lower: u64, upper: u64) -> u64 {
+        let env = self.env();
+        internal::Env::prng_u64_in_inclusive_range(env, lower.into(), upper.into())
+            .unwrap_infallible()
+            .into()
+    }
+
+    // Shuffles a given vector v using the Fisher-Yates algorithm.
+    pub fn prng_vec_shuffle<V>(&self, v: V) -> Vec<Val>
```
leighmcculloch, line 36: @graydon I noticed you requested `prng_` prefixes on all the functions. Is that necessary? Calls to the function will already be prefix as they'll look like this `env.prng().reseed(...)`.

soroban-sdk/src/prng.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) fn new(env: &Env) -> Prng {
+        Prng { env: env.clone() }
+    }
+
+    pub fn env(&self) -> &Env {
+        &self.env
+    }
+
+    // Reseeds the pseudorandom number generator (PRNG) with the provided `seed` value.
+    pub fn prng_reseed(&self, seed: &Bytes) {
```
leighmcculloch, line 22: @graydon Can we name this function `seed` instead of `reseed`? Most APIs I've seen simply call it seed, because the action is seeding the generator. The fact it was already seeded by default, which the `re` signals, is a concern of the generator in general, and not a concern of this specific function.
```suggestion
    pub fn seed(&self, seed: &Bytes) {
```

soroban-sdk/src/prng.rs
```diff
@@ -0,0 +1,48 @@
...
+        let env = self.env();
+        internal::Env::prng_reseed(env, seed.into()).unwrap_infallible();
+    }
+
+    // Returns a random u64 in the range between `lower` and `upper` inclusive.
+    pub fn prng_u64_in_inclusive_range(&self, lower: u64, upper: u64) -> u64 {
```
leighmcculloch, line 28: @graydon Can we rename this function to `u64_in_range`, and replace the `lower: u64`, `upper:64` parameters with `r: impl RangeBounds<u64>`?
```suggestion
    pub fn u64_in_range(&self, r: impl RangeBounds<u64>) -> u64 {
```

It's allows developer to specify the random using rust range syntax, which supports both inclusive and exclusive ranges. Use use this pattern in the Vec type's slice function already. For an example for how it works see:
https://github.com/stellar/rs-soroban-sdk/blob/1def3c6e5e87acc71bd63c71295f9588edd3805b/soroban-sdk/src/vec.rs#L761-L774

soroban-sdk/src/prng.rs
leighmcculloch, line 36: Could we remove `vec` from the function name? This will be usable with non-Vec's because it internally takes care of converting things to a Vec to make it work. The vec feels unnecessarily restrictive in name when not in implementation.
```suggestion
    pub fn shuffle<V>(&self, v: V) -> Vec<Val>
```

soroban-sdk/src/lib.rs
```diff
@@ -767,6 +767,7 @@ pub mod iter;
 pub mod ledger;
 pub mod logs;
 mod map;
+pub mod prng;
```
leighmcculloch, line 770: @graydon Do you feel strongly about this package and type being named "prng"? I think rnd instead of rng is more common abbreviation in libs and APIs and may be more easily recognizable. I don't feel strongly if this is important as is.

leighmcculloch (changes requested):

leighmcculloch:
Thanks @masonforest for leading the charge on this. I'm going to make the changes above, and some other additions, and see if we can slide this into the next release.

soroban-sdk/src/env.rs
```diff
@@ -300,6 +300,17 @@ impl Env {
         Crypto::new(self)
     }
 
+    /// Get a [Prng] for accessing the current functions which provide pseudo-randomness.
```
dmkozh, line 303: I think we should mention some additional important properties somewhere, such as:

- every contract invocation gets its own, independent seed, thus it is not possible to manipulate prng of another contract
- prng is not replayable, if an invocation fails, next invocation will still get a different seed.
- The rng results are observable immediately after the transaction is included into ledger (that's mentioned in the file, but I think it's important enough to also be mentioned at the top level)

Feel free to edit/modify this as you see fit. Also I realize that SDK is probably not the best place to document the network-dependent details, but OTOH I'm not sure what else is.

dmkozh (commented):
A general comment on the 'not secure' comments: while I understand what these are trying to convey, I think these are somewhat misleading and might lead to users using even worse approaches to randomness (e.g. seeding with ledger seq/timestamp as they do now).

I think what the comments mostly imply is that one shouldn't try to e.g. generate private keys with this. However, the main factor for that is the observability/reproducibility, which makes such usages impossible, no matter how good initial entropy is.  

So I think it's worth to at least state that while this is imperfect, is is as good for randomness, as it can get given the restrictions of running in blockchain.

dmkozh (approved):

soroban-sdk/src/env.rs
leighmcculloch: Thanks, great call out. I'll add these details into the module doc at the top.

> I realize that SDK is probably not the best place to document the network-dependent details

I was wondering this too. I added a note at the top that is really Stellar specific. I'll play with that and see if I can remove it.

leighmcculloch: Pushed updated docs. Lmk if you think anything should be changed.

leighmcculloch:
@graydon I'd appreciate your review of this, but I'm going to merge this now. Could circle back on it before we cut the next release?

PR #1032, opened 2023-07-10
Add Arbitrary impl for Duration and Timepoint

brson:
### What

Implement Arbitrary for Duration and Timepoint.

Also add some missing test cases.

### Why

These are both user-facing contract types. Implementing Arbitrary allows them to be accepted as inputs to fuzz tests.

### Known limitations

na


brson:
Rebased.

graydon (approved):

PR #1043, opened 2023-07-24
fix: use uppercase const name

chadoh:
### What

This fixes an issue where Rust Analyzer sometimes complains that:

    constant `symbol` should have UPPER_SNAKE_CASE name

### Why

When using `symbol_short!`, as recommended by the Getting Started → Hello World tutorial, I was getting confusing warnings from Rust Analyzer:

image

### Known limitations

Why not use this instead?

```rs
Ok(_) => quote! {
    #crate_path::Symbol::short(#s)
}
```

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -58,8 +58,8 @@ fn _symbol_short(crate_path: &str, s: &LitStr) -> TokenStream {
     match Symbol::try_from_small_str(&s.value()) {
         Ok(_) => quote! {{
             #[allow(deprecated)]
-            const symbol: #crate_path::Symbol = #crate_path::Symbol::short(#s);
-            symbol
+            const SYMBOL: #crate_path::Symbol = #crate_path::Symbol::short(#s);
+            SYMBOL
         }}
         .into(),
         Err(e) => Error::new(s.span(), format!("{e}"))


tomerweller (approved):

soroban-sdk-macros/src/lib.rs
```diff
@@ -58,8 +58,8 @@ fn _symbol_short(crate_path: &str, s: &LitStr) -> TokenStream {
     match Symbol::try_from_small_str(&s.value()) {
         Ok(_) => quote! {{
             #[allow(deprecated)]
-            const symbol: #crate_path::Symbol = #crate_path::Symbol::short(#s);
-            symbol
+            const SYMBOL: #crate_path::Symbol = #crate_path::Symbol::short(#s);
+            SYMBOL
         }}
```
chadoh, line 63: Why not this?

```rs
Ok(_) => quote! {
    #crate_path::Symbol::short(#s)
}
```

chadoh: @willemneal and I thought it was probably best to stick with the convention already in the code, but we're curious why the simpler alternative isn't being used.

tomerweller: @chadoh can you open a different PR for that? I'll resolve this so we can merge this one for now

leighmcculloch: The `Symbol::short` function call needs to be made in a const context. That's what the macro is creating a short hand for. Assigning the value to a `const` is how we force the expression to be evaluated at compile time. If the expression is executed outside of a const context it'll be evaluated at runtime.

chadoh: Fascinating, thanks @leighmcculloch!

PR #1055, opened 2023-08-15
Correct comment on String::copy_into_slice

graydon:
The copy routine here actually traps if you pass an output slice larger than the string. I think this is correct, since the previously-documented not-actually-performed behaviour (taking the minimum size of the output or the string) is a potential footgun / attack vector if someone tricks a contract into "copying N bytes" out of a string much smaller than N -- they get to basically force the output to be whatever the buffer's content was. The newly documented (already existing) behaviour does less magic / is less surprising / fails under more abnormal circumstances.

### DIFF
--- a/soroban-sdk/src/string.rs
+++ b/soroban-sdk/src/string.rs
@@ -211,11 +211,15 @@ impl String {
 
     /// Copy the bytes in [String] into the given slice.
     ///
-    /// The minimum number of bytes are copied to either exhaust [String] or fill
-    /// slice.
+    /// ### Panics
+    ///
+    /// If the output slice and string are of different lengths.
     #[inline(always)]
     pub fn copy_into_slice(&self, slice: &mut [u8]) {
         let env = self.env();
+        if self.len() as usize != slice.len() {
+            sdk_panic!("String::copy_into_slice with mismatched slice length")
+        }
         env.string_copy_to_slice(self.to_object(), Val::U32_ZERO, slice)
             .unwrap_optimized();
     }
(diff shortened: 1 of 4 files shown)


dmkozh (approved):

soroban-sdk/src/string.rs
```diff
@@ -211,8 +211,8 @@ impl String {
 
     /// Copy the bytes in [String] into the given slice.
     ///
-    /// The minimum number of bytes are copied to either exhaust [String] or fill
-    /// slice.
+    /// The number of bytes copied is dictated by the size of `slice`,
+    /// and if the [String] is not long enough the call traps.
```
leighmcculloch, line 215: The same comment needs correcting on the `Bytes::copy_into_slice` function.

leighmcculloch:
> the previously-documented not-actually-performed behaviour (taking the minimum size of the output or the string) is a potential footgun / attack vector if someone tricks a contract into "copying N bytes" out of a string much smaller than N -- they get to basically force the output to be whatever the buffer's content was.

I think it could be argued that `copy_into_slice` should fail unless the destination is exactly the same size. Right now it only fails if the source is smaller than the destination, but it silently works fine if it is larger than the destination.

For example, this code will work without error:

```rust
let env = Env::default();
let expected = "my string";
let my_string = String::from_slice(&env, expected);

let mut short_array: [u8; 2] = [0; 2];
my_string.copy_into_slice(&mut short_array);

assert_eq!(short_array, "my".as_bytes());
```

(Code is taken from an example posted by @vinamogit on the [stellar-dev Discord].)

[stellar-dev Discord]: https://discord.com/channels/897514728459468821/1141141399530635305/1141141399530635305

graydon:
@leighmcculloch I can see making that change on the specific API in the SDK you're describing; in the host interface it's actually _more_ general already (it takes a start-position within the string, i.e. it allows not just doing a copy that's short from either the start or the end of the string). Would you want the env interface to be tightened also, requiring the user to do an explicit string-slice host function call before doing a copy-out call?

leighmcculloch:
I think it's reasonable to cover the footgun in the SDK without an env change.

graydon:
@leighmcculloch ok I've updated it to panic on length mismatch, lmk if this is more to your liking.

Issue #1011, opened 2023-06-23
Linux arm64 / aarch64 builds of contracts broken


leighmcculloch:
It appears that linux arm64 / aarch64 builds of contracts fail with errors about relocations.

This is an example failing build:
> /usr/bin/ld: /home/runner/work/rs-soroban-sdk/rs-soroban-sdk/target/aarch64-unknown-linux-gnu/debug/deps/soroban_token_sdk.1c28rz4gpwkk2qll.rcgu.o: Relocations in generic ELF (EM: 183)
>           /usr/bin/ld: /home/runner/work/rs-soroban-sdk/rs-soroban-sdk/target/aarch64-unknown-linux-gnu/debug/deps/soroban_token_sdk.1c28rz4gpwkk2qll.rcgu.o: error adding symbols: file in wrong format

https://github.com/stellar/rs-soroban-sdk/actions/runs/5360474620/jobs/9725396502?pr=1009

It's unclear how long this has been broken because we are only just adding CI for a variety of operating systems now in #1009.

The impact of this is that anyone using Linux with arm64 / aarch64 will not be able to test contracts. They'll be able to build to wasm32, but not to the host architecture.

While Linux with arm64 / aarch64 is not common as a workstation configuration, it is common in Docker images because any macOS user who has an Apple Silicon chip will be running Linux with arm64 / aarch64 images by default.

leighmcculloch:
It's possible that the recent addition of `ctor` in https://github.com/stellar/rs-soroban-sdk/pull/1008 is the cause.

graydon:
This might or might-not block final -- depends on whether it's just preventing contract development (shouldn't block, not too many users are likely to use aarch64 hosts for testing) or whether it's preventing building and running core (should probably block, operators are fairly likely to deploy core on a graviton host)

graydon:
An update on this:

  - Building for aarch64 _on aarch64 hosts_ does not show this problem, only when _crossing_ from x64-linux to aarch64-linux (as we do in CI). I think this means that the M1-mac-with-docker `soroban contract build` scenario will not be a problem.
  - It also only happens when you have `crate-type = ["cdylib"]` in your `Cargo.toml`. If you build with _no_ crate type, or with `crate-type = ["lib"]`, it builds native libraries fine (I think static libs / rlibs).

IOW I think this is just a bug in the way cargo invokes the target linker when cross-compiling cdylibs (by the look of it, I think it's actually accidentally invoking the host linker).

[...]

PR #1060, opened 2023-08-18
re-enable linux arm64 builds

graydon:
Fix https://github.com/stellar/rs-soroban-sdk/issues/1011 by avoiding the use of `crate-type = ["cdylib"]` in native builds. 

It was causing some linker errors when cross-compiling from x64-linux to aarch64-linux, I believe the root cause is https://github.com/rust-lang/cargo/issues/4133 i.e. cargo failing to select the right target linker during cross builds.

The `cdylib` crate type is only relevant when building wasms, so this change just lowers that functionality into the specific commands that build wasms.


.github/workflows/rust.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       with:
         name: cargo-hack
         version: 0.5.16
-    - run: cargo-hack hack build --target wasm32-unknown-unknown --profile release
+    - run: for i in $(ls tests) ; do cargo-hack hack rustc --target wasm32-unknown-unknown --profile release --crate-type cdylib -p test_${i} ; done
```
leighmcculloch, line 86: Cargo-hack can do the lifting on assembling the list of packages to build, and if we exclude the problem package we can build more than the tests, but we can check that the SDK and other packages also build to wasm.
```suggestion
    - run: cargo-hack hack rustc --target wasm32-unknown-unknown --profile release --crate-type cdylib --workspace --exclude soroban-sdk-macros
```

Makefile
```diff
@@ -13,7 +13,7 @@ test: fmt build
 	cargo hack --feature-powerset --ignore-unknown-features --features testutils --exclude-features docs $(CARGO_TEST_SUBCOMMAND)
 
 build: fmt
-	cargo hack build --target wasm32-unknown-unknown --release
+	for P in $(shell ls tests) ; do cargo hack rustc --target wasm32-unknown-unknown --release --crate-type cdylib -p test_$$P; done
```
leighmcculloch, line 16:
```suggestion
	cargo-hack hack rustc --target wasm32-unknown-unknown --profile release --crate-type cdylib --workspace --exclude soroban-sdk-macros
```

soroban-token-sdk/Cargo.toml
```diff
@@ -11,7 +11,6 @@ edition = "2021"
 rust-version = "1.71"
 
 [lib]
-crate-type = ["cdylib", "rlib"]
```
leighmcculloch, line 14: I'm pretty sure we don't need to remove the `crate-type` fields. They can stay, which will make the tests in this repository more closely match examples and the guides. Can we keep them?

leighmcculloch (commented):
Can we make a couple tweaks? See inline.

soroban-token-sdk/Cargo.toml
graydon: I'm .. not sure what you mean. The whole point of this PR is removing them, because they force `cdylib` mode when building natively (when `cdylib` is only even appropriate / useful for wasm).

graydon: (And doesn't work for cross compiling)

graydon: (I should also mention: `soroban contract build` runs `cargo rustc --crate-type cdylib` so there's not a lot of need to have a `crate-type` specified in the examples either, unless you really want `cargo build --release --target wasm32-unknown-unknown` _specifically_ to build a `.wasm` file. I got the impression we were leaning into telling users to run `soroban contract build` anyways?)

leighmcculloch: The main reason we kept the `cdylib` still in the templates is for when folks are running traditional cargo commands so that they still get .wasm files. 

Does removing the values fix the problem, or does using the rustc command fix the problem, or both? I thought the rustc command is what fixed the problem.

leighmcculloch: I forgot also, the way that `soroban-cli` figures out which crates it should build .wasm files for, in a workspace, is it filters the crates in the workspace by crates that have `cdylib` in their `crate-type` list. It does this so that it can trivially build all contracts in a workspace where some crates are only shared libs, or are proc-macro libs.

leighmcculloch (approved):

PR #1061, opened 2023-08-18
Update map iterator to be index based

jayz22:
### What

Sdk part of https://github.com/stellar/rs-soroban-env/issues/992

### Why

### Known limitations

[TODO or N/A]


graydon (approved):
I would change `front_pos` and `back_pos` to `pos` and `len`, but otherwise looks good!

PR #1068, opened 2023-08-24
Handle errors in Events::all, Logs::all

brson:
### What

Change the testutils functions `Events::all` and `Logs::all` to return `Result`.

### Why

I am not sure this is the correct resolution, or if this is worth fixing at all.

These functions call `Host::get_events`, which returns `Result`. That function may return an error, in which case `Events::all` and `Logs::all` panic.

This patch avoids the panic, but doesn't necessarily improve the debugging experience, since calling these functions will still fail to retrieve useful event information.

`Host::get_events` calls `Events::externalize`, and that function may fail if any single event fails to serialize to XDR. It's not clear to me what the consequences of `externalize` failing are.

I have seen these functions fail after calling a contract while passing objects with invalid references (intentionally adversarial). Because those objects can't be visited by the host, serialization fails.

From a debugging standpoint, it would be more useful to return partial event lists, perhaps with the non-externalizable events replaced with a placeholder. But I also see that these events are externalized by the `Host::try_finish` method, which seems to be a step in finalizing a transaction, though I don't understand how it's used.

### Known limitations

These functions still fail to return any events if just one event can't be externalized.

### DIFF
--- a/soroban-sdk/src/events.rs
+++ b/soroban-sdk/src/events.rs
@@ -125,30 +125,24 @@ use crate::{testutils, xdr, Address, BytesN, TryIntoVal};
 #[cfg(any(test, feature = "testutils"))]
 #[cfg_attr(feature = "docs", doc(cfg(feature = "testutils")))]
 impl testutils::Events for Events {
-    fn all(&self) -> Vec<(crate::Address, Vec<Val>, Val)> {
+    fn all(&self) -> Result<Vec<(crate::Address, Vec<Val>, Val)>, soroban_env_host::HostError> {
         let env = self.env();
         let mut vec = Vec::new(env);
-        self.env()
-            .host()
-            .get_events()
-            .unwrap()
-            .0
-            .into_iter()
-            .for_each(|e| {
-                if let xdr::ContractEvent {
-                    type_: xdr::ContractEventType::Contract,
-                    contract_id: Some(contract_id),
-                    body: xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, data }),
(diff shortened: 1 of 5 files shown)


leighmcculloch:
> I have seen these functions fail after calling a contract while passing objects with invalid references (intentionally adversarial). Because those objects can't be visited by the host, serialization fails.

@graydon Do you know if we ended up adding safety checks on the host that would prevent passing invalid objects?

Once those checks are added, if we're still adding them, this should be a case that can't happen. Or rather, there will be a failure but the failure will be on the immediate call, not on serialization further down the flow.

I think any failure in those testutils functions should be considered a bug, which is why we opted with panics rather than returning Result. I don't expect a developer to handle the result.

brson:
If there have been changes to object validation recently then this fix could be obsolete. I only observed the behavior addressed here on a previous revision of the sdk, and did not think to test on main.

brson:
I am going to close this. I see that the main branch of soroban-env has object integrity checks that should make this patch irrelevant. I am seeing some strange behavior during event externalization still that I am still trying to understand. I'll file followup issues.

leighmcculloch:
> I am seeing some strange behavior during event externalization still that I am still trying to understand. I'll file followup issues.

I'm really interested to see more on this. Thanks!

brson:
> > I am seeing some strange behavior during event externalization still that I am still trying to understand. I'll file followup issues.
> 
> I'm really interested to see more on this. Thanks!

https://github.com/stellar/rs-soroban-env/pull/1028

PR #1086, opened 2023-09-15
Implement deployer functions that return the deployed contract id.

dmkozh:
### What

Implement deployer functions that return the deployed contract id.

### Why

These have the respective host fns now and thus can be implemented.

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/deploy.rs
+++ b/soroban-sdk/src/deploy.rs
@@ -22,7 +22,13 @@
 //! #     pub fn f(env: Env, wasm_hash: BytesN<32>) {
 //! #         let salt = [0u8; 32];
 //! #         let deployer = env.deployer().with_current_contract(salt);
-//! #         let contract_address = deployer.deploy(wasm_hash);
+//! #         // Deployed contract address is deterministic and can be accessed
+//! #         // without actually deploying the contract.
+//! #         let computed_deployed_address = deployer.deployed_address();
+//! #         let deployed_address = deployer.deploy(wasm_hash);
+//! #         if computed_deployed_address != deployed_address {
+//! #             panic!("computed deployed contract address has to be consistent with actual deployed address");
+//! #         }
 //! #     }
 //! # }
 //! #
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/deploy.rs
```diff
@@ -22,7 +22,13 @@
 //! #     pub fn f(env: Env, wasm_hash: BytesN<32>) {
 //! #         let salt = [0u8; 32];
 //! #         let deployer = env.deployer().with_current_contract(salt);
-//! #         let contract_address = deployer.deploy(wasm_hash);
+//! #         // Deployed contract address is deterministic and can be accessed
+//! #         // without actually deploying the contract.
+//! #         let computed_deployed_address = deployer.deployed_address();
+//! #         let deployed_address = deployer.deploy(wasm_hash);
+//! #         if computed_deployed_address != deployed_address {
+//! #             panic!("computed deployed contract address has to be consistent with actual deployed address");
```
leighmcculloch, line 30: Can we reduce this example as follows?
```suggestion
//! #         // Deployed contract address is deterministic and can be accessed
//! #         // before deploying the contract.
//! #         let _ = deployer.deployed_address();
```

I think including a comparison test in the example could be confusing to someone. Examples are for copying and pasting, and I don't think we hope folks will do the comparison themselves. Someone could interpret this to mean they need to compare it before and after deploy.

That would be a good thing to chuck in a test though in the sdk/tests folder.

soroban-sdk/src/deploy.rs
```diff
@@ -150,10 +156,14 @@ pub struct DeployerWithAddress {
 
 impl DeployerWithAddress {
     /// Return the address of the contract defined by the deployer.
-    #[doc(hidden)]
-    /// Returns what the address of the contract will be once deployed.
+    ///
+    /// The contract doesn't have to actually be deployed in order for this to
+    /// work.
```
leighmcculloch, line 161: Suggest taking the op to tell the story a little, and leading with when the function can be called, rather than the state of the contract.
```suggestion
    /// This function can be called at anytime, before or after the contract is deployed, because contract addresses are deterministic.
```

leighmcculloch (approved):
Awesome, thanks for plugging this gap. Suggestions inline for documentation.

soroban-sdk/src/deploy.rs
dmkozh: Yeah, we don't have great test coverage of this whole file for now (asset tests are needed as well), so probably tests should be done separately as a followup.

soroban-sdk/src/deploy.rs
dmkozh: Update the comment for both deployer types.

leighmcculloch (approved):

PR #1093, opened 2023-09-20
Adapt to minor env changes from code review

graydon:
Adapting SDK to https://github.com/stellar/rs-soroban-env/pull/1077 -- needs to have its refs updated when that merges.

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -1283,6 +1283,27 @@ impl Env {
 impl internal::EnvBase for Env {
     type Error = Infallible;
 
+    // This exists to allow code in conversion paths to upgrade an Error to an
+    // Env::Error with some control granted to the underlying Env (and panic
+    // paths kept out of the host). We delegate this to our env_impl and then,
+    // since our own Error type is Infallible, immediately throw it into either
+    // the env_impl's Error escalation path (if testing), or just plain panic.
+    #[cfg(not(target_family = "wasm"))]
+    fn error_from_error_val(&self, e: crate::Error) -> Self::Error {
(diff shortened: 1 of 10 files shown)


soroban-sdk/src/env.rs
```diff
@@ -1283,6 +1283,27 @@ impl Env {
...
+    // paths kept out of the host). We delegate this to our env_impl and then,
+    // since our own Error type is Infallible, immediately throw it into either
+    // the env_impl's Error escalation path (if testing), or just plain panic.
+    #[cfg(not(target_family = "wasm"))]
+    fn error_from_error_val(&self, e: crate::Error) -> Self::Error {
+        let host_err = self.env_impl.error_from_error_val(e);
+        #[cfg(any(test, feature = "testutils"))]
+        self.env_impl.escalate_error_to_panic(host_err);
+        #[cfg(not(any(test, feature = "testutils")))]
+        panic!("{:?}", host_err);
```
leighmcculloch, line 1297: Contracts and the SDK are never executed without tests in non-wasm mode. What use case are these two lines serving?

leighmcculloch (approved):

soroban-sdk/src/env.rs
graydon: Generating an `Infallible` when not-testutils (escalate_error_to_panic is gated on testutils)

graydon: (i.e. just to make it compile)

graydon: (and like .. on the off chance some weirdo runs it in that configuration. nobody will, but there needs to be a code path there)

PR #1094, opened 2023-09-20
Fix typos in arbitrary docs

brson:
### What

Just fixing a few typos.

### Why

na

### Known limitations

na

### DIFF
--- a/soroban-sdk/src/arbitrary.rs
+++ b/soroban-sdk/src/arbitrary.rs
@@ -203,11 +203,16 @@ mod api {
     /// Types that implement `SorabanArbitrary` have an associated "prototype"
     /// type that implements [`Arbitrary`].
     ///
-    /// This exists partly that the prototype can be named like
+    /// This exists partly so that the prototype can be named like
     ///
-    /// ```ignore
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/arbitrary.rs
```diff
@@ -203,11 +203,16 @@ mod api {
...
-    /// ```ignore
-    /// fuzz_target!(|input: <Bytes as SorobanArbitrary>::Arbitrary| {
-    ///   ...
+    /// ```
+    /// # macro_rules! fuzz_target {
+    /// #     (|$data:ident: $dty: ty| $body:block) => { };
+    /// # }
+    /// # use soroban_sdk::{Address, Env, Vec, Bytes};
+    /// # use soroban_sdk::arbitrary::SorobanArbitrary;
+    /// fuzz_target!(|input: <Bytes as SorobanArbitrary>::Prototype| {
```
brson, line 214: This line previously used the wrong associated type name. This change makes the example a doc-test so that can't happen again.

leighmcculloch: Ah this is very cool way to get that type checking while still not actually having to run the fuzzer in the test 👏🏻.

leighmcculloch (approved):
🎉

PR #1102, opened 2023-10-03
Feature: expose rem_euclid functions and add 128/256 bit conversions

mootz12:
### What

* Expose the `rem_euclid` functions added to the host in https://github.com/stellar/rs-soroban-env/pull/1091

* Add 128 bit to/from 256 bit helper functions

### Why

Context in env PR here: https://github.com/stellar/rs-soroban-env/pull/1091 

### Known limitations

None

### TODO:

- [x] update soroban-env version

### DIFF
--- a/soroban-sdk/src/num.rs
+++ b/soroban-sdk/src/num.rs
@@ -172,6 +172,13 @@ impl U256 {
         }
     }
 
+    pub fn from_u128(env: &Env, u: u128) -> Self {
+        let lo: Bytes = Bytes::from_array(env, &u.to_be_bytes());
+        let mut bytes: Bytes = Bytes::from_array(env, &[0u8; 16]);
+        bytes.append(&lo);
+        Self::from_be_bytes(env, &bytes)
+    }
+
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

soroban-sdk/src/num.rs
```diff
@@ -193,6 +200,17 @@ impl U256 {
         }
     }
 
+    pub fn to_u128(&self) -> Result<u128, ConversionError> {
```
leighmcculloch, line 203: I don't feel strongly about this, but I think it might be more common for this type of fallible conversion to return an `Option<>`. For example, the `.checked_add` function in the stdlib on integer types returns `Option<>` when an overflow occurs. https://docs.rs/num/latest/num/trait.CheckedAdd.html

leighmcculloch (commented):
One thing to consider about conversion conventions.

soroban-sdk/src/num.rs
mootz12: I was on the fence about `Result<_, ConversionError>`, and I think your suggestion makes more sense.

I'll get this updated!

leighmcculloch (approved):
I just reviewed d5fa16e, approving based on @dmkozh earlier review otherwise.

PR #1106, opened 2023-10-05
Add fn to expose max expiration available

leighmcculloch:
### What
Add fn to expose max expiration available.

### Why
It came up in the Discord meeting today that it would be helpful if contracts can access the max expiration as they may wish to simply bump to that. The host fn exists, but it isn't exposed in the SDK.


soroban-sdk/src/storage.rs
```diff
@@ -139,6 +139,14 @@ impl Storage {
         }
     }
 
+    /// Returns the maximum ledger sequence that an entry can live to
+    /// (inclusive).
+    pub fn max_expiration(&self) -> u32 {
+        internal::Env::get_max_expiration_ledger(&self.env)
```
dmkozh, line 145: I've just realized this doesn't work well with the bump interface, which operates on 'relative' ledger counts, not the absolute ledger seq. I think we should at least compensate for that in the SDK (subtract current ledger seq), but also maybe change the host fn semantics to not do unnecessary extra work.

leighmcculloch: Ah, it returns the max actual seq, not the max count. Thanks for catching that.

leighmcculloch: Fixed in b45806e.

soroban-sdk/src/ledger.rs
```diff
@@ -65,6 +65,14 @@ impl Ledger {
             .into()
     }
 
+    /// Returns the maximum ledger sequence number that data can live to.
+    #[doc(hidden)]
+    pub fn max_expiration_sequence(&self) -> u32 {
```
dmkozh, line 70: This is the same as `sequence`, is that intentional?

soroban-sdk/src/storage.rs
```diff
@@ -139,6 +139,15 @@ impl Storage {
         }
     }
 
+    /// Returns the maximum number of ledgers that an entry can live.
+    pub fn max_expiration(&self) -> u32 {
+        let seq = self.env.ledger().sequence();
+        let max: u32 = internal::Env::get_max_expiration_ledger(&self.env)
+            .unwrap_infallible()
+            .into();
+        max - seq
```
dmkozh, line 148: Need to add `+ 1` due to how math works for this (max entry lifetime includes the current ledger)

soroban-sdk/src/ledger.rs
leighmcculloch: It is not intentional. Heats really getting to me today. Thanks for catching this. Fixed in 7261049.

soroban-sdk/src/storage.rs
leighmcculloch: Fixed in d42a312.

dmkozh (approved):

PR #1110, opened 2023-10-07
Fix type of signature in auth

leighmcculloch:
### What
Change the type of the signature in custom auth from Vec<Val> to Val.

### Why
The protocol and env don't actually require it to be a Vec anymore, and so the SDK is limiting the type unnecessarily.


dmkozh (approved):
Sorry, lost track of this.

leighmcculloch:
Oh thanks. I had forgotten about this too!

Issue #1101, opened 2023-10-03
`Env::from_snapshot` no longer creates an `Env` consistent with `Env::default`


leighmcculloch:
### What version are you using?

`v20.0.0-rc2`

### What did you do?

Serialized an `Env` to a `LedgerSnapshot`, then reloaded it into an `Env`.

### What did you expect to see?

The `Env` loaded with the same state as if `Env::default()` had been called and the ledger state loaded, meaning auth recording is off, and a default base prng seed is set.

### What did you see instead?

An `Env` loaded with auth recording on and no default base prng seed.

### Discussion

It looks like when we added new behavior, and removed auth recording, in `Env::default`:
https://github.com/stellar/rs-soroban-sdk/blob/0992413f9b05e5bfb1f872bce99e89d9129b2e61/soroban-sdk/src/env.rs#L524-L532

We forgot to make the same changes in `Env::from_snapshot`:
https://github.com/stellar/rs-soroban-sdk/blob/0992413f9b05e5bfb1f872bce99e89d9129b2e61/soroban-sdk/src/env.rs#L1216-L1231

We should probably have `Env::from_snapshot` and `Env::default` share the same setup logic and the same storage type, where default is just an empty `LedgerSnapshot`.

leighmcculloch:
@brson You're welcome to take this, especially if it's blocking your work.

brson:
Yep. I will try to fix it.

PR #1111, opened 2023-10-09
fix(snapshot): set entry expiration info in `set_ledger_info`

ElliotFriend:
I've noticed some inconsistencies with snapshot creation and use where the various `*_entry_expiration` values are not set properly when create an `Env` from a snapshot that did have those values set.

This could possibly be related to #1101, but I'm not sure, since Leigh seemed more focused on prng.

Refs: #1101

### What

This commit updates the ledger-snapshot crate to also set the various ledger entry expiration values, if they are set on the snapshot being set in the environment.

### Why

When trying to load a ledger snapshot that I've created from an `Env` with valid entry expiration values, the new Env uses the defaults (which turn out to be `0`) for those expiration values.

I noticed trouble trying to run `env.register_contract()` or `env.register_stellar_asset_contract()` in a test file after loading a previously valid snapshot, and receive the error: `trying to bump past max expiration ledger`. (That expiration ledger, being `0` ledgers from now.)

### Known limitations

It's possible this is deliberate, and I'm misunderstanding how to use snapshots. If that's the case, we should probably bolster documentation around ledger snapshots to communicate that.

### DIFF
--- a/soroban-ledger-snapshot/src/lib.rs
+++ b/soroban-ledger-snapshot/src/lib.rs
@@ -85,6 +85,9 @@ impl LedgerSnapshot {
         self.timestamp = info.timestamp;
         self.network_id = info.network_id;
         self.base_reserve = info.base_reserve;
+        self.min_persistent_entry_expiration = info.min_persistent_entry_expiration;
+        self.min_temp_entry_expiration = info.min_temp_entry_expiration;
+        self.max_entry_expiration = info.max_entry_expiration;
     }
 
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

leighmcculloch:
@ElliotFriend heads up there's another related issue to restoring from snapshots that I'm fixing:
- https://github.com/stellar/rs-soroban-sdk/issues/1101

PR #1120, opened 2023-10-24
Use deterministic randomness in arbitrary tests

brson:
### What

Use a seeded StdRng instead of thread_rng in arbitrary tests.

### Why

Per https://github.com/stellar/rs-soroban-env/issues/810 randomized test cases should be deterministic.

### Known limitations

This is different from the approach in soroban-env-host from https://github.com/stellar/rs-soroban-env/pull/1124, where a test prng is attached to the env / host.

There are two more uses of thread_rng in this crate, but I don't know enough about that code to know how they should be removed, i.e. if they should follow a pattern similar to soroban-env-host. I am willing to replace all uses of thread_rng if given guidance about the preferred way to do it.

The implementation of StdRng is allowed to change in the future, so it's possible the exact numbers generated could change. I could import rand_chacha instead if desired.

### DIFF
--- a/soroban-sdk/src/arbitrary.rs
+++ b/soroban-sdk/src/arbitrary.rs
@@ -975,7 +975,7 @@ mod tests {
     };
     use crate::{Env, IntoVal};
     use arbitrary::{Arbitrary, Unstructured};
-    use rand::RngCore;
+    use rand::{RngCore, SeedableRng};
 
     fn run_test<T>()
     where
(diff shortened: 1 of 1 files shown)


brson:
The CI failure is strange.

It looks like a change in rust nightly linting behavior, and can be reproduced with

```
cargo +nightly check -p soroban-sdk --features testutils
```

leighmcculloch (approved):
Thanks!

leighmcculloch:
Thanks for sharing the command to minimally reproduce it. That helped. I've opened this to fix the lint warnings in nightly:
- https://github.com/stellar/rs-soroban-sdk/pull/1121

PR #1121, opened 2023-10-24
Fix nightly lints on unused internal exports

leighmcculloch:
### What
Fix nightly lints on unused internal exports

### Why
https://github.com/stellar/rs-soroban-sdk/pull/1120#issuecomment-1777913335

cc @brson

### DIFF
--- a/soroban-sdk/src/testutils.rs
+++ b/soroban-sdk/src/testutils.rs
@@ -21,7 +21,7 @@ pub trait ContractFunctionSet {
 }
 
 #[doc(inline)]
-pub use crate::env::testutils::*;
+pub use crate::env::internal::LedgerInfo;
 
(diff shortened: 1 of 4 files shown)


soroban-sdk/src/testutils.rs
```diff
@@ -21,7 +21,7 @@ pub trait ContractFunctionSet {
 }
 
 #[doc(inline)]
-pub use crate::env::internal::LedgerInfo;
+pub use crate::env::testutils::*;
```
leighmcculloch, line 24: I think this was the original intent for the env::testutils mod.

leighmcculloch: Actually, abandoning that env::testutils thing. Both LedgerInfo and Budget are exported in different ways in this mod and its better.

sisuresh (approved):

PR #1122, opened 2023-10-24
Generate all types of addresses when fuzzing.

### LINKED ISSUE #1096: impl SorobanArbitrary for Address always generates contract addresses
When fuzzing, arbitrary Addresses are always ScAddress::Contract types, not Account types.

This is the implementation of SorobanArbitrary:

```rust
    impl SorobanArbitrary for Address {
        type Prototype = ArbitraryAddress;
    }

    impl TryFromVal<Env, ArbitraryAddress> for Address {
        type Error = ConversionError;
        fn try_from_val(env: &Env, v: &ArbitraryAddress) -> Result<Self, Self::Error> {
            use crate::env::xdr::{Hash, ScAddress};

            let sc_addr = ScVal::Address(ScAddress::Contract(Hash(v.inner)));
            Ok(sc_addr.into_val(env))
        }
    }
```

I don't understand the consequences of using only contract addresses in tests, but this seems wrong.

We should probably generate both types, and it might be desirable to ask for only an account address or only a contract address.

brson:
### What

Generate both ScAddress::Contract and ScAddress::Account when fuzzing.

### Why

Currently only ScAddress::Contract is generated.

Closes #1096

### Known limitations

I don't really understand the consequences for fuzz testers of making this change. Maybe there aren't any. I have tested that the fuzzer in soroban-examples continues to work.

### DIFF
--- a/soroban-sdk/src/arbitrary.rs
+++ b/soroban-sdk/src/arbitrary.rs
@@ -532,7 +532,7 @@ mod objects {
 
     #[derive(Arbitrary, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
     pub struct ArbitraryAddress {
-        inner: [u8; 32],
+        inner: crate::env::xdr::ScAddress,
     }
 
     impl SorobanArbitrary for Address {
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
I wonder, there could be cases where someone specifically wants arbitrary contract addresses, but I think they can probably produce that themselves using the Address::random fn that just happens to only provide C addresses.

It's important the arbitrary on Address does both because contracts need to be able to handle both and this could be a good thing to catch.

PR #1133, opened 2023-11-06
Make Env::default and Env::from_snapshot configure the environment the same

### LINKED ISSUE #1101: `Env::from_snapshot` no longer creates an `Env` consistent with `Env::default`

brson:
### What

Make Env::default (when cfg testutils) and Env::from_snapshot configure the environment with a source account, a prng seed, and the debug diagnostics level. Do not turn on recording auth.

Fixes #1101

### Why

Users writing tests will expect a test environment to be configured similarly regardless of how it was constructed. At present these two constructors create differently-configured environments.

### Known limitations

na

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -459,7 +459,27 @@ impl Env {
         }
 
         let rf = Rc::new(EmptySnapshotSource());
-        let storage = internal::storage::Storage::with_recording_footprint(rf);
+        let info = internal::LedgerInfo {
+            protocol_version: 20,
+            sequence_number: 0,
+            timestamp: 0,
+            network_id: [0; 32],
+            base_reserve: 0,
+            min_persistent_entry_ttl: 4096,
+            min_temp_entry_ttl: 16,
+            max_entry_ttl: 6_312_000,
+        };
+
+        Env::new_for_testutils(rf, None, info)
+    }
+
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):
Thanks for fixing this 🙏🏻 .

I pushed a tweak to the assertions to understand what the error is in this case and that the expected error is indeed what's being triggered and not some other error by happenstance.

Issue #1082, opened 2023-09-07
Add test utilities that make it easier to test contract consistency over time


leighmcculloch:
It would benefit contract developers for the SDK to contain some really easy to use tools for testing contract consistency over time. By really easy to use, I mean they are automatically enabled by default, or it's possible to enable them or use the feature with a single flag.

The SDK provides a reasonably good framework for testing contracts and allows developers to write assertions on return values, emitted events, final ledger state. And it allows the developer to write these assertions and exercise multiple contracts at once, providing a complete integration test experience in the simplicity of unit tests. The framework also provides some convenience functions for saving the ledger state to file.

Developers can use the above to check that over time a contracts behavior doesn't change at all, or if it does, it is trivial to confirm if the change is intended by viewing diffs of ledger state and events.

However, for developers to get that type of consistency check, they need to write a few lines of code into ever test they write. This is a reasonably high bar.

We should investigate and find a way to have every test automatically write out to disk:
1. The ledger state at the end of the test.
2. The events published during the test.
3. A summary of all invocations and their return values.

If this data is written to disk to a common folder developers can choose to commit it to VCS (e.g. git), or choose to ignore it by adding the folder to their `.gitignore` or similar file. If committed developers will naturally see changes over time when upgrading envs, upgrading sdks, upgrading dependencies (code or contract), and making changes to their own contract.

Given that consistency is often paramount, and surprise inconsistency can lead to bugs and exploits, this seems worth putting some effort into even if the feature is not trivial to implement.

Related to https://github.com/stellar/rs-soroban-sdk/issues/1095.

leighmcculloch:
Once this is implemented, or if this is abandoned, we should document how to test contracts for consistency:
- https://github.com/stellar/soroban-docs/issues/561

PR #1135, opened 2023-11-07
Remove use of rand from generated addresses, nonces, salts, and issuer pks

leighmcculloch:
### What
Remove use of rand from generated addresses, nonces, salts, and issuer pks, replace with sequential values instead.

### Why
The SDK generates random addresses, nonces, salts, and issuer pks. This randomness makes it near impossible to write any test that produces predictable and consistent total system state.

The SDK doesn't need to be using random values for these things. It just needs to produce values that don't matter and that are unique.

Predictable and consistent values make it easier to write tests that can be used over time to assess the consistency of a contract. This is an important attribute of tests so that folks can evaluate if their contract is consistent even with dependency upgrades, SDK upgrades, and protocol upgrades.

This change also introduces the concept of a `Snapshot`, which is a superset of the existing `LedgerSnapshot`. The `LedgerSnapshot` contains everything needed to load in a ledger from a network or other source. `Snapshot` contains everything needed to rehydrate the state of an SDK testutils backed Env.

Related discussion in Discord: https://discord.com/channels/897514728459468821/1171226885171191949

For #1082

### DIFF
--- a/soroban-sdk/src/tests/env.rs
+++ b/soroban-sdk/src/tests/env.rs
@@ -72,8 +72,8 @@ fn default_and_from_snapshot_same_settings() {
 
     let logs1 = env1.logs().all();
     let logs2 = env2.logs().all();
-    assert!(!logs1.is_empty());
-    assert!(!logs2.is_empty());
+    assert_eq!(logs1, &["[Diagnostic Event] contract:0000000000000000000000000000000000000000000000000000000000000001, topics:[log], data:\"test\""]);
+    assert_eq!(logs2, &["[Diagnostic Event] contract:0000000000000000000000000000000000000000000000000000000000000001, topics:[log], data:\"test\""]);
 }
 
 #[test]
--- a/soroban-sdk/src/tests/env.rs
+++ b/soroban-sdk/src/tests/env.rs
@@ -75,3 +75,32 @@ fn default_and_from_snapshot_same_settings() {
     assert!(!logs1.is_empty());
     assert!(!logs2.is_empty());
 }
+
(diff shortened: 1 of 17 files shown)


soroban-sdk/src/tests/env.rs
```diff
@@ -75,3 +75,32 @@ fn default_and_from_snapshot_same_settings() {
...
+    let env2addr3 = env2.register_contract(None, Contract);
+    println!("env2 addr3 {:?}", env2addr3.contract_id());
+    let env3addr3 = env3.register_contract(None, Contract);
+    println!("env3 addr3 {:?}", env3addr3.contract_id());
+
+    // Check that contracts deployed in the envs are consistent and predictable.
+    assert_eq!(env2addr1.contract_id(), env1addr1.contract_id());
+    assert_eq!(env2addr2.contract_id(), env1addr2.contract_id());
+    assert_eq!(env2addr3.contract_id(), env1addr3.contract_id());
+    assert_eq!(env3addr3.contract_id(), env1addr3.contract_id());
```
leighmcculloch, line 105: Using sequential IDs for generated addresses and nonces works for the most part, until we get to env snapshotting. The generator state is not included in the snapshot. It can be added, but it also seems a bit odd given that the snapshots currently don't capture any test-only data.

leighmcculloch: I think what I'll do is create a TestSnapshot that wraps the LedgerSnapshot. The TestSnapshot will contain additional fields.

dmkozh (approved):

PR #1137, opened 2023-11-08
Autosave a test snapshot file on every test exit

### LINKED ISSUE #1082: Add test utilities that make it easier to test contract consistency over time

leighmcculloch:
### What
Autosave a test snapshot file on every test exit.

### Why
To encourage the over-time consistency testing of contracts. To provide a mechanism where any developer who commits the generated files can see changes in observable behavior that arises unexpectedly over time.

Close #1082

### Merging
To be merged to `main` after:

- [x] Adding events to the `Snapshot`.
- [ ] ~Adding budget to the `Snapshot`. _(Maybe, seems less clear if should be included.)_~
- [x] Support multiple Envs in a single test writing multiple files.
- [x] https://github.com/stellar/rs-soroban-sdk/pull/1142
- [x] https://github.com/stellar/rs-soroban-env/pull/1182
- [x] #1135


leighmcculloch:
Worth calling out there's no way to turn the generation of test snapshots off. I think there's a lot of value of it always being on. If we were going to add a switch for it, it'd be preferable I think to make it opt out rather than opt in.

Although, maybe it could slow down fuzzer tests so much that it'd be worth adding a flag for.

soroban-sdk/src/env.rs
```diff
@@ -1273,6 +1287,85 @@ impl Env {
...
+    /// as stored entries.
+    ///
+    /// ### Panics
+    ///
+    /// If there is any error writing the file.
+    pub(crate) fn to_test_snapshot_file(&self) {
+        let test = std::thread::current();
+        let test_name = test
+            .name()
+            .expect("test name to be retrieval for use as the name of the test snapshot file");
```
dmkozh, line 1338: to be retrieved?

dmkozh (approved):

soroban-sdk/src/env.rs
leighmcculloch:
```suggestion
            .expect("test name to be retrieved for use as the name of the test snapshot file");
```

PR #1140, opened 2023-11-09
Update soroban-env-* deps

leighmcculloch:
### What
Update soroban-env-* deps.

### Why
Keep the SDK updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1103,7 +1103,7 @@ checksum = "a507befe795404456341dfab10cef66ead4c041f62b8b11bbb92bffe5d0953e0"
 [[package]]
 name = "soroban-builtin-sdk-macros"
 version = "20.0.0-rc2"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=2674d867d7c6aa4212abab05ff30e5804ff1db90#2674d867d7c6aa4212abab05ff30e5804ff1db90"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9#6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9"
 dependencies = [
  "itertools",
  "proc-macro2",
@@ -1114,7 +1114,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-common"
 version = "20.0.0-rc2"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=2674d867d7c6aa4212abab05ff30e5804ff1db90#2674d867d7c6aa4212abab05ff30e5804ff1db90"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9#6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9"
 dependencies = [
  "arbitrary",
  "crate-git-revision",
@@ -1131,7 +1131,7 @@ dependencies = [
 [[package]]
 name = "soroban-env-guest"
 version = "20.0.0-rc2"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=2674d867d7c6aa4212abab05ff30e5804ff1db90#2674d867d7c6aa4212abab05ff30e5804ff1db90"
+source = "git+https://github.com/stellar/rs-soroban-env?rev=6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9#6d26ca68170ecf9d19b8c55bfcc157a1ad62b7e9"
 dependencies = [
  "soroban-env-common",
  "static_assertions",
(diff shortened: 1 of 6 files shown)


leighmcculloch:
I'm seeing the following failures relating to the arbitrary tests, but I'm not sure what these errors mean yet:

```
failures:

---- tests::proptest_val_cmp::test stdout ----
thread 'tests::proptest_val_cmp::test' panicked at soroban-sdk/src/tests/proptest_val_cmp.rs:14:1:
Test aborted: initial arbitrary call failed
        successes: 61
        local rejects: 0
        global rejects: 0


---- testutils::arbitrary::tests::test_map_error stdout ----
thread 'testutils::arbitrary::tests::test_map_error' panicked at soroban-sdk/src/testutils/arbitrary.rs:1136:68:
SorobanArbitrary: IncorrectFormat

---- testutils::arbitrary::tests::test_val stdout ----
thread 'testutils::arbitrary::tests::test_val' panicked at soroban-sdk/src/testutils/arbitrary.rs:1136:68:
SorobanArbitrary: IncorrectFormat

---- testutils::arbitrary::tests::test_vec_error stdout ----
thread 'testutils::arbitrary::tests::test_vec_error' panicked at soroban-sdk/src/testutils/arbitrary.rs:1136:68:
SorobanArbitrary: IncorrectFormat


failures:
    tests::proptest_val_cmp::test
    testutils::arbitrary::tests::test_map_error
    testutils::arbitrary::tests::test_val
    testutils::arbitrary::tests::test_vec_error

test result: FAILED. 273 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.78s
```

@brson Do you know from this output where a good place to start would be?

brson:
@leighmcculloch It must be related to this: https://github.com/stellar/rs-soroban-env/pull/1171/files#diff-d5fa2fac383f5ecf1e39b1233282102eb52ff9b9a2d110e3516491f475cd4831R18

leighmcculloch:
Agreed. Should I change the SDK arbitrary run_test fn to sit in a loop until it gets Ok(_) to make it tolerant of invalid arbitrary values? It looks like the Symbol type in the env-common crate was also changed to return the same error sometimes.

brson:
The impl of Arbitrary for Error in soroban-env-host looks like this:

```
impl<'a> Arbitrary<'a> for Error {
    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
        let scerror = ScError::arbitrary(u)?;
        let error = Error::from(scerror);
        // FIXME: fuzzer discovered that it can just return "InternalError" from
        // a contract to make the host think it had an InternalError. See
        // https://github.com/stellar/rs-soroban-env/issues/1175
        if error.is_code(ScErrorCode::InternalError) {
            Err(arbitrary::Error::IncorrectFormat)
        } else {
            Ok(error)
        }
    }
}
```

The tests currently don't expect any Arbitrary implementation to return an error if given enough bytes; and  since proptest_val_cmp is failing, I am imagine that the proptest_arbitrary_interop crate isn't expecting the `arbitrary` method to fail either.

Since it is valid for `arbitrary` to fail, the arbitrary test suite can probably account for this in the `run_test` method used by all the tests. The propetest probably needs to be fixed in the proptest_arbitrary_interop crate.

I'll try to hack at it right now but maybe resolve it tonight.

brson:
> Should I change the SDK arbitrary run_test fn to sit in a loop until it gets Ok(_) to make it tolerant of invalid arbitrary values?

Yes, I think so. You can go ahead since you are already hacking on it. You might just need to `#[ignore]` the proptest for now though.

leighmcculloch:
+1 I'm going to `#[ignore]` the proptest tests, as I don't see a way to handle fallible arbitrary without a modification to [`proptest_arbitrary_interop`](https://crates.io/crates/proptest-arbitrary-interop), which is an external crate owned by @graydon. @graydon do you see a way to support fallible arbitrary impl in proptest?

brson:
I'm willing to hack on proptest_arbitrary_interop to fix this in a followup, as I want to do some hacking on that crate for other reasons.

brson:
I guess I should say that I question whether

```rust
        if error.is_code(ScErrorCode::InternalError) {
            Err(arbitrary::Error::IncorrectFormat)
        } else {
            Ok(error)
        }
```

is correct - since it is possible for contracts to see this error code, it should be generated for fuzzing. But that's not a question for this pr.

leighmcculloch:
> I'm willing to hack on proptest_arbitrary_interop to fix this in a followup, as I want to do some hacking on that crate for other reasons.

That'd be awesome, thank you.

I've created an issue to track this and added it as a todo in the code on the ignore:
- https://github.com/stellar/rs-soroban-sdk/issues/1141

dmkozh (approved):

leighmcculloch:
> I guess I should say that I question whether...is correct

Agreed. There's an issue tracking that code here, and maybe worth posting some thoughts on that issue if you have ideas on what should be there instead:
- https://github.com/stellar/rs-soroban-env/issues/1175

PR #1143, opened 2023-11-10
Change testutils::Events::all to use a std Vec

leighmcculloch:
### What
Change testutils::Events::all to use a std Vec instead of an sdk Vec.

### Why
There's no need to use the sdk Vec for this type, and in other similar testutils, such as testutils::Logs::all we use the std Vec.

The main reason to use an sdk type is if folks are going to have an easier time using it. So we should use sdk types for any Vals such as Addresses. But there's little reason to use the sdk Vec over std Vec for test data such as this, and it will be easier for a developer to interrogate and assert on.


leighmcculloch:
I can't go ahead with this because `Val`s aren't comparable.

PR #1154, opened 2023-11-14
Add footprint to Env snapshot

### LINKED ISSUE #1152: Include auth state in Env test snapshots
Idea from @dmkozh here:
- https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1699985293524799?thread_ts=1699983645.263049&cid=C030Z9EHVQE

leighmcculloch:
### What
Add footprint to Env snapshot

### Why
Close #1152

### Merging

Intended to be merged to `main` after:
- #1153


leighmcculloch:
It's quickly obvious when testing this that it adds little to no value. In tests the entire footprint is largely writes to all keys. It really doesn't capture much of importance since we don't test footprints or provide capabilities for setting up read only footprints.

PR #1155, opened 2023-11-15
Don't write test snapshot if no thread name

brson:
### What

When dropping `Env`, if there is no thread name, don't write a snapshot file (and don't panic).

### Why

The current code for writing test snapshots under `testutils` assumes the `Env` has been constructed as part of a unit test, but this is not the case for e.g. fuzz tests.

A fuzz test runs on the main thread and does not configure a thread name, but the test snapshot writer panics if there is no thread name.

This patch adds a (very crude) heuristic to decide that it is not running as part of a unit test and it is ok to not write the test snapshot. I looked for more reliable ways to determine if code was running as part of a unit test but did not find one.

### Known limitations

This heuristic may be inaccurate for other unanticipated scenarios.

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -1332,10 +1332,13 @@ impl Env {
     ///
     /// If there is any error writing the file.
     pub(crate) fn to_test_snapshot_file(&self) {
-        let test = std::thread::current();
-        let test_name = test
-            .name()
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):

leighmcculloch:
Thanks for making this change. This is better. 👍

PR #1158, opened 2023-11-15
Add auth to Env snapshot

### LINKED ISSUE #1152: Include auth state in Env test snapshots

leighmcculloch:
### What
Include auth details in the Env snapshot that is persisted to file at the end of tests.

### Why
The details about what auth is performed during a test run is not currently captured in the Env snapshot, but it is something that is an important behavior of a test that should be tracked to test for consistency over time.

Close #1152 

Dependent on:
- https://github.com/stellar/rs-soroban-env/pull/1214
- https://github.com/stellar/rs-soroban-sdk/pull/1157
- https://github.com/stellar/rs-soroban-sdk/pull/1153

### DIFF
--- a/soroban-sdk/test_snapshots/tests/auth/auth_15_one_repeat/test.1.json
+++ b/soroban-sdk/test_snapshots/tests/auth/auth_15_one_repeat/test.1.json
@@ -3,6 +3,54 @@
     "address": 2,
     "nonce": 2
   },
+  "auth": [
+    [
+      "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFCT4",
+      {
+        "function": {
+          "contract_fn": {
+            "contract_address": "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM",
+            "function_name": "add",
+            "args": [
+              {
+                "address": "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFCT4"
+              },
+              {
+                "i32": 10
+              },
+              {
+                "i32": 12
+              }
+            ]
+          }
+        },
+        "sub_invocations": []
+      }
+    ],
+    [
+      "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFCT4",
+      {
+        "function": {
+          "contract_fn": {
+            "contract_address": "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM",
+            "function_name": "add",
+            "args": [
+              {
+                "address": "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFCT4"
+              },
+              {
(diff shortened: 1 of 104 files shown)


sisuresh (approved):

soroban-sdk/test_snapshots/tests/auth/auth_15_one_repeat/test.1.json
```diff
@@ -3,6 +3,54 @@
     "address": 2,
     "nonce": 2
   },
+  "auth": [
+    [
+      "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFCT4",
```
dmkozh, line 8: I think it's better than nothing, but OTOH it could be misleading because it's just the last auth. We probably could allow adding a hook for popping the last frame (== finishing a contract call) and then you'd be able to append auths of this particular call to some 'global' auth array. That doesn't need to block this PR though.

dmkozh (approved):

soroban-sdk/test_snapshots/tests/auth/auth_15_one_repeat/test.1.json
leighmcculloch: Hmm, this isn't the auth for all auths that occurred? Is there anyway to get all auths that have occurred?

dmkozh: No, we reset auth manager on every host call. This seemed most useful for testing because you don't really want to keep re-asserting on the old entries. We could either additionally require all auths, or add a hook to handle this in more generic fashion (maybe you'd want to report something else per contract call).

leighmcculloch: > on every host call

Do you mean on every top-level `invoke`/`try_invoke` call?

leighmcculloch: I could try persisting the results of calling .auth() in the SDK's `invoke`/`try_invoke` fns. But I don't think the SDK will be able to tell the difference between top-level and nested invoke calls.

dmkozh: That's why I'm suggesting to introduce a generic hook in env that would allow executing arbitrary code when we finish a top-level contract invocation.

leighmcculloch: A simple thing to add might be a function `is_in_frame()` or similar that returns true if we're currently inside an invocation. That would be useful in at least one other location in the SDK. Will DM you about how I can add this.

leighmcculloch: @dmkozh and I discussed offline and he highlighted that using a hook would be less error prone for the auth case as there's a few situations where auths can occur and we shoold capture the auths, like when the SDK as_contract fn gets called. We decided to go with the plan outlined in these issues:

- https://github.com/stellar/rs-soroban-sdk/issues/1164

And we'd be able to use this functionality to do some other interesting things:
- https://github.com/stellar/rs-soroban-sdk/issues/1165

PR #1163, opened 2023-11-15
Improvements to Arbitrary Options

brson:
### What

Simplify generic type bounds related to `ArbitraryOption`. Add tests. Add `Option` to `ArbitraryVal`.

### Why

The bounds on this type are more complex than other containers like `Vec<T>`. The reason is that the `TryFromVal` implementation assumes that the conversions it makes are fallible and propagates errors that necessitate complex bounds related to the error types.

These conversions are infallible though and don't need to worry about errors.

The third commit makes the impl of `Arbitrary` for `Val` generate `Option`s.

### Known limitations

na

---

I also slipped in a minor fix for `ArbitraryValVec::Map`, where it was generating the wrong type.

### DIFF
--- a/soroban-sdk/src/testutils/arbitrary.rs
+++ b/soroban-sdk/src/testutils/arbitrary.rs
@@ -869,7 +869,7 @@ mod composite {
         String(<Vec<String> as SorobanArbitrary>::Prototype),
         Symbol(<Vec<Symbol> as SorobanArbitrary>::Prototype),
         Vec(<Vec<Vec<u32>> as SorobanArbitrary>::Prototype),
-        Map(<Map<u32, u32> as SorobanArbitrary>::Prototype),
+        Map(<Vec<Map<u32, u32>> as SorobanArbitrary>::Prototype),
         Address(<Vec<Address> as SorobanArbitrary>::Prototype),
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -876,7 +869,7 @@ mod composite {
         String(<Vec<String> as SorobanArbitrary>::Prototype),
         Symbol(<Vec<Symbol> as SorobanArbitrary>::Prototype),
         Vec(<Vec<Vec<u32>> as SorobanArbitrary>::Prototype),
-        Map(<Map<u32, u32> as SorobanArbitrary>::Prototype),
+        Map(<Vec<Map<u32, u32>> as SorobanArbitrary>::Prototype),
```
leighmcculloch, line 872: Why does the Map contain a Vec or Maps?

leighmcculloch (commented):
👏🏻 Thanks for simplifying this. One question inline.

leighmcculloch (approved):

soroban-sdk/src/testutils/arbitrary.rs
brson: fwiw i recognize this is a lot of code just to produce random `Val`s, and there might be more clever and thorough ways to do it

PR #1167, opened 2023-11-17
Update env and xdr

leighmcculloch:
### What
Update env and xdr.

### Why
Keep it updated.

PR #1168, opened 2023-11-18
Snapshot all auths during a test not just last invokes

### LINKED ISSUE #1164: Integrate top-level call enter/exit hooks into generation of test snapshots
Follow up to:
- https://github.com/stellar/rs-soroban-sdk/pull/1152
- https://github.com/stellar/rs-soroban-sdk/pull/1158

Dependent on:

Once the environment has enter/exit hooks for top-level call (invoke), the SDK environment can use the hooks to make the auths included in test snapshots more complete. Right now the test snapshots only include the auths from the last top level call (invoke). The SDK will:

on exit hook:
```
|| {
    env.auths += host.auths()
}
```
then use it in test snapshots:
```
test_snapshot.auths = env.auths
```

leighmcculloch:
### What
Snapshot all auths during a test not just last invokes.

### Why
The test snapshot logic is using the authorization manager of the host to get auths that have occurred. However, the authorization manager only has the auths from the last invoke. To get all auths for the Envs lifetime the contract invocation hook needs to be used to tell us when each call to the host finishes so that the auths can be collected as the test executes.

Close #1164

### Merging
Intended for merging to `main` after:
- #1167

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -182,9 +182,9 @@ impl MaybeEnv {
             #[cfg(any(test, feature = "testutils"))]
             generators: None,
             #[cfg(any(test, feature = "testutils"))]
-            snapshot: None,
-            #[cfg(any(test, feature = "testutils"))]
             auth_snapshot: None,
+            #[cfg(any(test, feature = "testutils"))]
+            snapshot: None,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


soroban-sdk/src/env.rs
```diff
@@ -515,10 +527,27 @@ impl Env {
...
+
+        let auth_snapshot = Rc::new(RefCell::new(AuthSnapshot::default()));
+        let auth_snapshot_in_hook = auth_snapshot.clone();
+        env_impl
+            .set_top_contract_invocation_hook(Some(Rc::new(move |host, event| {
+                if let ContractInvocationEvent::Finish = event {
+                    let new_auths = host
+                        .get_authenticated_authorizations()
+                        // If an error occurs getting the authenticated authorizations
+                        // it means that no auth has occurred.
```
dmkozh, line 539: I don't think error can ever occur here, because by definition we call this hook after invoking a contract and that guarantees that auth is present (even if empty). So if it occurs, then we likely have a bug in host.

dmkozh (approved):

PR #1174, opened 2023-11-21
Restore previous auth state in client

### LINKED ISSUE #1084: Restore previous auth state after client modifies it
The generated client of contracts sets the auth state just before making a call based on what configuration the auth state has in the client. Today as of _ it doesn't return the state.

This results in a side-effect taking place. If the env was configured differently before the client is used, it'll be left in a surprising unexpected state.

There's a todo in the code about this:
https://github.com/stellar/rs-soroban-sdk/blob/ee44e420b9badeb16cab870ed25bef397c058f64/soroban-sdk-macros/src/derive_client.rs#L214-L234

The stellar/rs-soroban-env#785 issue that the todo is dependent on has been implemented it, and so this can happen now.

leighmcculloch:
### What
Restore previous auth state in client after client finishes invoke.

### Why
The client is presented in such a way that it looks like invocations with a client that has been configured with certain auth will apply the auth only during invocations with it.

It's unintuitive that the auths from the client apply forever after the invocation.

Close #1084

Dependent on:
- https://github.com/stellar/rs-soroban-env/pull/1244


soroban-sdk-macros/src/derive_client.rs
```diff
@@ -181,9 +181,10 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
             quote! {
                 #(#fn_attrs)*
                 pub fn #fn_ident(&self, #(#fn_input_types),*) -> #fn_output {
-                    // TODO: Undo the mock and restore previous auth state after
-                    // https://github.com/stellar/rs-soroban-env/issues/785 is
-                    // implemented.
+                    #[cfg(any(test, feature = "testutils"))]
+                    let old_auth_manager = (*self.env.in_contract).borrow().then(||
```
leighmcculloch, line 185: Wait, this is wrong, this should be occuring only if not in a contract.

soroban-sdk-macros/src/derive_client.rs
```diff
@@ -181,9 +181,10 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
...
+                        self.env.host().snapshot_auth_manager().unwrap()
```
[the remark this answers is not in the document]
leighmcculloch: Agreed, this is a mistake, the `in_contract` check is supposed to be only executing the the snapshot if we're _not_ in a contract, and I messed it up. Fixed that in 4048884, but unfortunately everything is broken now, so I must have something else going on I don't see yet.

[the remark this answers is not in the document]
leighmcculloch: Fwiw people can't use the other functions of the contract client in the wasm build. They can try, but it won't build.

soroban-sdk/src/env.rs
```diff
@@ -532,19 +547,23 @@ impl Env {
...
-                        // If an error occurs getting the authenticated authorizations
-                        // it means that no auth has occurred.
-                        .unwrap_or_default();
-                    (*auth_snapshot_in_hook).borrow_mut().0.extend(new_auths);
+                match event {
+                    ContractInvocationEvent::Start => {}
+                    ContractInvocationEvent::Finish => {
+                        let new_auths = host
+                            .get_authenticated_authorizations()
+                            // If an error occurs getting the authenticated authorizations
```
dmkozh, line 555: This comment is redundant now (and it's also incorrect; this simply shouldn't ever happen unless hook is bugged).

dmkozh (approved):

soroban-sdk/src/env.rs
leighmcculloch: I'll replace it with a simple unwrap.

leighmcculloch:
This change at commit 49d1ba0 worked, but there were some edge cases where technically a developer could end up in a odd place by instantiating an Env, that would think it is outside a contract when it was inside. So I've opened the following change on env and updated this change to use it:
- https://github.com/stellar/rs-soroban-env/pull/1244

PR #1175, opened 2023-11-22
Make try_ fns return smaller SDK error type

### LINKED ISSUE #1136: Try_ fns return host Errors but should return smaller subset
@dmkozh pointed out on Discord that the only non-contract-error a contract will ever see in response to any host fn call, specifically the `try_invoke` and other `try_` client calls will be a Context InvalidAction error.

Ref: https://discord.com/channels/897514728459468821/966788672164855829/1171588706684117112

However, the SDK exposes the host Error type in these try_ functions and so makes it appear to a contract developer that more errors will occur. In this way the SDK API is telling a not true story to the developer about what is possible and what errors could be handled.

I think we should replace the use of the host Error type in the SDK with a subset Error type:
```rust
pub enum Error {
    ContextInvalidAction,
    ContractError(u32),
}
```

or:
```rust
pub enum Error {
    Abort,
    Contract(u32),
}
```

or something like that.

leighmcculloch:
### What
Make try_ fns return smaller SDK error type instead of the large host error type.

### Why
The host error type contains a massive number of possible errors. However, the host obfuscates the error when it is passed to the guest, and so the guest and sdk only ever see the Context/InvalidAction error.

It's confusing for developers if we expose the host error type, then don't actually surface most of the errors on it. A developer might think they can distingish between different error states, when in reality they cannot.

Close #1136


soroban-sdk/src/error.rs
```diff
@@ -0,0 +1,33 @@
+use core::convert::Infallible;
+
+use crate::xdr;
+
+/// InvokeError captures errors returned from the invocation of another
+/// contract.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum InvokeError {
+    /// Abort occurs if the invoked contract function exited due to some
+    /// irrecoverable error, such as exceeding budget, or out-of-bounds access.
```
dmkozh, line 10: That's not quite correct. The real irrecoverable errors (such as exceeding budget) aren't observable by the guest, as they cause the whole invocation to fail. I don't think we need to document that though (i.e. I wouldn't mention exceeding budget at all). I think a fair description would be that this covers any type of trap in the executed contract, either induced by an explicit `panic!`, or a host function failure, or a runtime error (like out-of-bounds access mentioned here).

soroban-sdk/src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+    Abort,
+    /// Contract error occurs if the invoked contract function exited due to
```
dmkozh, line 12: I would probably mention that this is an error that occurs when the contract explicitly returns an error or calls `panic_with_error`. I realize it's not necessarily in the scope of this enum, but I feel like it would be useful for the reader to make this connection.

soroban-sdk/src/error.rs
leighmcculloch: Thanks you're right, this is entirely wrong. I'll change it.

soroban-sdk/src/error.rs
```diff
@@ -0,0 +1,39 @@
...
+use crate::xdr;
+
+/// InvokeError captures errors returned from the invocation of another
+/// contract.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum InvokeError {
+    /// Abort occurs if the invoke contract panicks with a [`panic!`], or a host
+    /// function of the environment has a failure, or a runtime error occurs.
+    /// either induced by an explicit panic!, or a host function failure, or a
+    /// runtime error (like out-of-bounds access mentioned here).
```
dmkozh, line 12: Seems like lines 3-4 of the comment should be removed.

dmkozh (approved):

Issue #1195, opened 2024-01-05
Add semver API checks to CI / release process


leighmcculloch:
Now that the SDK is released it will be easy for us to accidentally make breaking changes to the exported API of the SDK. Especially because the SDK reexports much of the API of the soroban-env-common crate.

We can try and do our best to identify breakages, but there are tools in the Rust ecosystem we should use as aids or as checks.

leighmcculloch:
Tools to eval:
- https://github.com/obi1kenobi/cargo-semver-checks
- https://github.com/Enselic/cargo-public-api

leighmcculloch:
`cargo-semver-checks` definitely seems like the simplest to plugin to a CI and use to identify errors. But `cargo-public-api` looks like a good tool for broadly viewing and diffing APIs, so probably worth having in our back pocket for debugging and understanding how the API has changed.

Likely next steps:
- Try integrating cargo-semver-checks into our CI.

PR #1201, opened 2024-01-08
Export functions for creating Timepoint/Duration

### LINKED ISSUE #1197: Timepoint and Duration types are difficult to construct in tests

`v20.0.3`

### What did you do?

Tried to create a `Timepoint` or `Duration` using functions on the tests.

Some way to do that.

### What did you see instead?

No way to do that.

Other than building the XDR and converting that way:
```rust
use soroban_sdk::{xdr, IntoVal, Timepoint};

let t: Timepoint = xdr::ScVal::Timepoint(xdr::TimePoint(1)).into_val(&e);
```

leighmcculloch:
### What
Export functions for creating and accessing Timepoint and Duration.

### Why
The types are not creatable inside a contract, and outside a contract in tests are only creatable using the XDR ScVal type. If developers are to use the Timepoint and Duration types, they need to be able to construct the types from primitive values.

The `Timepoint` `from_unix` and `to_unix` functions were added using the unix terminologgy because that's the type of time values the Stellar network uses.

The `Duration` `from_seconds` and `to_seconds` functions were added using the seconds terminology because it seemed like the most direct way to communicate intent. I originally intended to use `secs` which I think would also be reasonable and feedback on the issue suggested `seconds` which also seems reasonable to me.

While for both types functions were repurposed/removed, those functions were private only and should have no impact on compatibility.

Close #1197


soroban-sdk/src/env.rs
```diff
@@ -76,7 +76,6 @@ pub mod internal {
 }
 
 pub use internal::xdr;
-pub use internal::Compare;
```
leighmcculloch, line 79: Being removed by:
- #1199

graydon (approved):

leighmcculloch:
Holding until this other issue is closed:
- https://github.com/stellar/rs-soroban-sdk/issues/1195

PR #1202, opened 2024-01-08
Add CI job to run cargo-semver-checks

### LINKED ISSUE #1195: Add semver API checks to CI / release process

leighmcculloch:
### What
Add CI job to run cargo-semver-checks

### Why
To monitor whether the changes being made to the soroban-sdk and its exports are a match for the version being targeted.

Close #1195


leighmcculloch:
The semver check tooling is not going to detect breaking changes to types in the env packages that are reexported unfortunately. I tried a few different test cases, things like removing a trait impl such as the Default impl on Val, which doesn't appear to be used by anything but is visible in the soroban-sdk docs on the Val type, so should be considered part of the public API. And nothing in the env repo exported seems to get detected.

graydon (approved):

PR #1209, opened 2024-01-17
Add storage update fns

leighmcculloch:
### What
Add fns to the storage interface for updating (getting then setting) key-value pairs.

### Why
In contracts when I want to do a get and set, an interface like this where I only need to provide the key once, and select the storage type once, is less noisy.

Here's a small example from the `mint-lock` example in the soroban-examples repo. It's not really much less lines of code, but line count isn't the motivation here, rather grouping the "open" and "close" of the update.
```diff
-            let minter_stats = env
-                .storage()
-                .temporary()
-                .get::<_, MinterStats>(&minter_stats_key)
-                .unwrap_or_default();
-            let new_minter_stats = MinterStats {
-                consumed_limit: minter_stats.consumed_limit + amount,
-            };
-            if new_minter_stats.consumed_limit > config.limit {
-                return Err(Error::DailyLimitInsufficient);
-            }
-            env.storage()
-                .temporary()
-                .set::<_, MinterStats>(&minter_stats_key, &new_minter_stats);
+            env.storage().temporary().try_update(
+                &minter_stats_key,
+                |val: Option<MinterStats>| {
+                    let mut val = val.unwrap_or_default();
+                    val.consumed_limit += amount;
+                    if val.consumed_limit > config.limit {
+                        return Err(Error::DailyLimitInsufficient);
+                    }
+                    Ok(val)
+                },
+            )?;

```

### DIFF
--- a/soroban-sdk/src/storage.rs
+++ b/soroban-sdk/src/storage.rs
@@ -230,7 +230,8 @@ impl Storage {
     ///
     /// Loads the value, calls the function with it, then sets the value to the
     /// returned value of the function.  If no value is stored with the key then
-    /// the function is called with None.
+    /// the function is called with None.  If the function returns an error it
+    /// will be passed through.
     ///
     /// The returned value is the value stored after updating.
     pub(crate) fn try_update<K, V, E>(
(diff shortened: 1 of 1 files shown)


sisuresh (approved):

soroban-sdk/src/storage.rs
```diff
@@ -201,6 +201,57 @@ impl Storage {
             .unwrap_infallible();
     }
 
+    /// Update a value stored against a key.
+    ///
+    /// Loads the value, calls the function with it, then sets the value to the
+    /// returned value of the function.  If no value is stored with the key then
```
dmkozh, line 207: nit: Seems like you have double-spaces in a few places here.

leighmcculloch: Oops. Annoyingly for some reason when VSCode rewraps the lines it adds two spaces between sentences.

PR #1216, opened 2024-01-31
Bug 1076 conversion error flattening

graydon:
This is the SDK companion to https://github.com/stellar/rs-soroban-env/pull/1339 -- it won't work until that change has landed and its refs have updated to point to it.

There's an explanation of the change over there, but briefly: we were discarding meaningful errors when passing through the `TryFromVal<ScVal> for Val` and `TryFromVal<Val> for ScVal` impls in env-common, and now (or at least if 1339 lands) we're not. Instead of returning `ConversionError` we'd be returning `Error`.

The SDK happens to _also_ have conversions with `ConversionError` signatures, and I've opted in this PR to _not_ go a step further and actually change those to `Error`, instead re-burying the `Error` into `ConversionError` in the SDK, keeping the SDK APIs as they are. I could also change them. But I figured that could be left to @leighmcculloch's choice; he might want to keep the SDK's error repertoire simpler, I'm not sure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1095,8 +1095,8 @@ checksum = "4dccd0940a2dcdf68d092b8cbab7dc0ad8fa938bf95787e1b916b0e3d0e8e970"
 
 [[package]]
 name = "soroban-builtin-sdk-macros"
-version = "20.1.0"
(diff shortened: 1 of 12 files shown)


jayz22 (approved):

graydon:
See also https://github.com/stellar/rs-soroban-sdk/pull/1218 for alternative approach

sisuresh (approved):

PR #1218, opened 2024-01-31
Bug 1076 conversion error flattening 2

graydon:
This is an alternative approach to PR https://github.com/stellar/rs-soroban-sdk/pull/1216 where we plumb the `env::Error` code all the way up to the surface of the SDK's UI.

I don't feel strongly about doing this or the other one, they both seem to work, it's a question of UI complexity. Neither approach suffers the problem that motivated this change at the env level, where we went from `Error` to `ConversionError` and then _back_ to `Error` and thereby produced misleading `Error` codes.


graydon:
decided against this one for now.

PR #1231, opened 2024-02-14
Add Bytes to_buffer and to_alloc_vec

leighmcculloch:
### What
Add Bytes to_buffer and to_alloc_vec functions that provide access to the bytes in local guest memory. The to_buffer function provides access to the bytes in a statically specified buffer. The to_alloc_vec function is available only when the alloc feature is enabled and provides access to an alloc::vec::Vec containing the bytes.

### Why
@tomerweller was talking to me about allocator usage in dev and testing when experimenting and costs are not that high, and it got me thinking that I mostly don't use the allocator because it's not trivial to get data into an allocated type. It's not that hard, but I don't think it's obvious and it's something we need to explain to people how to do.

These functions provide two ways to get access to the data inside a variable length Bytes value, whether using the allocator or not.

The first, `to_buffer` accepts as a constant generic parameter a buffer size, and returns the buffer filled with the contents of the bytes along with a `Range` that describes the slice of that buffer that contains the bytes. The bytes may not fill the full buffer and the range that case describes where the bytes begins (always 0) and ends. The function is usable regardless of whether the allocator is in use, and is suitable when a reasonable maximum size is known, but an exact size is not.

The `to_alloc_vec` function requires the `alloc` feature as it dynamically allocates the memory required to store the bytes on the guest side.

### Costs

Examples of costs of using these functions:

#### `to_alloc_vec`

The `to_alloc_vec` function with a small bytes value results in:

- Contract Size: 3,835 bytes
- CPU Used: 5,001,490
- Bytes Read: 3,992
- Fee Charged: 59,245

```
#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn empty(env: Env) -> u32 {
        let bytes = Bytes::from_slice(&env, &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
        let len = bytes.to_alloc_vec().as_slice().len();
        len as u32
    }
}
```

#### `to_buffer::<10>`

The `to_buffer` function with a buffer size of 10 and a small bytes value results in:

- Contract Size: 599 bytes
- CPU Used: 3,828,048
- Bytes Read: 756
- Fee Charged: 44,351

```
#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn empty(env: Env) -> u32 {
        let bytes = Bytes::from_slice(&env, &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
        let len = bytes.to_buffer::<10>().as_slice().len();
        len as u32
    }
}
```

#### `to_buffer::<1024>`

The `to_buffer` function with a buffer size of 1024 and a small bytes value results in:

- Contract Size: 807 bytes
- CPU Used: 3,914,448
- Bytes Read: 964
- Fee Charged: 45,418

```
#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn empty(env: Env) -> u32 {
        let bytes = Bytes::from_slice(&env, &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
        let len = bytes.to_buffer::<10>().as_slice().len();
        len as u32
    }
}
```

### DIFF
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -649,8 +649,11 @@ impl Bytes {
     }
 }
 
-/// A BytesBuffer stores bytes in a buffer of given size and makes the bytes
-/// stored available.
+/// A `BytesBuffer` stores a variable number of bytes, up to a fixed limit `B`.
+///
+/// The bytes are stored in a fixed-size non-heap-allocated structure. It is a
+/// minimal wrapper around a fixed-size `[u8;B]` byte array and a length field
+/// indicating the amount of the byte array containing meaningful data.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BytesBuffer<const B: usize> {
     buffer: [u8; B],
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -629,7 +629,7 @@ impl Bytes {
         }
         BytesBuffer {
             buffer,
-            range: 0..len,
+            len,
         }
     }
 
@@ -654,7 +654,7 @@ impl Bytes {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BytesBuffer<const B: usize> {
     buffer: [u8; B],
-    range: Range<usize>,
+    len: usize,
 }
 
 impl<const B: usize> Borrow<[u8]> for BytesBuffer<B> {
@@ -667,7 +667,7 @@ impl<const B: usize> Borrow<[u8]> for BytesBuffer<B> {
 impl<const B: usize> BytesBuffer<B> {
     /// Returns a borrow slice of the bytes stored in the BytesBuffer.
     pub fn as_slice(&self) -> &[u8] {
-        &self.buffer[self.range.clone()]
+        &self.buffer[..self.len]
     }
 }
 
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -1,10 +1,13 @@
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
 use core::{
     borrow::Borrow,
     cmp::Ordering,
     convert::Infallible,
     fmt::Debug,
     iter::FusedIterator,
-    ops::{Bound, RangeBounds},
+    ops::{Bound, Range, RangeBounds},
 };
 
 use super::{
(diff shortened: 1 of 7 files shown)


soroban-sdk/src/bytes.rs
```diff
@@ -602,6 +605,70 @@ impl Bytes {
...
+
+    /// Copy the bytes into a Rust alloc Vec of size matching the bytes.
+    ///
+    /// Returns the Vec. Allocates using the built-in allocator.
+    ///
+    /// Suitable when the size of the bytes isn't a fixed size and the allocator
+    /// functionality of the sdk is enabled.
+    #[cfg(feature = "alloc")]
+    #[must_use]
+    pub fn to_alloc_vec(&self) -> alloc::vec::Vec<u8> {
```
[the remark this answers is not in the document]
graydon: Unsure. I think `Box<[u8]>` is a bit niche / unknown among rust programmers; but it's a smaller type (2 word fat-pointer vs. 3-word fat-pointer-with-bound), so technically a more minimal thing to return. And you can trivially upgrade it (without reallocating) via `Vec<u8>::from`. Maybe add two methods, `to_alloc_boxed` and `to_alloc_vec` that just calls `to_alloc_boxed().into()`?

soroban-sdk/src/bytes.rs
```diff
@@ -602,6 +605,70 @@ impl Bytes {
...
+        vec
+    }
+}
+
+/// A BytesBuffer stores bytes in a buffer of given size and makes the bytes
+/// stored available.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BytesBuffer<const B: usize> {
+    buffer: [u8; B],
+    range: Range<usize>,
```
graydon, line 657: Why a range rather than just a single usize indicating the limit?

soroban-sdk/src/bytes.rs
```diff
@@ -602,6 +605,70 @@ impl Bytes {
...
+    pub fn to_alloc_vec(&self) -> alloc::vec::Vec<u8> {
+        let len = self.len() as usize;
+        let mut vec = alloc::vec::from_elem(0u8, len);
+        self.copy_into_slice(&mut vec);
+        vec
+    }
+}
+
+/// A BytesBuffer stores bytes in a buffer of given size and makes the bytes
+/// stored available.
```
graydon, line 653:
```suggestion
/// A `BytesBuffer` stores a variable number of bytes, up to a fixed limit `B`, in a fixed-size
/// non-heap-allocated structure. It is a minimal wrapper around a fixed-size `[u8;B]` byte
/// array and a length field indicating the amount of the byte array containing meaningful data.
```

graydon: Given this type really only exists to cope with awkwardness around heap-allocation and size-variability, I think it'd be helpful to highlight exactly what it means in those terms in its doc-comment.

graydon (changes requested):
Generally like it! I expect most people will use BytesBuffer<N> for some fixed-size N and  be perfectly happy with that! Couple small nits inline.

soroban-sdk/src/bytes.rs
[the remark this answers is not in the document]
leighmcculloch: Sounds good to me.

It occurred to me it's not particularly simple with stable APIs to make a boxed slice without first making a Vec anyway. So I'll leave it as the vec.

graydon (approved):
You decided against Box<[u8]>? Either way, LGTM

leighmcculloch:
> You decided against Box<[u8]>?

The functionality for creating Box for slices is all nightly apis. I couldn't see a way to create one without using unsafe so I bailed. You can create one from a Vec<u8> but that seemed a bit pointless as it would still pull in Vec. 

Lmk if I missed something and there's a safe way to do it.

Issue #1221, opened 2024-02-02
Don't generate native account addresses with Arbitrary


brson:
### What version are you using?

20.2

---

The `Arbitrary` impl for `Address` currently can generate either contract addresses or native addresses, since https://github.com/stellar/rs-soroban-sdk/pull/1122.

After further experience writing tests with native addresses, I think adding this support was a mistake, as the SDK does not provide the additional tools needed to correctly test using native addresses.

In particular, native addresses need to have storage initialized, trustlines setup, and they cannot sign transactions without additional support code that is not obviously available to testers.

I have written tests where I had to filter out the native addresses because I didn't have those capabilities available.

For now, I think it would be good to just remove the ability to generate native addresses via Arbitrary. This might be considered a regression, but I suspect few are using it anyway.

---

My token fuzzer [does have those capabilities](https://github.com/brson/soroban-token-fuzzer/blob/master/src/addrgen.rs#L16) but only with a bunch of custom code that is quite fiddly. It does not use the SDKs ability to generate addresses, but instead has a randomly-generated address "generator" (builder) that can set up storage, create trustlines, and sign transactions.

Something similar might be useful in the SDK.

PR #1240, opened 2024-03-04
Revert "Generate all types of addresses when fuzzing. (#1122)"

brson:
This reverts commit 6b04bfc0509519f846379928f0143c5079b1f936.

Closes https://github.com/stellar/rs-soroban-sdk/issues/1221

### What

In https://github.com/stellar/rs-soroban-sdk/pull/1122, I changed the impl of `Arbitrary` for `Address` to generate all types of addresses instead of only contract addresses. I no longer think that was the correct thing to do, so this changes the impl back to only generate contract addresses.

### Why

Described in the issue https://github.com/stellar/rs-soroban-sdk/issues/1221

In short, native addresses are difficult to test with, particularly in conjunction with the native token contract, because they require set up of storage and trustlines. Doing this setup is possible, but is not obvious, and users running into such errors during testing are likely to get stuck.

### Known limitations

This is arguably a breaking change, though only to the testing environment, and only to fuzzers. I think it is likely there are so few fuzzing users that nobody will notice; better to do it sooner than later.

### DIFF
--- a/soroban-sdk/src/testutils/arbitrary.rs
+++ b/soroban-sdk/src/testutils/arbitrary.rs
@@ -627,7 +627,7 @@ mod objects {
 
     #[derive(Arbitrary, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
     pub struct ArbitraryAddress {
-        inner: crate::env::xdr::ScAddress,
+        inner: [u8; 32],
     }
 
     impl SorobanArbitrary for Address {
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
I think this is reasonable and right thing to do. 👍🏻

leighmcculloch:
The semver-checks job failed due to network connectivity. I'm rerunning the job in isolation to check that's the case, and if it passes will merge again.

PR #1242, opened 2024-03-15
Switch SDK to env v20.3.0

dmkozh:
### What

Switch SDK to env v20.3.0.

This doesn't require any functional changes (preliminary integration work has already been done before).

### Why

Keeping the dependencies up to date.

### Known limitations

N/A


anupsdf (approved):
LGTM

PR #1254, opened 2024-04-17
Bump versions; separate extend contract code and TTL for contract code and instance

jayz22:
### What
- Bump env and xdr, pick up `Hash` type.
- Adapt the deployer to allow bumping TTL for contract code and instance separately.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,18 +42,18 @@ soroban-token-sdk = { version = "20.5.0", path = "soroban-token-sdk" }
 
 [workspace.dependencies.soroban-env-common]
 version = "=21.0.0"
-git = "https://github.com/stellar/rs-soroban-env"
-rev = "8867a5d5b05681bde5eb8cf2031f615b1e8d8889"
+# git = "https://github.com/stellar/rs-soroban-env"
+# rev = "8867a5d5b05681bde5eb8cf2031f615b1e8d8889"
 
 [workspace.dependencies.soroban-env-guest]
 version = "=21.0.0"
-git = "https://github.com/stellar/rs-soroban-env"
-rev = "8867a5d5b05681bde5eb8cf2031f615b1e8d8889"
+# git = "https://github.com/stellar/rs-soroban-env"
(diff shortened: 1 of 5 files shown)


soroban-sdk/src/deploy.rs
```diff
@@ -152,7 +152,39 @@ impl Deployer {
     /// The TTL is the number of ledgers between the current ledger and the final ledger the data can still be accessed.
     pub fn extend_ttl(&self, contract_address: Address, threshold: u32, extend_to: u32) {
         self.env
-            .extend_contract_instance_and_code_ttl(
+            .extend_contract_instance_ttl(
```
dmkozh, line 155: I don't think this implementation should be changed - we anyway need to maintain the old host fn and it's more optimal to make a single call.

jayz22: The implementation inside the host of these two versions are identical. Replacing the call to "extend_contract_instance_and_code_ttl" gives chance that this function can be retired (with `max_supported_protocol_number`) one day, since no new contracts will be linked to it. The extra call to host function is negligible (310 cpu instructions for the dispatch).

Cargo.toml
```diff
@@ -43,17 +43,17 @@ soroban-token-sdk = { version = "20.5.0", path = "soroban-token-sdk" }
 [workspace.dependencies.soroban-env-common]
 version = "=21.0.0"
 git = "https://github.com/stellar/rs-soroban-env"
-rev = "4d9b1020f7b5baa0f972eb08ed8aa2cd3343ad7c"
+rev = "8867a5d5b05681bde5eb8cf2031f615b1e8d8889"
```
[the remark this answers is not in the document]
jayz22: Done.

soroban-sdk/src/deploy.rs
jayz22: It's a fair argument, there is contract code size as well as small perf benefits to keeping it. I reverted it back.

dmkozh (approved):

PR #1257, opened 2024-04-18
Bump version to 21.0.0-preview.

dmkozh:
### What

Bump version to 21.0.0-preview.

### Why

We're still going to have some small, but breaking changes, so this can't be treated as stable release yet.

### Known limitations

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,16 +29,16 @@ members = [
 ]
 
 [workspace.package]
-version = "21.0.0"
+version = "21.0.0-preview"
 rust-version = "1.74.0"
 
 [workspace.dependencies]
-soroban-sdk = { version = "21.0.0", path = "soroban-sdk" }
-soroban-sdk-macros = { version = "21.0.0", path = "soroban-sdk-macros" }
-soroban-spec = { version = "21.0.0", path = "soroban-spec" }
-soroban-spec-rust = { version = "21.0.0", path = "soroban-spec-rust" }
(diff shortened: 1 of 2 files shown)


sisuresh (approved):

Cargo.toml
```diff
@@ -29,16 +29,16 @@ members = [
 ]
 
 [workspace.package]
-version = "21.0.0"
+version = "21.0.0-preview"
```
leighmcculloch, line 32:
```suggestion
version = "21.0.0-preview.1"
```
Can the preview release be given a number? The number isn't required for cargo, but our versioning scheme always places a number so that the format is consistent. See https://docs.google.com/document/d/17P-COUnYGGE8JWdkhFfARORTpnqpBhDbuodC1CqL9KY.

leighmcculloch, line 32:
```suggestion
version = "21.0.0-preview.1"
```
Can the preview release be given a number? The number isn't required for cargo, but our versioning scheme always places a number so that the format is consistent. See https://docs.google.com/document/d/17P-COUnYGGE8JWdkhFfARORTpnqpBhDbuodC1CqL9KY.

leighmcculloch (changes requested):

Cargo.toml
dmkozh: Yeah, we can do that, but I believe this particular PR can probably be closed in favor of automatically created PR that picks up Hash-related changes.

Issue #1162, opened 2023-11-15
Imitate state archival invariants in testutils


dmkozh:
Currently host doesn't contain any state archival-related logic besides providing a way to bump the entries. It's up to the embedder (Core) to enforce the state archival invariants, such as never passing expired entries to the host. We probably don't want to change this logic beyond maybe enforcing the invariants better (https://github.com/stellar/rs-soroban-env/issues/1216).

Test framework, however, doesn't try to enforce and maintain these invariants, which causes confusing behavior of host when e.g. ledger sequence is bumped or storage is set up with expired entries. So we should make sure that expired entries are treated as missing (in case of temp storage) or as errors (in case of persistent storage). This will probably need some env-side support (conditioned on testutils).

leighmcculloch:
I think this probably is entirely an env side feature and we should move the issue there. The SDK is for the most part rather lightweight in its test utilities, just adding sugar and glue, but the solids of the test utilities is really the host.

leighmcculloch:
After reading your comment on this other issue (quoted below), I better understand what is being proposed here. Yes implementing this in the SnapshotSource would be an option. I'd still probably do it in the Env, but it can be used here.

> Of course, neither of these invariants holds in the test environment at the moment. I think we should approach this holistically and probably try imitating the Core behavior via customizing Snapshot implementation. That would take some time though, so I've opened some issues: https://github.com/stellar/rs-soroban-sdk/issues/1162, https://github.com/stellar/rs-soroban-env/issues/1216

Ref: 

I

dmkozh:
Yeah, I've been thinking about using the snapshot (which provides a way to inject some custom storage behavior). But while writing this I've realized that it might not be sufficient. I wouldn't sweat on the project too much - the end result is SDK usability, so I'd keep this here. We can add env support as needed.

brson:
I have added a workaround for the test environment seeing expired storage entries to soroban-token-fuzzer (in https://github.com/brson/soroban-token-fuzzer/commit/6a2dc1766ed1b245414116a5f69a4c988ebda4b5#diff-4e7dcf48a54935577ceeef30454dd5161735a3659dc613b6190490df86eef4bbR654).

The way soroban-token-fuzzer advances time is by exporting a snapshot into the env, mutating the sequence number and timestamp of the snapshot, then creating a new Env from the mutated snapshot.

As part of this process, it now just iterates over the snapshot ledger entries and deletes expired ones:

```rust
[...]

PR #1267, opened 2024-04-30
Add wrappers for `live_until_ledger` getters.

dmkozh:
### What

Add wrappers for `live_until_ledger` getters.

Also cover the emulation of state expiration with some tests.

### Why

Improving test utilities: https://github.com/stellar/rs-soroban-sdk/issues/1162

### Known limitations

N/A


sisuresh (approved):

soroban-sdk/src/storage.rs
```diff
@@ -617,18 +617,39 @@ mod testutils {
...
+                .unwrap()
+        }
     }
 
     impl testutils::storage::Temporary for Temporary {
         fn all(&self) -> Map<Val, Val> {
             all(&self.storage.env, xdr::ContractDataDurability::Temporary)
         }
+
+        fn get_live_until_ledger<K: IntoVal<Env, Val>>(&self, key: &K) -> u32 {
```
leighmcculloch, line 647: What does the phrase "live until ledger" mean? The phrase is not particularly clear or intuitive. Is the value returned the ledger it will live till?

dmkozh: I documented this as `Gets the ledger until which the temp entry with a given key lives.` . This is the term we've been using for a while in env - TTL describes the *time* to the entry lives (i.e. the number of ledgers), whereas here we're talking specifically about the *sequence* of the ledger until which it lives. But I'm open to suggestions for a more clear name to use in the SDK.

leighmcculloch: `get_ttl` // get number of ledgers it lives
`get_ttl_ledger` // get ledger it lives till

leighmcculloch: `get_live_till_ledger`

dmkozh: Hmm, 'till' and 'until' have the exactly same meaning, don't they? (at least according to the online resources I could find with a quick search. 
`get_ttl` looks better, but as I've mentioned TTL is different from the ledger sequence. I actually don't mind returning the TTL instead of the ledger sequence and naming these all `get_ttl` - honestly, I don't have a good intuition as to which one is more convenient to use in the tests. WDYT?

leighmcculloch: I agree, I had the same thought pattern. TTL is instantly recognisable, but yup, it shouldn't return an exact sequence. Given this is for tests I think we have an opportunity to try one and improve it by adding another testutil in a minor version release if it doesn't quite meet the mark.

We could add `get_ttl` today, and if it doesn't satisfy folks use cases, we could add `get_ttl_seq` or `get_archive_ledger_seq` or something else.

@brson thoughts? since you're someone who ran into a need for these utilities.

dmkozh: I've changed the names/semantics to TTL. I actually think that makes a bit more sense, given that we have `extend_ttl` functions.

leighmcculloch: I love it. I feel like we're in sync today, I was just about to comment to say about the concept consistency with `extend_ttl` 😆. 👍🏻

leighmcculloch (approved):
🥳

PR #1272, opened 2024-05-07
Add a conversion from Hash->Bytes

leighmcculloch:
### What
Add a conversion from `Hash`->`Bytes`.

### Why
In a prototype I'm working on I take a `Hash` value and append it to an existing `Bytes`, and noticed that prior to this release of the SDK I could call `.into()` on the value because it was prior a `BytesN<32>`. Now that it is a `Hash` I don't seem to be able to call `.into()` on it in the same context.

Since the type is trivially convertible to a `BytesN<32>` it seems reasonable to keep the trivial convertibility into a `Bytes`.

### DIFF
--- a/soroban-sdk/src/crypto.rs
+++ b/soroban-sdk/src/crypto.rs
@@ -71,6 +71,12 @@ impl<const N: usize> IntoVal<Env, BytesN<N>> for Hash<N> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
I'd like to slip this into the final release of the SDK.

jayz22 (approved):

dmkozh (approved):

PR #1281, opened 2024-06-07
Variety of small clippy inspired refactors

leighmcculloch:
### What
Variety of small clippy inspired refactors

### Why
Some are meaningful, some are clippy opinionated. Just a little house keeping.

### DIFF
--- a/soroban-sdk-macros/src/derive_enum.rs
+++ b/soroban-sdk-macros/src/derive_enum.rs
@@ -88,7 +88,7 @@ pub fn derive_type_enum(
                     enum_ident,
                     &case_num_lit,
                     &case_name_str_lit,
-                    &case_name,
+                    case_name,
                     case_ident,
                     &variant.attrs,
                     &variant.fields,
@@ -114,7 +114,7 @@ pub fn derive_type_enum(
                     enum_ident,
                     &case_num_lit,
                     &case_name_str_lit,
-                    &case_name,
+                    case_name,
                     case_ident,
                     &variant.attrs,
                 );
@@ -378,7 +378,7 @@ fn map_tuple_variant(
         ScSpecUdtUnionCaseV0::TupleV0(ScSpecUdtUnionCaseTupleV0 {
             doc: docs_from_attrs(attrs),
             name: case_name.try_into().unwrap_or_else(|_| StringM::default()),
-            type_: field_types.try_into().unwrap(),
+            type_: field_types,
         })
     };
 
--- a/soroban-sdk-macros/src/derive_enum_int.rs
+++ b/soroban-sdk-macros/src/derive_enum_int.rs
@@ -20,7 +20,7 @@ fn derives_copy(attrs: &[Attribute]) -> bool {
                 if let Some(_tt) = ml.tokens.clone().into_iter().find(|tt| {
                     // match this token with what we want
                     if let proc_macro2::TokenTree::Ident(id) = tt {
-                        id.to_string() == "Copy"
+                        id == "Copy"
                     } else {
                         false
                     }
@@ -48,10 +48,7 @@ pub fn derive_type_enum_int(
     if !derives_copy(attrs) {
         errors.push(Error::new(
             enum_ident.span(),
-            format!(
-                "enum integer {} must have `derive(Copy)`",
-                enum_ident.to_string()
-            ),
+            format!("enum integer {enum_ident} must have `derive(Copy)`"),
         ));
     }
 
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -28,7 +28,7 @@ use proc_macro::TokenStream;
 use proc_macro2::{Literal, Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
 use sha2::{Digest, Sha256};
-use std::fs;
+use std::{fmt::Write, fs};
 use syn::{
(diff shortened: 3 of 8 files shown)


soroban-sdk-macros/src/derive_enum.rs
```diff
@@ -88,7 +88,7 @@ pub fn derive_type_enum(
                     enum_ident,
                     &case_num_lit,
                     &case_name_str_lit,
-                    &case_name,
+                    case_name,
```
leighmcculloch, line 91: The & is redundant as its already a ref.

soroban-sdk-macros/src/derive_enum.rs
```diff
@@ -378,7 +378,7 @@ fn map_tuple_variant(
         ScSpecUdtUnionCaseV0::TupleV0(ScSpecUdtUnionCaseTupleV0 {
             doc: docs_from_attrs(attrs),
             name: case_name.try_into().unwrap_or_else(|_| StringM::default()),
-            type_: field_types.try_into().unwrap(),
+            type_: field_types,
```
leighmcculloch, line 381: The type is already the same type as the destination, so the conversion is redundant.

soroban-sdk-macros/src/derive_enum_int.rs
```diff
@@ -20,7 +20,7 @@ fn derives_copy(attrs: &[Attribute]) -> bool {
                 if let Some(_tt) = ml.tokens.clone().into_iter().find(|tt| {
                     // match this token with what we want
                     if let proc_macro2::TokenTree::Ident(id) = tt {
-                        id.to_string() == "Copy"
+                        id == "Copy"
```
leighmcculloch, line 23: The value is already inherently comparable and doesn't need converting into a String.

soroban-sdk-macros/src/derive_enum_int.rs
```diff
@@ -48,10 +48,7 @@ pub fn derive_type_enum_int(
     if !derives_copy(attrs) {
         errors.push(Error::new(
             enum_ident.span(),
-            format!(
-                "enum integer {} must have `derive(Copy)`",
-                enum_ident.to_string()
-            ),
+            format!("enum integer {enum_ident} must have `derive(Copy)`"),
```
leighmcculloch, line 51: Embedding the value calls the same logic as to string.

soroban-sdk-macros/src/lib.rs
```diff
@@ -346,11 +346,10 @@ pub fn contractmeta(metadata: TokenStream) -> TokenStream {
...
             "__CONTRACT_KEY_{}",
-            args.key
-                .as_bytes()
-                .iter()
-                .map(|b| format!("{b:02x}"))
-                .collect::<String>()
+            args.key.as_bytes().iter().fold(String::new(), |mut s, b| {
+                let _ = write!(s, "{b:02x}");
+                s
+            })
```
leighmcculloch, line 352: I almost didn't make this change but clippy made the case that the original code creates a lot of strings and allocates more. It doesn't really matter though as this is a short bytes to convert. I mostly updated it to get rid of the clippy warning.

soroban-sdk/src/crypto.rs
```diff
@@ -71,15 +71,15 @@ impl<const N: usize> IntoVal<Env, BytesN<N>> for Hash<N> {
...
+        v.0.into()
     }
 }
 
-impl<const N: usize> Into<BytesN<N>> for Hash<N> {
-    fn into(self) -> BytesN<N> {
-        self.0
+impl<const N: usize> From<Hash<N>> for BytesN<N> {
+    fn from(v: Hash<N>) -> Self {
+        v.0
```
leighmcculloch, line 82: From impls result in an automatic Into implementation anyway, but provide the additional from functions on the Bytes type.

soroban-sdk/src/prng.rs
```diff
@@ -454,9 +454,7 @@ impl Fill for u64 {
 impl Gen for u64 {
     fn gen(prng: &Prng) -> Self {
         let env = prng.env();
-        internal::Env::prng_u64_in_inclusive_range(env, u64::MIN.into(), u64::MAX.into())
-            .unwrap_infallible()
-            .into()
+        internal::Env::prng_u64_in_inclusive_range(env, u64::MIN, u64::MAX).unwrap_infallible()
```
leighmcculloch, line 457: All the into's were redundant as the types are already the correct type, u64.

soroban-sdk/src/storage.rs
```diff
@@ -175,7 +175,7 @@ impl Storage {
         V: TryFromVal<Env, Val>,
     {
         let key = key.into_val(&self.env);
-        if self.has_internal(key, storage_type.clone()) {
+        if self.has_internal(key, storage_type) {
```
leighmcculloch, line 178: The value is Copy so cloning is unnecessary.

soroban-sdk/src/string.rs
```diff
@@ -239,6 +239,11 @@ impl String {
         self.env().string_len(self.obj).unwrap_infallible().into()
     }
 
+    #[inline(always)]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
```
leighmcculloch, line 245: It's customary to provide an is empty fn alongside any len function.

graydon (approved):
lgtm

PR #1292, opened 2024-07-11
Remove need for Copy on error enums

### LINKED ISSUE #630: Add a clear error message for when an integer enum doesn't have derive(Copy)
Integer enums to be used as contracttype's need to have `derive(Copy)`. This is so that we can convert a reference to an integer enum into a u32.

Right now the error message displayed is ambiguous because it is an error generated from generated code that uses the type in a way that requires a Copy.

We should try and detect from the other attributes on the type whether the type has Copy derived. If it hasn't, we should display a focused error message.

This is a follow up to https://github.com/stellar/rs-soroban-sdk/issues/217

leighmcculloch:
### What
Remove need for Copy/Clone/etc on error enums and int enums.

### Why
Less compiler errors for folks just getting started.

The generated code currently requires that error enums impl the Copy trait because they use a copy within. That copy is unnecessary and this change removes it, removing the need for Copy to be derived on error enums.

I was confident that the copy had been put there as an optimisation, but when I looked into the code that was relying on it I found not optimisation was in use. When I tried to add the optimisation, I couldn't find a contract size difference.

A while ago we in https://github.com/stellar/rs-soroban-sdk/pull/1293 added better error messages for needing copy, which was a good fix at the time, I just think we can remove the need for this all together.

Close #630


soroban-sdk-macros/src/derive_enum_int.rs
```diff
@@ -144,15 +143,19 @@ pub fn derive_type_enum_int(
                 type Error = #path::xdr::Error;
                 #[inline(always)]
                 fn try_into(self) -> Result<#path::xdr::ScVal, #path::xdr::Error> {
-                    Ok((*self as u32).into())
+                    Ok(match self {
+                        #(#try_intos,)*
+                    })
```
leighmcculloch, line 148: This optimisation doesn't matter, it's in testutils, it doesn't impact wasm builds.

dmkozh (approved):

PR #1295, opened 2024-07-12
Add in contract assertions to prng functions

### LINKED ISSUE #1165: Use Host has_frame logic in SDK to panic fns called when in or not in a contract as required
Dependent on:

Once the environment has enter/exit hooks for top-level call (invoke), the SDK environment can use the hooks to track if its currently executing inside a contract or not, and use that information to provide more useful error messages when developers accidentally call Env SDK functions that only work inside contracts.

The SDK will:

on enter hook:
```
|| {
    env.in_contract = true
}
```
on exit hook:
```
|| {
    env.in_contract = false
}
```
then use it in fns that only work inside contract fns:
```
if !env.in_contract {
    panic!("fn {myname} called outside of a contract, use env.as_contract or call the function inside a contract function...")
}
```

leighmcculloch:
### What
Add in contract assertions to prng functions.

### Why
It's really easy to think you can use the prng functions inside your tests to help setup random test data, without reaslising that these functions are restricted to use within a contract.

Close #1165

### DIFF
--- a/soroban-sdk/src/storage.rs
+++ b/soroban-sdk/src/storage.rs
@@ -86,8 +86,7 @@ impl Storage {
     /// This should be used for data that requires persistency, such as token
     /// balances, user properties etc.
     pub fn persistent(&self) -> Persistent {
-        #[cfg(any(test, feature = "testutils"))]
-        self.verify_in_contract("persistent");
+        assert_in_contract!(self.env);
 
         Persistent {
             storage: self.clone(),
@@ -106,8 +105,7 @@ impl Storage {
     /// This should be used for data that needs to only exist for a limited
     /// period of time, such as oracle data, claimable balances, offer, etc.
     pub fn temporary(&self) -> Temporary {
-        #[cfg(any(test, feature = "testutils"))]
-        self.verify_in_contract("persistent");
+        assert_in_contract!(self.env);
 
         Temporary {
             storage: self.clone(),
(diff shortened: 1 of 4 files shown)


soroban-sdk/src/storage.rs
```diff
@@ -299,17 +296,6 @@ impl Storage {
...
-    #[cfg(any(test, feature = "testutils"))]
-    fn verify_in_contract(&self, storage_kind: &str) {
-        if !self.env.in_contract() {
-            panic!(
-                "'{storage_kind}' storage is not accessible outside of the \
-                    contract, wrap the call with `env.as_contract()` to \
-                    access it from a particular contract"
-            );
-        }
-    }
```
leighmcculloch, line 312: @dmkozh Hope you don't mind I took the change you pushed in #1288 and repurposed it for use across the SDK as a quick assertion drop in macro. I changed it to a macro so the assertion will render as occurring at the location of use and generalised the error message to find a balance between being as useful to being easy to drop around the SDK.

leighmcculloch (commented):
There are more places we can drop this assertion, such as the deploy functions. I'm marking the original issue #1165 as closed after this change because once this has merged we have addressed the two biggest occurrences where folks get confused which is the storage and prng APIs.

soroban-sdk/src/storage.rs
dmkozh: That's indeed nicer, thanks.

dmkozh (approved):

PR #1297, opened 2024-07-16
Fix client code gen to use efficient symbol

### LINKED ISSUE #1290: Client is not the zero-cost abstraction it hoped to be

Unknown

### What did you do?

Reported here that when importing the client the size of the wasm was 184 bytes larger than when not importing it:

> Changing to
> 
> ```rust
> mod wallet {
>     use soroban_sdk::auth::Context;
>     soroban_sdk::contractimport!(file = "../target/wasm32-unknown-unknown/release/webauthn_secp256r1.wasm");
> }
> ```
> 
> and
> 
> ```rust
> wallet::Client::new(&env, &address).add(&id, &pk, &true);
> ```
> 
> Takes the contract from 1209 bytes to 1393.

Ref: https://github.com/stellar/stellar-protocol/discussions/1499#discussioncomment-10001935

Much less change in size. While the compiler will behave differently with different inputs and so we can't guarantee a truly zero cost to using the client the client was designed to cost no more than not using the client.

### What did you see instead?

184 byte change in size.

leighmcculloch:
### What
Change the client generated code to use the efficient compile time const Symbol generation for contract invocation function names when the function names are small enough.

### Why
The client is intended to be a zero-cost abstraction when compared with invoking a contract with `env.invoke_contract(cid, fn, args)`, but currently it is ~200 bytes more expensive in WASM size. This is because the generated client code is using runtime conversion from a &str to a Symbol for all symbols, even small ones that can be converted to a Symbol at compile time using const expressions.

This change reduces the size of contracts using a single short symbol by ~200 bytes.

Close #1290

### DIFF
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -1,8 +1,8 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{Error, FnArg, Path, Type, TypePath, TypeReference};
+use syn::{Error, FnArg, LitStr, Path, Type, TypePath, TypeReference};
 
-use crate::syn_ext;
+use crate::{symbol, syn_ext};
 
 pub fn derive_client_type(crate_path: &Path, ty: &str, name: &str) -> TokenStream {
     let ty_str = quote!(#ty).to_string();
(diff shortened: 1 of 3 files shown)


graydon (approved):

leighmcculloch:
@kalepail this is released in `v21.2.0`.

PR #1302, opened 2024-07-17
Fix Hash<> not cloneable

leighmcculloch:
### What
Implement Clone for the Hash<> type introduced in v21.

### Why
The BytesN<> type implements Clone, and the Hash<> type was added to replace BytesN<> in certain situations. It is harder to upgrade code in the wild that was written using BytesN<> because Hash<> isn't cloneable.

Hash<> can be cloned, there's no reason to keep it.

The type is also marked as transparent same as BytesN<> in this change.


sisuresh (approved):

leighmcculloch:
Released in 21.4.0.

PR #1308, opened 2024-07-25
Add fn for changing an Env's test config

### LINKED ISSUE #1306: Feature to easily opt out of test_snapshots generation
Test snapshots are big and not always used, especially in the early days of contract development. My preference is they be opt in, but at the very least there should be an easy way to opt out.

```rust
#[test]
#![no_test_snapshot]
fn test() {
}
```
or
```rust
fn test() {
let env = Env::default();
env.no_test_snapshots();
}
```

Pain, frustration and madness. (and large directories)

---

Should we be encouraging folks to `.gitignore` test_snapshots? Seems like yes again as they are large files and tbh I'm not entire sure what the intended flow for actually making use of them is. Do we actually document test snapshot usage anywhere?

leighmcculloch:
### What
Add fn for changing an Envs config after creation.

### Why
There are ways to create Envs without an opportunity to set a config.

It's cumbersome to offer a way to set the config with all of the different ways, and adding a function for changing it after creation is the simplest way today to achieve this.

This could create problems if in the future the config contains items that aren't easily changeable after instatiation of the Env. That problem doesn't exist today at least. In the future if we introduce items like that we'd probably need to resort to checking if the value is changing, and if so, panicing with a meaningful error message. That's not great, but still better for the overall API I think, especially given this is in tests.

Close #1306

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -487,6 +487,7 @@ impl Env {
         f((*self.test_state.generators).borrow_mut())
     }
 
+    /// Create an Env with the test config.
     pub fn new_with_config(config: EnvTestConfig) -> Env {
         struct EmptySnapshotSource();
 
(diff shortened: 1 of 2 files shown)


dmkozh (approved):

kalepail (approved):
Love it! Thanks

leighmcculloch:
Released in 21.4.0.

PR #1311, opened 2024-08-08
Expose rust code gen of contracts without the file import

leighmcculloch:
### What
Expose rust code gen of contracts without the file import.

### Why
The code gen function for rust code generates the interface and types as well as an import of the wasm binary. The CLI uses this function and wants to output the generated code without the wasm binary.

Related to:
- https://github.com/stellar/stellar-cli/pull/1532#discussion_r1708371038

### DIFF
--- a/soroban-spec-rust/src/lib.rs
+++ b/soroban-spec-rust/src/lib.rs
@@ -57,6 +57,14 @@ pub fn generate_from_wasm(
 }
 
 pub fn generate(specs: &[ScSpecEntry], file: &str, sha256: &str) -> TokenStream {
+    let generated = generate_without_file(specs);
+    quote! {
+        pub const WASM: &[u8] = soroban_sdk::contractfile!(file = #file, sha256 = #sha256);
+        #generated
+    }
+}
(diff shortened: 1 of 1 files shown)


Ifropc (approved):
Thanks Leigh, very helpful change! One question, we return `Result<TokenStream, syn::Error>` in `to_formatted_string`, is it intended? Downstream consumers can't use `?` operator for chaining the result without bringing syn dependency explicitly.
I'm not super familiar with canonical Rust way of forwarding errors, but would we want to wrap it into another Error enum?

leighmcculloch:
> One question, we return `Result<TokenStream, syn::Error>` in `to_formatted_string`, is it intended? Downstream consumers can't use `?` operator for chaining the result without bringing syn dependency explicitly.

That's a good point. You could `.map_err` to avoid the syn dep, but we could do as you say and wrap it. That's a bit unrelated to this change specifically, and would be a breaking change so it'd need to be something we'd do for a later release.

sisuresh (approved):

PR #1322, opened 2024-08-19
Allow contractimpls across mods

### LINKED ISSUE #1321: Splitting contract impls across modules leads to errors

v21.5.1

### What did you do?

`Cargo.toml`:
```toml
[package]
name = "test_modular"
version.workspace = true
authors = ["Stellar Development Foundation <info@stellar.org>"]
license = "Apache-2.0"
edition = "2021"
publish = false
rust-version.workspace = true

[lib]
crate-type = ["cdylib"]
doctest = false

[dependencies]
soroban-sdk = {path = "../../soroban-sdk"}

[dev-dependencies]
soroban-sdk = {path = "../../soroban-sdk", features = ["testutils"]}
```

`src/lib.rs`:
```rust
#![no_std]

mod feat1;
mod feat2;

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn base() -> u32 {
        0
    }
}
```

`src/feat1.rs`:
```rust
use soroban_sdk::contractimpl;

use crate::Contract;
use crate::ContractClient;

#[contractimpl]
impl Contract {
    pub fn one() -> u32 {
        1
    }
}
```

`src/feat2.rs`:
```rust
use soroban_sdk::contractimpl;

use crate::Contract;

#[contractimpl]
impl Contract {
    pub fn two() -> u32 {
        2
    }
}
```

Compiled successfully, and test successfully.

### What did you see instead?

```
$ soroban contract build
cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
   Compiling test_modular v21.5.1 (/Users/leighmcculloch/Code/rs-soroban-sdk/tests/modular)
error[E0412]: cannot find type `ContractClient` in this scope
 --> tests/modular/src/feat1.rs:6:1
  |
6 | #[contractimpl]
  | ^^^^^^^^^^^^^^^ not found in this scope
  |
  = note: this error originates in the attribute macro `soroban_sdk::contractclient` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this struct
  |
1 + use crate::ContractClient;
  |

error[E0412]: cannot find type `ContractClient` in this scope
 --> tests/modular/src/feat2.rs:6:1
  |
6 | #[contractimpl]
  | ^^^^^^^^^^^^^^^ not found in this scope
  |
  = note: this error originates in the attribute macro `soroban_sdk::contractclient` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this struct
  |
1 + use crate::ContractClient;
  |

For more information about this error, try `rustc --explain E0412`.
error: could not compile `test_modular` (lib) due to 2 previous errors
```

After adding in these lines to `feat1.rs` and `feat2.rs`:

```rust
use crate::ContractClient;
```

Build worked fine, but `cargo test` failed:

```
❯ cargo test
   Compiling test_modular v21.5.1 (/Users/leighmcculloch/Code/rs-soroban-sdk/tests/modular)
error[E0433]: failed to resolve: use of undeclared crate or module `__Contract_fn_set_registry`
 --> tests/modular/src/feat1.rs:6:1
  |
6 | #[contractimpl]
  | ^^^^^^^^^^^^^^^ use of undeclared crate or module `__Contract_fn_set_registry`
  |
  = note: this error originates in the attribute macro `contractimpl` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this module
  |
1 + use crate::__Contract_fn_set_registry;
  |

error[E0433]: failed to resolve: use of undeclared crate or module `__Contract_fn_set_registry`
 --> tests/modular/src/feat2.rs:6:1
  |
6 | #[contractimpl]
  | ^^^^^^^^^^^^^^^ use of undeclared crate or module `__Contract_fn_set_registry`
  |
  = note: this error originates in the attribute macro `contractimpl` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this module
  |
1 + use crate::__Contract_fn_set_registry;
  |

For more information about this error, try `rustc --explain E0433`.
error: could not compile `test_modular` (lib test) due to 2 previous errors
```

leighmcculloch:
### What
Add a new trait `ContractFunctionRegister` that wraps the existing `register` function that all contracts use to create a table of all their functions.

### Why

When a contract is registered with the test environment it uses the table of functions to know what functions a contract has defined.

Today contract functions get added to the table by using the ctor crate to call the `#mod::register` function where `#mod` is a generated module at the same location as where the `#[contract]` macro gets used.

The problem with this, and that was reported in #1321, is that it's valid in Rust to place type implementations anywhere in a crate that is able to reference the type, but there's no way for the generated code to know what relative or absolute path to use to reference that `#mod`.

The generated code assumes that the `#mod` is imported into the current scope, which it never will be because users have no reason to manually do it because it is hidden from them and they don't know it exists.

This situation leads to the following error that makes little sense to a developer because it refers to hidden generated code:
```
use of undeclared crate or module `__Contract_fn_set_registry`
```

Most of the time when folks add additional `impl` blocks to a type they will import the type into the current module so as to reference it, and so we can improve the existing situation by shifting the call to `#mod::register` to be a call to `Contract::register`.

To ensure that the `register` function does not clash with a contract function that may also be named `register` the function is defined on a hidden trait, `ContractFunctionRegister`, and is only used in the context of that trait.

There are some other minor changes included that remove assumptions about the type name of a contract being a simple identifier, so that when it is more than an identifier, such as a path (e.g. `path::Identifier`), the full type name is preserved everywhere it is needed.

Close #1321


leighmcculloch:
TODO:
- [ ] Make it possible to use the path keyword `crate` to reference the contract. For some reason using it is problematic because I think we combine it with `super::` somewhere.
- [ ] Find a way to auto-reference the contract client, because what happens is even if the user references the contract directly they still need to then import the contract client because of generated impls on the client.

leighmcculloch:
I think we should merge this as is, even if there are some edge cases not covered, because it covers an 80% of what's required to get it to work, and the edge cases may require significantly more refactoring. So it'd be good to get the minimal merged so that the refactor can be broken up, and be easier to review.

dmkozh (approved):
That's really nice, thanks!

PR #1334, opened 2024-09-10
Add __check_auth wasm test vectors and tests

leighmcculloch:
### What
Add new test vectors containing simple approve and decline __check_auth impls, and tests that use those wasm contracts.

### Why
To run some of the same tests that already exist, but with the contracts loaded in their WASM form, to ensure consistency across the handling of auth with natively loaded and wasm loaded contracts.

Originally these tests were written while debugging:
- https://github.com/stellar/rs-soroban-sdk/issues/1332
- https://github.com/stellar/rs-soroban-sdk/pull/1333


leighmcculloch:
Tests will pass after the following change merges:
- #1333

dmkozh (approved):

PR #1338, opened 2024-09-17
Add infallible From conversions for types to ScVal

leighmcculloch:
### What
Add infallible From conversions for types to ScVal, replacing the existing fallible TryFrom conversions.

### Why
When I look through tests I see lots of `try_into().unwrap()` when converting types to their ScVal values. This is unnecessary, we know that the conversions will succeed because the host won't give the SDK types that aren't valid and couldn't be converted. It's just a result of the way the Rust types are setup we can't guarantee the type is always convertible, but in practice they are.

Note that this is _not_ really a breaking change, because in the Rust core lib there is a blanket impl for all From conversions to also provide a TryFrom conversion. The only behaviour change is that if there were cases that would fail conversion previously a panic will occur instead of an error when using a TryFrom to do the conversion. In all cases there is no expected failure because the Env guarantees valid host types to be created and we're converting host types to ScVal types.


soroban-sdk/src/address.rs
```diff
@@ -124,18 +124,21 @@ impl TryFromVal<Env, &Address> for Val {
 }
 
 #[cfg(not(target_family = "wasm"))]
-impl TryFrom<&Address> for ScVal {
-    type Error = ConversionError;
-    fn try_from(v: &Address) -> Result<Self, ConversionError> {
-        Ok(ScVal::try_from_val(&v.env, &v.obj.to_val())?)
+impl From<&Address> for ScVal {
+    fn from(v: &Address) -> Self {
+        // This conversion occurs only in test utilities, and theroetically all
```
dmkozh, line 129: typo: theoretically (also in every other file that has this comment)

dmkozh (approved):

PR #1327, opened 2024-09-06
Support for constructor-related functionality in Soroban SDK.

dmkozh:
### What

The main non-test changes are updates for the auth-related data structures and a wrapper for the new deployer function (`create_contract_with_constructor`).

The remaining changes concern the test infrastructure support, specifically contract creation utilities that support passing constructor arguments.

### Why

Supporting constructors introduced in protocol 22.

### Known limitations

N/A

PR #1339, opened 2024-09-17
Revert version bump from #1327

leighmcculloch:
### What
Revert version bump from #1327.

### Why
The release process is that we bump just before release.

When we bump before release, it makes it harder to use unreleased changes as a patch against existing contracts dependent on the current version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,7 +1216,7 @@ dependencies = [
 
 [[package]]
 name = "soroban-ledger-snapshot"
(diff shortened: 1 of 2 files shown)


leighmcculloch:
@dmkozh had you needed to bump the major version because of the semver check failing?

leighmcculloch:
Ah yeah, that's why. I guess the release process we have was setup before we had the semver check and we haven't adapted since.

PR #1344, opened 2024-09-19
Fix warnings on contract and contractimpl macros

leighmcculloch:
### What
Fix and disable warnings on contract and contractimpl macros for misformatted item names, and missing fns.

### Why
Starting with the last release warnings started to display on the contract and contractimpl macros because of generated code not following the naming conventions for some items.

In one case it is easier to fix the naming convention, and so that is what was done.

In the other case it is easier to silence the warning, and so that is what was done.

One of the warnings was about a missing function. In this case I think there's an issue with rust-analyzer and how we were generating two dependent parts of code different ways. In one are we were always generating the code feature gated, in the other we were generating it only based on the feature. Rust-analyzer doesn't rebuild proc-macros frequently, and so I think caching of generated code that then had the feature off was to blame.

### Backporting

This change should be backported to the 21 versions as a patch on 21.7 after merging to main.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -244,15 +244,13 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
                 #imp
                 #derived_ok
             };
-            if cfg!(feature = "testutils") {
-                let cfs = derive_contract_function_registration_ctor(
-                    crate_path,
-                    ty,
-                    trait_ident,
-                    pub_methods.into_iter(),
-                );
-                output.extend(quote! { #cfs });
-            }
+            let cfs = derive_contract_function_registration_ctor(
+                crate_path,
+                ty,
+                trait_ident,
+                pub_methods.into_iter(),
+            );
+            output.extend(quote! { #cfs });
             output.into()
         }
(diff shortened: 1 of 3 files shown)


soroban-sdk-macros/src/lib.rs
```diff
@@ -244,15 +244,13 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
-            if cfg!(feature = "testutils") {
-                let cfs = derive_contract_function_registration_ctor(
-                    crate_path,
-                    ty,
-                    trait_ident,
-                    pub_methods.into_iter(),
-                );
-                output.extend(quote! { #cfs });
-            }
+            let cfs = derive_contract_function_registration_ctor(
```
dmkozh, line 247: I'm not sure I understand why did we remove the feature gate here.

leighmcculloch: Instead of feature getting the generation off the code, I've changed it to always generate and I've added a cfg feature gate as an attribute on the code. 

I don't know why, I'd guess a rust analyser bug, when we do it the original way it gets out of sync with the function it calls which is always generate and has a cfg feature gate via an attribute.

dmkozh (approved):

PR #1349, opened 2024-09-25
Use ConstructorArgs trait for args on deploy with constructor

### LINKED ISSUE #1347: Merge the deploy and deploy_with_constructor env fns
### What
Merge the deploy and deploy_with_constructor env fns.

### Why

Same reasoning as:
- https://github.com/stellar/rs-soroban-sdk/issues/1341

Follow up to:
- https://github.com/stellar/rs-soroban-sdk/pull/1343

leighmcculloch:
### What
Use ConstructorArgs trait for args on deploy with constructor, and rename deploy with constructor to deploy_v2.

### Why
So that types can be passed directly to the deploy function like they can now be passed to the register function since #1343.

I renamed the function to deploy_v2 for consistency with what we did when we added a second register for stellar asset contract function. I considered merging with the deploy function which would have been a breaking change, that I think we agreed would be worthwhile, but breaking is just such a bad experience and I think a _v2 is almost as good. We can consider merging the deploy functions at a later date.

Close #1347


dmkozh (approved):

soroban-sdk/src/deploy.rs
```diff
@@ -258,38 +260,36 @@ impl DeployerWithAddress {
...
-        let env = &self.env;
-        let address_obj = env
-            .create_contract(
-                self.address.to_object(),
-                wasm_hash.into_val(env).to_object(),
-                self.salt.to_object(),
-            )
-            .unwrap_infallible();
-        unsafe { Address::unchecked_new(env.clone(), address_obj) }
+        self.deploy_v2(wasm_hash, ())
```
leighmcculloch, line 265: Undo this change, so that deploy v1 still calls the old host fn.
```suggestion
        let env = &self.env;
        let address_obj = env
            .create_contract(
                self.address.to_object(),
                wasm_hash.into_val(env).to_object(),
                self.salt.to_object(),
            )
            .unwrap_infallible();
        unsafe { Address::unchecked_new(env.clone(), address_obj) }
```

PR #1361, opened 2024-10-04
Filter events the env exposes in test snapshots to contracts to contract events

leighmcculloch:
### What
Filter events the env exposes in test snapshots to contracts to contract events.

### Why
The main thing we need to test in test snapshots is that non-diagnostic events are consistent and captured.

Diagnostic events are noise here.

### Notes
There's no good time to make this change.

Ideally the change is made at a major version because it's breaking.

But at a major version is the ideal time to avoid the change so that people see only meaningful diffs on their test snapshots.


leighmcculloch:
@dmkozh Maybe we should also retain system events. I don't actually remember what system events exist though...

dmkozh:
> Maybe we should also retain system events. I don't actually remember what system events exist though...

Yeah, let's retain system events. They're underutilized, but we do currently emit a system event for the contract updates.

leighmcculloch:
@dmkozh I've updated it to retain system events. Ready for review.

graydon (approved):
I don't feel strongly about this as a feature/change either way; but practically it looks like a correct implementation so approving.

PR #1373, opened 2024-10-22
expose `Fr` into its own type

jayz22:
### What

Resolves https://github.com/stellar/rs-soroban-sdk/issues/1352

### Why

`Fr` is the 255-bit scalar with a fixed modulus (equal to the subgroup order `r`) and all its operations are modulo arithmetics. Exposing it as separate type (zero-sized wrapping of an `U256`) is better signaling that the two have different semantics and avoids potential footgun.

### Known limitations

[TODO or N/A]


soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -84,6 +88,14 @@ pub struct Fp(BytesN<48>);
...
+/// field of order `r` (the order of the G1 and G2 groups). The struct is
+/// internally represented with an `U256`, all arithmetic operations follow
+/// modulo `r`.
+#[derive(Clone)]
+#[repr(transparent)]
+pub struct Fr(U256);
+
 impl_bytesn_repr!(G1Affine, 96);
 impl_bytesn_repr!(G2Affine, 192);
 impl_bytesn_repr!(Fp, 48);
```
leighmcculloch, line 101: Can we add a conversion to BytesN? Or does that not make sense?

leighmcculloch (approved):

leighmcculloch:
Builds will pass after the following PR is merged:
- https://github.com/stellar/rs-soroban-sdk/pull/1375

soroban-sdk/src/crypto/bls12_381.rs
jayz22: yeah it makes sense. added conversion to/from BytesN is via U256.

leighmcculloch: @jayz22 Could you add a `impl_bytesn_rep!(Fr, 32)`, instead of implementing the two functions on their own?

jayz22: The `impl_bytesn_rep` is intended for a `BytesN` wrapping type, here the `Fr` is wrapping an `U256` and the only extra methods needed are conversion from bytes?

jayz22: Or you want me to add conversions to/from array (`[u8;32]`) as well, this are the only extra methods `impl_bytesn_rep` provides (the val-conversion methods are already implemented via U256).

leighmcculloch: Oh, sorry, yes you're right.

leighmcculloch: This is fine, lets merge. Feel free to add the follow up to/from, but I'll merge it now to avoid churning because this is good.

leighmcculloch: I assume the bytes are accessible via U256 anyway. Sorry for the above, I forgot it wasn't wrapping a BytesN.

PR #1401, opened 2024-12-06
Fix bug in contractimpl code generation for impl traits using associated types

### LINKED ISSUE #1400: Self references in CustomAccountInterface impls are broken by SDK v22

`22.0.0`

### What did you do?

This is an issue @dmkozh saw while updating the soroban-examples and he mentioned it to me.

Implement the custom account interface, using a `Self` reference in the `__check_auth` function.

The code worked before sdk 22, and required updating as follows to build:
```diff
 #[contractimpl]
 impl CustomAccountInterface for IncrementContract {
     type Signature = BytesN<192>;
     type Error = AccError;
     #[allow(non_snake_case)]
     fn __check_auth(
         env: Env,
         signature_payload: Hash<32>,
-         agg_sig: Self::Signature,
+         agg_sig: BytesN<192>,
         _auth_contexts: Vec<Context>,
     ) -> Result<(), AccError> {
         // The sdk module containing access to the bls12_381 functions
         let bls = env.crypto().bls12_381();
```
Ref: https://github.com/stellar/soroban-examples/pull/330/files#diff-6b53b8afe379dad726d8f8ac76049b737fd0b9a935d5fe5b0b5c0ac8a2535be9

The existing code and the `Self::Signature` should be allowed.

### What did you see instead?

```
   Compiling soroban-bls-signature v0.0.0 (/home/runner/work/soroban-examples/soroban-examples/bls_signature)
error[E0223]: ambiguous associated type
  --> src/lib.rs:78:18
   |
78 |         agg_sig: Self::Signature,
   |                  ^^^^^^^^^^^^^^^
   |
help: if there were a trait named `Example` with associated type `Signature` implemented for `IncrementContractArgs`, you could use the fully-qualified path
   |
78 |         agg_sig: <IncrementContractArgs as Example>::Signature,
   |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Ref: https://github.com/stellar/soroban-examples/actions/runs/12189741202/job/34005512450?pr=330

leighmcculloch:
### What

Allow associated types in impl of traits, specifically the CustomAccountInterface.

### Why

The use of associated types was allowed until v22 of the SDK, where an assumption in the code generation `contractimpl` macro is causing ambiguous code to be generated when associated types are in use.

Close #1400

### DIFF
--- a/soroban-sdk-macros/Cargo.toml
+++ b/soroban-sdk-macros/Cargo.toml
@@ -23,7 +23,7 @@ soroban-spec = { workspace = true }
 soroban-spec-rust = { workspace = true }
 soroban-env-common = { workspace = true }
 stellar-xdr = { workspace = true, features = ["curr", "std"] }
-syn = {version="2.0",features=["full"]}
+syn = {version="2.0.77",features=["full"]}
 quote = "1.0"
 proc-macro2 = "1.0"
 itertools = "0.10.5"
--- a/soroban-sdk-macros/src/syn_ext.rs
+++ b/soroban-sdk-macros/src/syn_ext.rs
@@ -1,5 +1,6 @@
+use std::collections::HashMap;
 use proc_macro2::TokenStream;
-use quote::{quote, ToTokens};
+use quote::{format_ident, quote, ToTokens};
 use syn::{
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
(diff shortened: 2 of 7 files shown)


leighmcculloch:
I wanted to do just as the error message said and replace the `Self::Signature` with `<Contract as CustomAccountInterface>::Signature`. We have all the information to do that, and that's what I did in this commit:
- [change the approach to using a qualified qself](https://github.com/stellar/rs-soroban-sdk/pull/1401/commits/c4b8cf16427c681e33ecc19cece8cd50a4d728fa)

However, that then broke the contract spec generation, the `map_type` function specifically, which can't handle those qself types.

But then digging further, it looks like `Self::Signature` never actually worked properly because `map_type` doesn't know how to map it to the underlying type.

So, in the interest of trying to do something I thought would be simple, I've gone with the approach that when the SDK macro first parses the `impl` block, if the block is implementing a trait, it flattens any associated types in the functions. At the moment in this PR it does it only for function inputs, but after this merges I'd like to follow up with doing the same thing with function return values.

leighmcculloch:
I also had to increase the version requirement on the syn package. The sdk said it required any 2.0 version, but in practice we were using 2.0.77 in the repo. In this PR I started using some functions that were introduced somewhere along the way. The fuzz example has its own lock file and it was using 2.0.37, which it turns out doesn't have those functions. Minimum-version-selection, if Cargo ever adopts it, would prevent situations like this.

sisuresh (approved):

PR #1403, opened 2024-12-06
Change visibility levels of generated soroban-sdk testutils code in acordance with clippy

cgorenflo:
### What
With the feature `testutils` enabled in the soroban-sdk, the procedural macro `contract` generates a private module named `__{contract_name}_fn_set_registry` for each contract. Some members used to have the visibility level `pub(crate)` and are now changed to visibility level `pub`.

### Why
This previously triggered the clippy rule [redundant_pub_crate](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pub_crate). As the rule explains, defining `pub(crate)` inside a private module is unnecessary due to the module's visibility level, and can be potentially misleading.

### Known limitations
This change does not impact functionality in any way and removes the clippy warning for teams that have it enabled, so there are no known limitations.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -154,15 +154,15 @@ pub fn contract(metadata: TokenStream, input: TokenStream) -> TokenStream {
                 use std::sync::Mutex;
                 use std::collections::BTreeMap;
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Released in:
- https://github.com/stellar/rs-soroban-sdk/releases/tag/v22.0.2

Issue #1319, opened 2024-08-16
Add automated resource cost reports


leighmcculloch:
### What problem does your feature solve?

For a developer who has invested in significant test coverage, they should be able to, through the execution of those tests, get a good signal for the costs their contract is expected require.

### What would you like to see?

An option when running tests, maybe enabled by default if there is no significant performance issue with having it enabled, where a summary of every contract function that was executed along with it's range of costs.

An option for a detailed view where costs are shown per test.

### What alternatives are there?

Developers can get this themselves on each test, but it's difficult to get that broad overview.

### Caveat

The budget in the unit tests is not as accurate to when txs are submitted, and this issue is probably of very low impact unless that accuracy is improved in the Soroban Env when simulating in tests.

dmkozh:
This should be addressed now by https://github.com/stellar/rs-soroban-sdk/pull/1411

PR #1411, opened 2024-12-12
Expose invocation metering in the SDK.

dmkozh:
### What

Expose invocation metering in the SDK.

### Why

Improving testing UX.

https://github.com/stellar/rs-soroban-sdk/issues/1319

### Known limitations

N/A


soroban-sdk/src/env.rs
```diff
@@ -576,6 +578,91 @@ impl Env {
...
+    /// The metered resources for the last invocation can be retrieved with
+    /// `get_last_invocation_resources()` and the estimated fee corresponding
+    /// to these resources can be estimated with
+    /// `estimate_last_invocation_fee()`.
+    ///
+    /// While the resource metering may be useful for contract optimization,
+    /// keep in mind that resource and fee estimation may be imprecise. Use
+    /// simulation with RPC in order to get the exact resources for submitting
+    /// the transactions to the network.
+    pub fn enable_invocation_metering(&self) {
```
leighmcculloch, line 598: What cost does having this enabled incur? Is it noticeable? If it's not material we should enable this by default.

Instead of a function here, I think we should move this into the EnvTestConfig. It'll keep the surface area of testutils functions lower, and is then easily discoverable there. But I'd still have it enabled by default.

soroban-sdk/src/env.rs
```diff
@@ -576,6 +578,91 @@ impl Env {
...
+    /// correspond only to the operations that have happened during the host
+    /// invocation, i.e. this won't try to simulate the work that happens in
+    /// production scenarios (e.g. certain XDR rountrips). This also doesn't try
+    /// to model resources related to the transaction size.
+    ///
+    /// The returned value is as useful as the preceding setup, e.g. if a test
+    /// contract is used instead of a Wasm contract, all the costs related to
+    /// VM instantiation and execution, as well as Wasm reads/rent bumps will be
+    /// missed.    
+    pub fn get_last_invocation_resources(&self) -> Option<InvocationResources> {
```
leighmcculloch, line 618: I'd like to move the functions for getting the resources used into some other type which is available on Env, because otherwise Env becomes a bit of a function soup over time.

We currently have `env.budget()` that returns a testutils `Budget` (defined in this repo, not the same as the env's Budget). Should these functions replace that? Or can we reimagine both together in a single location? Budget getting the raw usage, and this doing something else? From best I can tell looks like this replaces the existing Budget object.

leighmcculloch (commented):
💯 This is _so needed_ and awesome to see getting added to the SDK.

I have some thoughts just about the API and how we integrate it. Lmk what you think, and we can DM / chat too to do a little more realtime.

soroban-sdk/src/env.rs
dmkozh: The cost is not material, but this does change the observed behavior as the budget will be reset in-between invocations. So in theory this might break someone's tests, which is why I was hesitant about enabling this by default (maybe that should have been enabled in the major release, but now we have to wait until the next one).

soroban-sdk/src/env.rs
dmkozh: > From best I can tell looks like this replaces the existing Budget object.

Not quite, `Budget` provides the detailed breakdown of the CPU costs, while this provides a high level overview of all the resources used. I think both have a place in development process.

That said, I agree that we could try converging this functionality at the SDK level, especially given that with this change everything will be describing the last invocation only.

soroban-sdk/src/env.rs
leighmcculloch: > the budget will be reset in-between invocations

I think this is a good thing, we should I think have some consistency across these types of functions where top-level invocation is seen as a single unit, much like how you can get auth for the last invocation.

> maybe that should have been enabled in the major release

+1.

dmkozh: Yeah, I think we're on the same page here, I'd like to enable this by default ASAP, but probably it's safer to wait until the next release. We can make an announcement though and add some examples.

soroban-sdk/src/env.rs
dmkozh: That seems fine, though I'd like to emphasize that these are estimates so it's hopefully not as tempting to use this instead of simulation. So I ended up with `cost_estimate()`.

I've rebased everything from scratch as I was debugging some silly issues, PTAL.

soroban-sdk/src/cost_estimate.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Returns the detailed CPU and memory metering information recorded thus
+    /// far.
+    ///
+    /// The metering resets before every top-level contract level invocation.
+    ///
+    /// Note, that unlike `resources()`/`fee()` this will always return some
+    /// value. If there was no contract call, then the resulting value will
+    /// correspond to metering any environment setup that has been made thus
+    /// far.
+    pub fn detailed_metering(&self) -> Budget {
```
jayz22, line 108: Is this necessary since there is already an `Env::budget()` function

dmkozh: `Env::budget()` can be deprecated in favor of more consistent interface.

soroban-sdk/src/cost_estimate.rs
```diff
@@ -0,0 +1,111 @@
...
+    ///
+    /// Take the return value with a grain of salt as both the resource estimate
+    /// and the fee rates may be imprecise.
+    ///
+    /// The returned value is as useful as the preceding setup, e.g. if a test
+    /// contract is used instead of a Wasm contract, all the costs related to
+    /// VM instantiation and execution, as well as Wasm reads/rent bumps will be
+    /// missed.    
+    pub fn fee(&self) -> Option<FeeEstimate> {
+        // This is a snapshot of the fees as of 2024-12-11.
```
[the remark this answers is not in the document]
dmkozh: I don't think that's necessary. We still need some defaults as setting up the real ledger snapshots for simple unit tests is cumbersome. We could provide a way of overriding this, but I don't see much value in that yet; fees don't change to often and we're in a much better position to update the values in the SDK just once for everyone, then deferring that to the users. We can add the override functionality later if necessary.

soroban-sdk/src/cost_estimate.rs
jayz22: Got it. The budget contains metering info as well as a bunch of auxiliary things (such as the cost schedule and trackers) that the user won't need (and might be confused to equate it with "detailed_metering"), and you can call methods on it. 
For the sake of UX, instead of returning the full `Budget`, it may be better to create a read-only "detailed_metering" type that just contains cpu and mem consumption for various costs, similar what's being shown by Budget::Display.

dmkozh: Maybe it could be better if we designed this from scratch, but currently this would just duplicate the functionality that we already have in budget. Maybe we just should leave budget as is and not try to shove it into the new class, as I agree that it seems kind of odd to have the same exact thing in two different places (and we can't easily get rid of budget either as it would be a breaking change). WDYT @leighmcculloch ?

soroban-sdk/src/cost_estimate.rs
```diff
@@ -0,0 +1,111 @@
+use soroban_env_host::{fees::FeeConfiguration, FeeEstimate, InvocationResources};
```
leighmcculloch, line 1: Can we move this file into the testutils directory? We keep as much as possible in there for testutils only stuff, and everything in that directory will be marked as testutils only in docs so folks don't think it's usable outside tests.

soroban-sdk/src/cost_estimate.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// The metered resources for the last invocation can be retrieved with
+    /// `resources()`, the estimated fee corresponding to these resources can be
+    /// retrieved with `fee()`, and the detailed CPU and memory memory
+    /// breakdown can be retrieved with `detailed_metering()`.
+    ///
+    /// While the resource metering may be useful for contract optimization,
+    /// keep in mind that resource and fee estimation may be imprecise. Use
+    /// simulation with RPC in order to get the exact resources for submitting
+    /// the transactions to the network.    
+    pub fn enable(&self) {
```
leighmcculloch, line 31: We should make this function pub(crate) since it's intended use is via the EnvTestConfig.

soroban-sdk/src/cost_estimate.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// correspond only to the operations that have happened during the host
+    /// invocation, i.e. this won't try to simulate the work that happens in
+    /// production scenarios (e.g. certain XDR rountrips). This also doesn't try
+    /// to model resources related to the transaction size.
+    ///
+    /// The returned value is as useful as the preceding setup, e.g. if a test
+    /// contract is used instead of a Wasm contract, all the costs related to
+    /// VM instantiation and execution, as well as Wasm reads/rent bumps will be
+    /// missed.    
+    pub fn resources(&self) -> Option<InvocationResources> {
```
leighmcculloch, line 51: I think we should return the value, and panic on incorrect use.

Why: Convenience, and this is in tests, so the convenience of not having to unwrap something all the time is a big benefit. If someone makes a mistake it'll be in a test and they'll be able to fix it straight away. The panic can tell them what needs enabling exactly.

soroban-sdk/src/cost_estimate.rs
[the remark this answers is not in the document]
leighmcculloch: +1 that might be necessary in the future, but isn't necessary today. 

Same comment as above about the Option<> return value, I think we should unwrap and panic.

soroban-sdk/src/lib.rs
```diff
@@ -802,6 +802,8 @@ pub mod data {
 }
 pub mod auth;
 mod bytes;
+#[cfg(any(test, feature = "testutils"))]
+pub mod cost_estimate;
```
leighmcculloch, line 806: Move this into the testutils directory and you wo'nt need a cfg testutils here.

Also note that whenever we cfg testutils, we also need to cfg_attr a doc attribute to tell the docs generation that it's only usable in testutils. You won't need to do that if you move it into the testutils directory, but just a knowledge share.

soroban-sdk/test_wasms/README.md
```diff
@@ -0,0 +1,6 @@
+# test_wasms
+
+Files contained in this directory are used in a few SDK tests that are sensitive
+to Wasm content changes.
+
+`test_contract_data.wasm` is a build of `contract_data`  test contract.
```
leighmcculloch, line 6: We don't need to prebuild the wasm and commit it. All of CI runs `make build` before the tests, and we have other tests that read wasms straight out of the target directory, so should follow suit.

soroban-sdk/src/cost_estimate.rs
leighmcculloch: +1 we can deprecate Env::budget() in this PR and say to use the new interface. It's helpful to group these together into a single location to look at this kind of stuff.

leighmcculloch (commented):
👏🏻  Love the level of docs we're adding around these functions.

soroban-sdk/Cargo.toml
```diff
@@ -53,6 +53,7 @@ derive_arbitrary = { version = "~1.3.0" }
 proptest = "1.2.0"
 proptest-arbitrary-interop = "0.1.0"
 libfuzzer-sys = "0.4.7"
+expect-test = "1.4.1"
```
leighmcculloch, line 56: 👏🏻 This is pretty cool, love it's use here, seems very appropriate for these types of test expectations.

soroban-sdk/src/env.rs
leighmcculloch: > So I ended up with `cost_estimate()`.

👍🏻 I think that's wise to call out their estimates.

soroban-sdk/test_wasms/README.md
dmkozh: We do need to store this Wasm for the sake of the test stability. CI already manages to build Wasm that slightly differs in size on different platforms, and in the future any SDK change would lead to this test breaking. The intention of this test not to cover how SDK builds Wasm, but rather that SDK uses the env functionality properly, so I think storing a prebuilt Wasm will save us a lot of maintenance pain (again, not mention that it would require some hacks to make it pass today). The alternative is to not cover the exact numbers at all, but I don't really like that - the test is a nice demonstration of what things are metered.

soroban-sdk/src/cost_estimate.rs
dmkozh: Ok, I've added a deprecation notice. Still not sure about the naming. Maybe `budget` is more appropriate, as it does have some budget-related functionality. I don't have a strong preference here though.

soroban-sdk/src/cost_estimate.rs
dmkozh: Good point, done.

soroban-sdk/src/cost_estimate.rs
dmkozh: Agree, updated. FWIW we only need to panic here, as `fee` just uses this.

soroban-sdk/src/lib.rs
dmkozh: Acknowledged.

soroban-sdk/src/cost_estimate.rs
dmkozh: I've tried that, however the issue is that we allow updating the config without resetting the env via `set_config`. The host function we now is one way (I didn't anticipate this ever being disabled after being enabled). So we can't provide appropriate config behavior right now. I see several options here:

- Leave things as is in this PR (i.e. enable via `cost_estimate`) directly
- Updating the host to also allow disabling this. This might require some semver wrestling though, as this is a public function.
- Maybe just enable this by default after all. The scope of potential breakage seems really narrow (observing the budget and asserting on exact values). We also don't intentionally try to avoid this sort of breakages in general, as any internal setup change might lead to the changes in the observed budget. I realize that here the change is more explicit than e.g. updating the implementation of instantiation function, but likely the benefits (better usability, less code, less cleanup after next release) outweigh the (improbable) breakage risks.

soroban-sdk/test_wasms/README.md
leighmcculloch: Right, but we already have other tests that make this assumption, so we're making this test work differently to the other tests in this repo. Let's address that problem in the future when it becomes a problem and address it holistically?\

leighmcculloch: Or are you saying we already have that issue right now with this test?

soroban-sdk/src/lib.rs
```diff
@@ -802,6 +802,7 @@ pub mod data {
 }
 pub mod auth;
 mod bytes;
+
```
leighmcculloch, line 805:
```suggestion
```

soroban-sdk/src/cost_estimate.rs
leighmcculloch: Hmm. I think we can do this:

If set_config is called with turning the value off, when the value is already turned on (should be an easy comparison), panic with an informative message saying that once it is enabled it cannot be disabled.

This addresses the 90% case, people will turn it on, and leave it on.

And a panic here, is fine.

leighmcculloch (commented):
See the suggestion inline about how to deal with set_config.

soroban-sdk/src/testutils/cost_estimate.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///
+    /// The returned value is as useful as the preceding setup, e.g. if a test
+    /// contract is used instead of a Wasm contract, all the costs related to
+    /// VM instantiation and execution, as well as Wasm reads/rent bumps will be
+    /// missed.    
+    pub fn resources(&self) -> InvocationResources {
+        if let Some(res) = self.env.host().get_last_invocation_resources() {
+            res
+        } else {
+            panic!("Invocation cost estimate is not available. Make sure invocation cost metering is enabled and this is called after an invocation.")
```
leighmcculloch, line 51: Suggest mentioning EnvTestConfig once we add that.
```suggestion
            panic!("Invocation cost estimate is not available. Make sure invocation cost metering is enabled in the EnvTestConfig and this is called after an invocation.")
```

soroban-sdk/test_wasms/README.md
dmkozh: Yes, we do have divergence right now. I'm not sure if fixing this is the right use of time - even if Wasm was consistent between all the build configurations, it's still annoying that unrelated changes might require updating this test (I'd prefer it only to need updates when the protocol and thus metering changes - this is how the env tests behave and this feature is unfortunately pretty tightly coupled with the env internals).

dmkozh:
> If set_config is called with turning the value off, when the value is already turned on (should be an easy comparison), panic with an informative message saying that once it is enabled it cannot be disabled.

That makes sense and I've tried to sketch this. But this doesn't pass the semver check, as we're adding a new pub field to a struct, so initialization will be broken. So I'm even more in favor of enabling this by default.

soroban-sdk/src/env.rs
```diff
@@ -254,15 +254,28 @@ struct EnvTestState {
 pub struct EnvTestConfig {
     /// Capture a test snapshot when the Env is dropped, causing a test snapshot
     /// JSON file to be written to disk when the Env is no longer referenced.
-    /// Defaults to true.
+    /// Defaults to `true``.
```
leighmcculloch, line 257:
```suggestion
    /// Defaults to `true`.
```

soroban-sdk/src/env.rs
```diff
@@ -254,15 +254,28 @@ struct EnvTestState {
...
+    ///
+    /// The top-level contract invocations and lifecycle operations (such as
+    /// `register` or `env.deployer()` operations) will be metered and
+    /// all the metering information will reset in-between them (note, that this
+    /// includes the CPU budget). Metering will not be reset while inside the
+    /// call (e.g. if a contract calls or creates another contract, that won't
+    /// reset metering).
+    ///     
+    /// Defaults to `false`.
+    pub enable_invocation_cost_estimate: bool,
```
leighmcculloch, line 270: I think this can be shortened to:
```suggestion
    pub enable_cost_estimate: bool,
```
The average developer won't distinguish between an invocation cost estimate vs another type of cost estimate.

leighmcculloch:
> this doesn't pass the semver check, as we're adding a new pub field to a struct

Darn, I forgot about this when I created this struct. In v23 we should change the struct into private fields with functions to modify the state. I created an issue to do that here:
- https://github.com/stellar/rs-soroban-sdk/issues/1419

> So I'm even more in favor of enabling this by default.

👍🏻

soroban-sdk/test_wasms/README.md
leighmcculloch: 👍🏻

dmkozh:
Based on the previous discussion I've enabled this by default, PTAL. 

One caveat that I forgot about is that besides budget we also reset all the events between the calls. This makes sense in general and may actually highlight issues people might have had (if an event was expected to happen after a wrong contract call).

leighmcculloch:
I think it's reasonable to treat that like a bug fix. The way it works today is too unspecific as to when events occur.

soroban-sdk/test_wasms/README.md
leighmcculloch, line 6:
```suggestion
`test_contract_data.wasm` is a build of `tests/contract_data`  test contract.
```

leighmcculloch (approved):

PR #1412, opened 2024-12-12
Preserve the recording auth settings when temporarily switching to recording auth.

dmkozh:
### What

Preserve the recording auth settings when temporarily switching to recording auth.

### Why

Previously this didn't matter as we've only switched for the contract instantiation, but now the contract instantiation may run custom logic due to constructors, so we need to preserve the settings (currently only disallowing non-root auth).

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -965,7 +965,9 @@ impl Env {
             .unwrap();
 
         let prev_auth_manager = self.env_impl.snapshot_auth_manager().unwrap();
-        self.env_impl.switch_to_recording_auth(true).unwrap();
+        self.env_impl
+            .switch_to_recording_auth_inherited_from_snapshot(&prev_auth_manager)
+            .unwrap();
         self.invoke_contract::<()>(
             &token_id,
             &soroban_sdk_macros::internal_symbol_short!("set_admin"),
@@ -1014,7 +1016,9 @@ impl Env {
         constructor_args: Vec<Val>,
     ) -> Address {
         let prev_auth_manager = self.env_impl.snapshot_auth_manager().unwrap();
-        self.env_impl.switch_to_recording_auth(true).unwrap();
+        self.env_impl
+            .switch_to_recording_auth_inherited_from_snapshot(&prev_auth_manager)
+            .unwrap();
         let args_vec: std::vec::Vec<xdr::ScVal> =
             constructor_args.iter().map(|v| v.into_val(self)).collect();
         let contract_id: Address = self


soroban-sdk/src/env.rs
```diff
@@ -965,7 +965,9 @@ impl Env {
             .unwrap();
 
         let prev_auth_manager = self.env_impl.snapshot_auth_manager().unwrap();
-        self.env_impl.switch_to_recording_auth(true).unwrap();
+        self.env_impl
+            .switch_to_recording_auth_inherited_from_snapshot(&prev_auth_manager)
```
leighmcculloch, line 969: What does it mean that it's inherited? Or, why do we need it to be inherited? In this instance we're intending to invoke the SAC contract without any existing auth setup bleeding into it, because it's its own top-level invocation.

dmkozh: The settings of the recording mode are inherited (currently only non-root auth setting). Conceptually these probably could have been an env property and not the auth manager property, but that would be a bigger change. While SAC registration currently doesn't care about this setting, I think it's generally more appropriate to just use the same pattern everywhere.

leighmcculloch: I think the auth setting being inherited is a bug, but if it's impractical to fix that bug, then this does make the presence of the bug clearer.

leighmcculloch (dismissed):
I think the auth setting being inherited is a bug, but if it's impractical to fix that bug, then this does make the presence of the bug clearer.

soroban-sdk/src/env.rs
leighmcculloch: Wait wait, so does the old `switch_to_recording_auth(true)` cause inheritance too? Are these two functions doing the same thing? Because if `switch_to_recording_auth(true)` doesn't inherit, that seems like the better thing to happen here.

dmkozh: > I think the auth setting being inherited is a bug,

It's not a bug, it's the intended behavior. Users should be able to generically change how the recording auth operates. Otherwise we just have unpreventable errors due to non-root auth in constructor.

> Wait wait, so does the old switch_to_recording_auth(true) cause inheritance too?

No, it overrides any setting the user had on the setting, which is a bug. Again, we can't know what is the user intention here and we so we shouldn't override it.

dmkozh: https://discord.com/channels/897514728459468821/1315694921054945340 for some context

leighmcculloch: Thanks for the context. I saw this comment in the discussion:

> basically if you call `register`, we accidentally override your setting for allowing non-root auth.

Ref: https://discord.com/channels/897514728459468821/1315694921054945340/1316177239477456969

If I understand correctly:
- 1️⃣ The developer configured non-root auth.
- 2️⃣ They then called register, which:
   - 3️⃣ took a snapshot of auth
   - 4️⃣ switched to recording auth 
   - 5️⃣ switched back to the snapshot 

Is the bug at the line item where 👈🏻 is?

It seems like this PR is fixing the bug by making 4️⃣ keep the non-root auth so that while auth is snapshotted, cleared, and then restored, some auth config is not cleared then restored, it just persists throughout the process. Is that correct?

Imo we should be fixing the issue in 3️⃣ and 5️⃣. The auth snapshot is supposed to snapshot and allow restoring the full auth state, so that steps in between, i.e. 4️⃣, can work with a completely neutral state that's unaffected by any caller configuration.

Why fix it at 4️⃣?

leighmcculloch: I think implementation decisions (e.g. auth manager's scope) are influencing the API.

If a user is using Env without non-root auth enabled, but we have an Env function that needs to temporarily enable it, perform some custom auth, then disable it again, is that possible with the direction of this fix?

dmkozh: > we have an Env function that needs to temporarily enable it,

We never *need* to enable this setting. This setting has nothing to do with the business logic of deployment function, it's purely a user decision as to whether enable it to prevent footguns or not (when they know what they're doing). It's only reasonable for the SDK to suggest a safer default (i.e. enable the setting).

leighmcculloch (approved):

PR #1416, opened 2024-12-13
Improve deploy docs

leighmcculloch:
### What
Remove the `let _ = deployer.deployed_address();` line from the two existing deploy examples that are in the rust docs, and add an example that discusses the `deployer.deployed_address()` in more detail.

### Why
Plopping the `let _ = deployer.deployed_address();` into the middle of an example that's about a specific scenario is distracting as to the main purpose of the example. The docs can show how to use the function in an example focused on it specifically.


dmkozh (approved):
Thanks for fixing this!

PR #1427, opened 2025-01-21
Include the token interface in the stellar asset interface

leighmcculloch:
### What
Include the token interface in the stellar asset interface, and therefore also in the client.

### Why
The separation of the token functions from the stellar asset interface and client was originally done to ensure that developers didn't use the stellar asset client in situations where they really should be using the token client, to interface with tokens of any kind.

### DIFF
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -80,7 +80,7 @@ pub use TokenClient as Client;
 /// There are no functions in the token interface for minting tokens. Minting is
 /// an administrative function that can differ significantly from one token to
 /// the next.
-#[contractspecfn(name = "StellarAssetSpec", export = false)]
+#[contractspecfn(name = "TokenSpec", export = false)]
 #[contractclient(crate_path = "crate", name = "TokenClient")]
 pub trait TokenInterface {
     /// Returns the allowance for `spender` to transfer from `from`.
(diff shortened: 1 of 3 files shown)


leighmcculloch:
There's a semver failure on the StellarAsset trait changing, but I'm unconvinced we should treat it like a breakage. The trait is very likely not implemented, unless someone is implementing it and the token trait to mirror the Stellar Asset Contract.

@dmkozh thoughts?

dmkozh:
> There's a semver failure on the StellarAsset trait changing, but I'm unconvinced we should treat it like a breakage. The trait is very likely not implemented, unless someone is implementing it and the token trait to mirror the Stellar Asset Contract.

Yeah, I don't think anyone should be actually implementing StellarAsset trait.

soroban-sdk/src/token.rs
```diff
@@ -80,7 +80,7 @@ pub use TokenClient as Client;
 /// There are no functions in the token interface for minting tokens. Minting is
 /// an administrative function that can differ significantly from one token to
 /// the next.
-#[contractspecfn(name = "StellarAssetSpec", export = false)]
+#[contractspecfn(name = "TokenSpec", export = false)]
```
dmkozh, line 83: Why does token spec have `approve` and `burn_from` functions, but not `transfer_from`?

leighmcculloch: In the `TokenSpec` trait:
- The `approve` function is on line 121.
- The `burn_from` function is on line 209.
- The `transfer_from` function is on line 168.

dmkozh: Sorry, apparently I haven't looked at the full fille, that's embarrassing.

dmkozh (approved):

leighmcculloch:
I'm happy to hold this for protocol 23. This area of code is unlikely to get some churn. And it's the conservative approach given the semver. Also, the build won't let me merge while there's a semver breakage.

PR #1429, opened 2025-01-22
Limit attributes copied into generated code

### LINKED ISSUE #1426: `#[contractimpl]` also copies the procedural macros for the helper functions
### What did you do?

Tried to annotate a smart contract function with a custom macro, and it failed miserably due to how `#[contractimpl]` macro expands.

More background: we are writing a `pausable` extension for smart contracts, and we wanted to be able to annotate the functions with `#[when_paused]` and `#[when_not_paused]` instead of polluting the function body. 

Our macro requires the first parameter of the annotated function to be of type `&Env` or `Env`. 

And here is the relevant snippet of our code:
```rust
#[contractimpl]
impl ExampleContract {
    pub fn __constructor(e: Env, owner: Address) {
        e.storage().instance().set(&DataKey::Owner, &owner);
        e.storage().instance().set(&DataKey::Counter, &0);
    }
    
    /* redacted */

    #[when_paused] // HERE!
    pub fn emergency_reset(e: Env) {
        e.storage().instance().set(&DataKey::Counter, &0);
    }
}
```

So far so good... 

However, when this macro is put to action inside `#[contractimpl]`, the below errors occur. 

Image

I was able to expand the `#[contractimpl]` macro and see the root of the error. Below are the relevant snippets that is generated by the `#[contractimpl]` expansion (helper functions on related to `emergency_reset` function):

1.
```rust
    impl ExampleContract {
        #[allow(non_snake_case)]
        pub const fn spec_xdr_emergency_reset() -> [u8; 36usize] {
            *b"\0\0\0\0\0\0\0\0\0\0\0\x0femergency_reset\0\0\0\0\0\0\0\0\0"
        }
    }
```
2.
```rust
        pub fn emergency_reset(&self) -> () {
            use core::ops::Not;
            use soroban_sdk::{IntoVal, FromVal};
            let res = self.env.invoke_contract(
                &self.address,
                &{ soroban_sdk::Symbol::new(&self.env, "emergency_reset") },
                ::soroban_sdk::Vec::new(&self.env),
            );
            res
        }
```
3.
```rust
        pub fn try_emergency_reset(
            &self,
        ) -> Result<
            Result<(), <() as soroban_sdk::TryFromVal<soroban_sdk::Env, soroban_sdk::Val>>::Error>,
            Result<soroban_sdk::Error, soroban_sdk::InvokeError>,
        > {
            use soroban_sdk::{IntoVal, FromVal};
            let res = self.env.try_invoke_contract(
                &self.address,
                &{ soroban_sdk::Symbol::new(&self.env, "emergency_reset") },
                ::soroban_sdk::Vec::new(&self.env),
            );
            res
        }
```

### Problem
I think I know the problem already (I've dealt with procedural macros myself). `#[contractimpl]` macro copies the function name, and creates the necessary helper functions. However, during the process, it keeps the macro annotations for these copies as well. And I think, no macros should be preserved for these copies. 

### Solution
`#[contractimpl]` macro should not keep the macros that are annotating the function for the generated helper functions, it should only preserve the macro for the original function.

leighmcculloch:
### What
Limit the attributes that get copied into code generated to:
- `doc` - rust docs
- `cfg` - used to control whether functions are included or not depending on features
- `allow`/`deny` - used to ignore or require compiler warnings

### Why
The copying of attributes into generated code is from early code developed for the SDK where we were trying to deal with functions that weren't named consistently with Rust's naming scheme. We wanted to carry through to the generated code `allow` statements so that the generated code didn't generate warnings that were ignored at the source level.

However, this has introduced a problem. For any dev who is introducing macros into their contracts and where those macros appear as attributes on contract functions, the macros are then being duplicated into the generated code and injected onto the contract client functions, and modules.

Docs attributes are retained because it's valuable if docs show up when hovering over generated contract client functions. The value there is that when someone hovers over a call to a contract client function in an IDE, it'll display the docs for that function.

Cfg attributes are retained because folks might put a contract function behind a feature, or behind the test flag, so that the function is only present when that feature is enabled, or tests are enabled.

Allow and deny are added because folks might be allowing or denying non-std function names and it'd benefit to pass those allows down to the client at least.

Close #1426

### Known Limitations

A limitation with this change, and a risk, is that it is a breaking change that is very difficult to pick up on. Someone may be attaching an attribute to their contract function and expecting it to carry over to the client. After this change is merged unless there is an error compiling or an easily observable change in behaviour, the developer may not know that the attribute no longer applies to the client. I think the risk is lowered because it's unexpected and not normal that the attributes are copied over, so unlikely someone knows they could depend on it, and it's undocumented and not a difficult "feature" to discover. Any use of this kind is likely to be accidental.

We could keep this change for the v23 major version to reduce risk, although this change is actually a bug fix blocking development, so I think we should treat it as a bug fix and merge it into a v22 patch release.

### DIFF
--- a/soroban-sdk-macros/src/attribute.rs
+++ b/soroban-sdk-macros/src/attribute.rs
@@ -1,0 +1,10 @@
+use syn::Attribute;
+
+/// Returns true if the attribute is an attribute that should be preserved and
+/// passed through to code generated for the item the attribute is on.
+pub fn pass_through_attr_to_gen_code(attr: &Attribute) -> bool {
+    attr.path().is_ident("doc")
+        || attr.path().is_ident("cfg")
+        || attr.path().is_ident("allow")
+        || attr.path().is_ident("deny")
+}
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -197,7 +197,11 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
                 .unzip();
             let fn_output = f.output();
             let fn_try_output = f.try_output(crate_path);
-            let fn_attrs = f.attrs;
+            let fn_attrs = f
+                .attrs
+                .iter()
+                .filter(|attr| attr.path().is_ident("doc"))
+                .collect::<Vec<_>>();
             if cfg!(not(feature = "testutils")) {
                 quote! {
                     #(#fn_attrs)*
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -2,7 +2,7 @@ use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
 use syn::{Error, FnArg, LitStr, Path, Type, TypePath, TypeReference};
 
-use crate::{symbol, syn_ext};
+use crate::{attribute::pass_through_attr_to_gen_code, symbol, syn_ext};
 
 pub fn derive_client_type(crate_path: &Path, ty: &str, name: &str) -> TokenStream {
     let ty_str = quote!(#ty).to_string();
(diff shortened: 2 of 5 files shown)


ozgunozerk:
Works! Thanks for this rapid implementation 🚀

tomerweller (approved):

PR #1441, opened 2025-02-24
Muxed address support sketch for Soroban SDK.

dmkozh:
This introduces the `MuxedAddress` type that maps to either `AddressObject` or `MuxedAddressObject`.

In order to not 'leak' the muxed addresses everywhere (as they have pretty narrow use cases), introduce a separate token interface that supports muxed addresses for transfer (could also support e.g. transfer_from). This is basically token extension, but a bit clunkier due to the lack of overload support. The amount of copy-paste could be reduced though with some code generation (I'm not sure that's necessary though). Also only expose muxed addresses to generate clients when that's explicitly requested, as again users normally don't need to worry about them.

### DIFF
--- a/soroban-sdk/src/string.rs
+++ b/soroban-sdk/src/string.rs
@@ -65,10 +65,6 @@ impl PartialOrd for String {
 
 impl Ord for String {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        #[cfg(not(target_family = "wasm"))]
-        if !self.env.is_same_env(&other.env) {
-            return ScVal::from(self).cmp(&ScVal::from(other));
-        }
         let v = self
             .env
             .obj_cmp(self.obj.to_val(), other.obj.to_val())
--- a/soroban-token-sdk/src/event.rs
+++ b/soroban-token-sdk/src/event.rs
@@ -1,8 +1,14 @@
-use soroban_sdk::{symbol_short, Address, Env, Symbol};
+use soroban_sdk::{symbol_short, Address, Env, IntoVal, MuxedAddress, Symbol, Val};
 
 pub struct Events {
     env: Env,
 }
+
+mod sealed {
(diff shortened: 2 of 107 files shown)


soroban-sdk/src/string.rs
```diff
@@ -65,10 +65,6 @@ impl PartialOrd for String {
 
 impl Ord for String {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        #[cfg(not(target_family = "wasm"))]
-        if !self.env.is_same_env(&other.env) {
-            return ScVal::from(self).cmp(&ScVal::from(other));
-        }
```
leighmcculloch, line 71: Why are the comparison functions changing?

soroban-token-sdk/src/event.rs
```diff
@@ -17,7 +23,12 @@ impl Events {
...
 
-    pub fn transfer(&self, from: Address, to: Address, amount: i128) {
+    pub fn transfer<AddressType: sealed::IsAddressType + IntoVal<Env, Val>>(
+        &self,
+        from: AddressType,
+        to: AddressType,
+        amount: i128,
+    ) {
         let topics = (symbol_short!("transfer"), from, to);
         self.env.events().publish(topics, amount);
```
leighmcculloch, line 33: Looks like the muxed address is emitted in the topic, so it wouldn't be very easy for a consumer to filter only on the underlying address.

soroban-sdk/src/string.rs
dmkozh: That's unrelated to the sketch, I just didn't want to refactor the SDK to support https://github.com/stellar/rs-soroban-env/pull/1516.

leighmcculloch:
I think this was superseded by:
- https://github.com/stellar/rs-soroban-sdk/pull/1454

PR #1442, opened 2025-02-25
Fix BytesN to ScVal conversion in testutils

leighmcculloch:
### What
Change BytesN to ScVal conversion from TryFrom to From in test utilities.

### Why
Unwrapping is safe because values in test environment must be valid ScVal. Even if this was wrong, it's a good tradeoff that improves the developer experience. The Bytes->ScVal and String->ScVal conversions already use this pattern and it seems BytesN wasn't updated at the same time to use that pattern.


sisuresh (approved):

leighmcculloch:
Build failure being fixed by:
- https://github.com/stellar/rs-soroban-sdk/pull/1445

PR #1447, opened 2025-02-26
Muxed address support sketch for Soroban SDK, `transfer_muxed` edition

dmkozh:
This just uses a separate interface/contract function that accepts `MuxedAddress` and doesn't introduce any opt-ins/outs of exposing the muxed address type.

### DIFF
--- a/soroban-spec-rust/src/types.rs
+++ b/soroban-spec-rust/src/types.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Literal, TokenStream};
 use quote::{format_ident, quote};
-use stellar_xdr::curr as stellar_xdr;
+use stellar_xdr::next as stellar_xdr;
 use stellar_xdr::{
     ScSpecTypeDef, ScSpecUdtEnumV0, ScSpecUdtErrorEnumV0, ScSpecUdtStructV0, ScSpecUdtUnionV0,
 };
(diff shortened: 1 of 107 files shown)


soroban-spec-rust/src/types.rs
```diff
@@ -148,6 +148,13 @@ pub fn generate_type_ident(spec: &ScSpecTypeDef) -> TokenStream {
         ScSpecTypeDef::Error => quote! { soroban_sdk::Error },
         ScSpecTypeDef::Bytes => quote! { soroban_sdk::Bytes },
         ScSpecTypeDef::Address => quote! { soroban_sdk::Address },
+        ScSpecTypeDef::AddressV2(supports_muxing) => {
+            if *supports_muxing {
+                quote! { soroban_sdk::MuxedAddress }
+            } else {
+                quote! { soroban_sdk::Address }
+            }
```
leighmcculloch, line 156: What's the benefit of AddressV2 which is configurable, vs introducing a `MuxedAddress` that maps directly to the `MuxedAddress` in the SDK?

dmkozh: Yeah, it could be `MuxedAddress`, with a caveat that it would be the only spec type that's seamlessly compatible with another spec type (`Address`). The option here tries to convey that both are semantically the same type, but with some additional options available. Both options are a bit weird; probably support for the variant types in the spec would be more clear than either one. But for a one-off exception I think both options are ok-ish (not pretty, but ultimately not too hard to deal with when parsing the spec).

leighmcculloch: > it would be the only spec type that's seamlessly compatible with another spec type

The spec types already have some oddities that cause them to overlap with other spec types. For example, The Option<> spec type overlaps with Void and every other value.

leighmcculloch:
I think this was superseded by:
- https://github.com/stellar/rs-soroban-sdk/pull/1454

Issue #1330, opened 2024-09-09
Support referencing and re-exporting types from other libs for type sharing


leighmcculloch:
### What problem does your feature solve?

It's desirable to be able to export types from a contract that have been defined in another crate or contract.

See discussions about doing that here:
- https://discord.com/channels/897514728459468821/1281696488199553025

### What would you like to see?

Ideally a way to use a type in a contract functions interface, where that type is defined in another crate or another contract.

### What alternatives are there?

A manual way to redefine the same type, as a type-alias of sorts, and somehow adopt the structure of the type defined elsewhere and be compatible.

leighmcculloch:
This issue is similar and related to:
- https://github.com/stellar/rs-soroban-sdk/issues/1326

Issue #1063, opened 2023-08-18
Cannot use shared code between contracts


karol-bisztyga:
### What did you do?

I have a project that uses rust workspaces (each contract is a separate workspace member).
I want to share some code between them, mostly types.

repro: here is a repro https://github.com/karol-bisztyga/soroban-test

The problem is, it builds and deploys but when I invoke the contract I get `MissingEntry("DataType")` error. When I replace the type with just `soroban_sdk::String`, it works, but when I get the type from the shared crate, it breaks (it's still just a string just aliased). Looks like during the build or deployment the imported code's not attached to the contract.

To reproduce (more or less):
```
git clone https://github.com/karol-bisztyga/soroban-test.git
cd soroban-test
docker run --rm -it -d -p 8000:8000 --name stellar stellar/quickstart:soroban-dev@sha256:ed57f7a7683e3568ae401f5c6e93341a9f77d8ad41191bf752944d7898981e0c --standalone --enable-soroban-rpc
./run.sh
```

### What did you expect to see?

Working invoke.

### What did you see instead?

`MissingEntry("DataType")` error.

Please let me know if it's the right place to post this, if not, pls point me to the right one.

leighmcculloch:
Transferred this issue from stellar/soroban-tools#870 to stellar/rs-soroban-sdk#1063 because the limitation preventing this from working is in the SDK. I believe the contract spec is extendable to address this problem. 👀

leighmcculloch:
From best I can tell with the experiment I ran, types can live in other libraries and be imported. See https://github.com/stellar/rs-soroban-sdk/pull/1070 for an example of it working.

I think the issue that's occurring in the example above is that a type alias is being used. Soroban contracttype's don't support a type alias.

We could add support for it, although it would be somewhat noisy in contract interfaces. Contract interfaces would have to store the original type, and the alias type. So the contract will be smaller and cheaper to execute if the alias was ommitted. If it's inefficient, folks will probably avoid it, so it's questionable if we should add logic and complexity to handle it.

Thoughts anyone? @stellar/contract-committers?

leighmcculloch:
As a another data point for evaluating adding this feature, this feature was requested in Discord:
- https://discord.com/channels/897514728459468821/1237832562139988090

leighmcculloch:
[...]

PR #1449, opened 2025-03-06
Implement BLS12-381 `contracttype` support

jayz22:
### What

Add support for BLS12-381 curve points (`Fp`, `Fp2`, `G1Affine`, `G2Affine`, `Fr`) for internal data storage (via `contracttype`) and contract invocation arguments (contract spec). 

### Why

Improves usability of the BLS12-381 features in a Groth16 verifier application ([example contract](https://github.com/stellar/soroban-examples/pull/350))
-  `Arbitrary` and several missing conversions are needed to use `G1Affine`, `G2Affine` as contract data.

### Known limitations

[TODO or N/A]


soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -115,6 +217,14 @@ impl G1Affine {
     }
 }
 
+// impl TryFromVal<Env, Val> for G1Affine {
+//     type Error = ConversionError;
+
+//     fn try_from_val(env: &Env, v: &Val) -> Result<Self, Self::Error> {
+//         Ok(G1Affine(BytesN::try_from_val(env, v)?))
+//     }
+// }
```
leighmcculloch, line 226: ?

soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -7,15 +7,66 @@ use crate::{
...
+            slice[i * 8..(i + 1) * 8].copy_from_slice(&limb_bytes);
+        }
+    }
+}
+
+impl<const N: usize, const M: usize> Into<BigInt<N>> for BytesN<M> {
+    fn into(self) -> BigInt<N> {
+        if M != N * 8 {
+            sdk_panic!("BytesN::Into<BigInt> - length mismatch")
+        }
```
leighmcculloch, line 55: The `M != N*8` check occurs at runtime. We should perform this check at compile time so that devs find out earlier in the dev cycle that they've written code that will only fail.

Since [Rust 1.79](https://doc.rust-lang.org/beta/releases.html#version-1790-2024-06-13) we can do that using the `const { /* ... */ }` expression. The SDK requires at least Rust 1.79, so it is safe to use it in the SDK.

The const block will cause everything in it to be evaluated at build time, and the panic will cause a compile error.

```suggestion
        const {
            if M != N * 8 {
                sdk_panic!("BytesN::Into<BigInt> - length mismatch")
            }
        }
```

The other change we should make is that `sdk_panic!` is only useful in runtime contexts. The macro makes sure that the text given in the panic message is not stored in the wasm for use at runtime. In a const / compile time context using `panic!` has no cost, and is better, because it will mean that even when building with a wasm target will the panic text be outputted in the compile error.

So this is what I'd suggest.

```suggestion
        const {
            if M != N * 8 {
                panic!("BytesN::Into<BigInt> - length mismatch")
            }
        }
```

We should use this pattern in the other places where the values of N and M are validated.

soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -26,7 +77,7 @@ pub struct Bls12_381 {
 ///   `Fp`
 /// - The most significant three bits (bits 0-3) of the first byte are reserved
 ///   for encoding flags:
-///   - compression_flag (bit 0): Must always be set (1), as only uncompressed
+///   - compression_flag (bit 0): Must always be unset (0), as only uncompressed
```
leighmcculloch, line 80: Is this behaviour that's changing that's related to this change, or a bug in documentation? If latter, and since this looks isolated from everything else here, could you make this fix in a separate PR? so that it is captured clearly in the changelog as its own line item, and in general to increase the visibility.

soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -101,6 +152,57 @@ impl_bytesn_repr!(G2Affine, 192);
 impl_bytesn_repr!(Fp, 48);
 impl_bytesn_repr!(Fp2, 96);
 
+impl Fp {
+    pub fn env(&self) -> &Env {
+        self.0.env()
+    }
+
+    pub fn map_to_g1(&self) -> G1Affine {
```
leighmcculloch, line 160: It looks like this method converts an `Fp` to a `G1Affine`. Is that accurate? Or is there semantics about the `map` part of it that represent more than a conversion from one type to the next?

Rust naming conventions suggest using `into`, `as_`, and `to_` for type conversions, so it'd be preferred to follow that naming convention unless this shouldn't be thought of as a conversion. The most appropriate for this one is probably `to_g1` because it's a borrowed to owned. See https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv.

```suggestion
    pub fn to_g1(&self) -> G1Affine {
```

soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -101,6 +152,57 @@ impl_bytesn_repr!(G2Affine, 192);
...
+impl Neg for Fp {
+    type Output = Fp;
+
+    fn neg(self) -> Self::Output {
+        if self.to_array() == [0; 48] {
+            return self;
+        }
+
+        let env = self.env().clone();
+        let fp_bigint: BigInt<6> = self.0.into();
```
leighmcculloch, line 188: `to_array` contains a host call to copy the memory to the guest, and when using `neg` that host call and copy is occurring twice, because it occurs at the beginning of the neg, and then inside the `Into<BigInt<6>>::into`.

It would be ideal to avoid the additional host call and host-guest memcpy.

You could add a function `zero` to `BigInt<6>` and do the zero check after converting into BigInt. The conversion for zero will be more expensive, but is zero usage an edge case? That tradeoff might be more appropriate if it is, so we don't hit the double memcpy for the general case.

leighmcculloch (commented):
Thanks for making these types usable with contracttype 👏🏻.

This pull request looks like it contains a bundle of changes related only by topic, and otherwise independent. It's helpful if it's split, both for changelog capturing, visibility, and readers. I understand that because the example (https://github.com/stellar/soroban-examples/pull/350) was being developed at the same time that also might not be desirable. No need to do now if too inconvenient, just fyi as the reader initially felt like things were related and had to piece through what was and wasn't, and intent for each change.

I've left a few comments inline to do with rust conventions, and changing a runtime check to a compile time check.

For the low-level BLS operations, I'm not the best person to review, so I'm tagging @stellar/contract-committers to review.

jayz22:
@leighmcculloch I've split out this PR into three parts
1. https://github.com/stellar/rs-soroban-sdk/pull/1455 contains the comment fix
2. https://github.com/stellar/rs-soroban-sdk/pull/1456 contains the implementation of point negation
3. this PR just contains the `contracttype` support

they are currently based on top of one another, so they should be merged in the above order. I will mention and address your review comments pertaining to the those changes in the their corresponding PR.

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -680,6 +681,64 @@ mod objects {
...
+    impl SorobanArbitrary for G1Affine {
+        type Prototype = ArbitraryG1Affine;
+    }
+
+    impl TryFromVal<Env, ArbitraryG1Affine> for G1Affine {
+        type Error = ConversionError;
+
+        fn try_from_val(env: &Env, v: &ArbitraryG1Affine) -> Result<Self, Self::Error> {
+            let mut bytes = v.bytes;
+            bytes[0] &= 0b0001_1111; // Clear the top 3 bits
```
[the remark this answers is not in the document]
jayz22: added comments in the code.

soroban-sdk/src/crypto/bls12_381.rs
jayz22: removed

soroban-sdk/src/crypto/bls12_381.rs
jayz22: great suggestion, and good to know about the const expression feature! 
fixed it in the other PR [here](https://github.com/stellar/rs-soroban-sdk/pull/1456#discussion_r2026154467) and [here](https://github.com/stellar/rs-soroban-sdk/pull/1456#discussion_r2026155118)

soroban-sdk/src/crypto/bls12_381.rs
jayz22: it is a bug in the documentation. separated out the fix in https://github.com/stellar/rs-soroban-sdk/pull/1455

soroban-sdk/src/crypto/bls12_381.rs
jayz22: actually this is not a simple conversion, but performs mapping algorithms to map a number into a point. (the particular method is described [here](https://www.rfc-editor.org/rfc/rfc9380.html#name-simplified-swu-for-ab-0)). I've also added a comment in the code.

soroban-sdk/src/crypto/bls12_381.rs
jayz22: This is a great point. The `zero` checking is rather a corner case that isn't worth always paying the extra memcpy cost for. It should be sufficient to be done after the `Into::<BigInt<6>>` conversion, as you suggested. I've made the change in the other PR (https://github.com/stellar/rs-soroban-sdk/pull/1456#discussion_r2026156499)

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -680,6 +681,64 @@ mod objects {
...
+    impl SorobanArbitrary for Fr {
+        type Prototype = ArbitraryFr;
+    }
+
+    impl TryFromVal<Env, ArbitraryFr> for Fr {
+        type Error = ConversionError;
+
+        fn try_from_val(env: &Env, v: &ArbitraryFr) -> Result<Self, Self::Error> {
+            // Convert bytes to Fr via U256
+            Ok(Fr::from_bytes(BytesN::from_array(env, &v.bytes)))
```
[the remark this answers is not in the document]
jayz22: that's right.

leighmcculloch (commented):
This looks good to me. I only have one concern, mentioned inline, and my concern may be naive.

Let's get https://github.com/stellar/rs-soroban-sdk/pull/1456 merged, and then once this PR is updated I'll approve this PR.

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -680,6 +681,76 @@ mod objects {
...
+
+        fn try_from_val(env: &Env, v: &ArbitraryG1Affine) -> Result<Self, Self::Error> {
+            let mut bytes = v.bytes;
+            // the top 3 bits in a G1 point are reserved for flags:
+            // compression_flag (bit 0), infinity_flag (bit 1) and sort_flag
+            // (bit 2). Only infinity_flag is possible to be set, in which case
+            // the rest of the bytes must be zeros. The host will reject any
+            // invalid input. Clearing the flags here just to give it better
+            // chance of a valid input.
+            bytes[0] &= 0b0001_1111;
```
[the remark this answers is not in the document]
jayz22: Yeah you're right. The infinity is a valid point that should be allowed to be generated.
Fixed the handling of infinity point by checking the flag and clearing the rest of the bits.

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -680,6 +681,76 @@ mod objects {
...
+
+        fn try_from_val(env: &Env, v: &ArbitraryG2Affine) -> Result<Self, Self::Error> {
+            let mut bytes = v.bytes;
+            // the top 3 bits in a G1 point are reserved for flags:
+            // compression_flag (bit 0), infinity_flag (bit 1) and sort_flag
+            // (bit 2). Only infinity_flag is possible to be set, in which case
+            // the rest of the bytes must be zeros. The host will reject any
+            // invalid input. Clearing the flags here just to give it better
+            // chance of a valid input.
+            bytes[0] &= 0b0001_1111;
```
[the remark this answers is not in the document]
jayz22: fixed

leighmcculloch (commented):
> Add support for BLS12-381 curve points (G1Affine, G2Affine, Fr) as contract data (via contracttype)

There's no test testing that we've got everything we need here for it to work. Could you add a test in the sdk tests directory to make sure we aren't missing something?

Is the goal only to be able to write and read these from storage? Or also to be able to pass it as parameters into the contract? For the latter, which I think makes sense, we need to hook up specs. Normally types that live in the SDK end up in the ScSpecType list.

The only other SDK types that use `contracttype` macros internally are for auth, and those conveniently don't need exporting of specs (we turn that off) because they aren't externally callable so this limitation and gap hasn't had an impact until now.

The way we did this in the token sdk was use the contracttype macro, and when you import the crate its spec gets reexported implicitly. But that won't work to do that here because everyone imports this SDK.

These types are all aliases for `BytesN<_>`. The other case we have for that is `Hash`, but again that was a case not callable.

🤔 This is a bit of a conundrum. 

We can merge this as is, it'll work for storage, but fall apart at the boundaries of the contract as a fn input or output and when other tools like the CLI try and pass the value.

We have a couple paths:
- Do nothing, maybe that's enough for now.
- Add to the spec XDR as a type
- Find a way for the specs to see it as a BytesN for the sake of specs https://github.com/stellar/rs-soroban-sdk/blob/13263e86813c3991e805f3d55666fa0402c1bf26/soroban-spec-rust/src/types.rs#L174-L177
- Explore other approaches

Related issues:
- https://github.com/stellar/rs-soroban-sdk/issues/1330

leighmcculloch:
> * Find a way for the specs to see it as a BytesN for the sake of specs

The way we'd do this is modify the `contractimpl` macro that generates the functions, so that for any input parameter named one of these types, we map them to the `BytesN<_>` type of appropriate size.

soroban-sdk/src/crypto/bls12_381.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(not(target_family = "wasm"))]
+use crate::xdr::ScVal;
```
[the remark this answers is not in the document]
jayz22: This is necessary because of the `impl_bytesn_repr` macro.

leighmcculloch (approved):
The test shows that we don't have everything we need here for these to work everywhere `contracttype`s do.

If I deploy the test and run this command, it errors:
```
$ stellar contract invoke --id c -- --help
❌ error: Missing Entry G1Affine
```

But there is enough here for these to work as internal types for storage. So I think we can merge this, because it is additive, and what's added works for that narrow purpose, but we should follow up on filling the gap on contract function params / return values.

leighmcculloch:
> The way we'd do this is modify the `contractimpl` macro that generates the functions, so that for any input parameter named one of these types, we map them to the `BytesN<_>` type of appropriate size.

If we solve this problem generically, not just hard code these types specifically, then I think we are also solving the type alias issue, if we also solve it for type wrapping:
- #1063

leighmcculloch:
To be clearer there are two contracttype issues outstanding:
- The SDK doesn't have a way to define a type that is optionally exported.
- The contract specs don't have a way to define type aliases, where one type is really another type. In this case it is the G1Affine newtype idiom around the BytesN type. This will be addressed by #1063.

In the interim of solving both problems above I think it is okay if the contracttype macro sees G1Affine and treats it as if the user had typed `BytesN<...>`. Doing that will avoid both problems for the moment.

jayz22:
> To be clearer there are two contracttype issues outstanding:
> 
>     * The SDK doesn't have a way to define a type that is optionally exported, although the stellar-cli may do that for us in the future. [`stellar contract build` should strip contract specs, removing types that are not referenced by functions or events stellar-cli#2030](https://github.com/stellar/stellar-cli/issues/2030)
> 
>     * The contract specs don't have a way to define type aliases, where one type is really another type. In this case it is the G1Affine newtype idiom around the BytesN type. This will be addressed by [Support type aliases as contracttypes #1063](https://github.com/stellar/rs-soroban-sdk/issues/1063).
> 
> 
> In the interim, until both problems above are solved, I think it is okay if the `contractimpl` macro sees `G1Affine` and treats it as if the user had typed `BytesN<...>`. Doing that will avoid both problems for the moment.

Thanks @leighmcculloch, I've added the type spec mapping for the BLS types to its inner types (with a comment referencing #1063).
Also added a contract example and a test invoking it. I think that should do it for now. 
Agree it would be nice to support generic type inference (either via new type idiom or type alias).

soroban-sdk-macros/src/map_type.rs
```diff
@@ -37,6 +40,31 @@ pub fn map_type(t: &Type, allow_hash: bool) -> Result<ScSpecTypeDef, Error> {
...
+                    // Idiom. For more details, see the tracking issue for supporting
+                    // type aliases:
+                    // https://github.com/stellar/rs-soroban-sdk/issues/1063
+                    "G1Affine" => Ok(ScSpecTypeDef::BytesN(ScSpecTypeBytesN {
+                        n: G1_SERIALIZED_SIZE,
+                    })),
+                    "G2Affine" => Ok(ScSpecTypeDef::BytesN(ScSpecTypeBytesN {
+                        n: G2_SERIALIZED_SIZE,
+                    })),
+                    "Fr" => Ok(ScSpecTypeDef::U256),
```
leighmcculloch, line 67: @jayz22 Do the `Fp` and `Fp2` types need listing here too?

jayz22: Hmm good call. My initial intuition was no since these are just field elements that makes up the coordinates for G1 and G2. No need to store/pass them as function argument directly. 
But looking at it again, we do have host functions `bls12_381_map_fp2/1_to_g2/1` that maps a field element to a point. 
So I guess let's just add it for completeness.

jayz22: Added Fp and Fp2 and updated the test contract to invoke host functions using those.

leighmcculloch:
I pushed a small change to make the warning stand out a little more, using rustdocs warning class:

Before:
Screenshot 2025-05-16 at 1 44 16 pm

After:
Screenshot 2025-05-16 at 1 43 56 pm

leighmcculloch (approved):

Issue #1428, opened 2025-01-21
Support `wasm32v1-none` target


dmkozh:
`wasm32-none` target has been released in Rust 1.84 and SDK and its dependencies should support it. This is a tracking issue for the necessary steps:

- [x] Ensure `rs-soroban-env` supports `wasm32-none` via CI
- [x] Ensure `rs-stellar-xdr` supports `wasm32-none` via CI
- [x] Ensure `rs-soroban-sdk` supports `wasm32-none` via CI
- [x] Adapt CLI to build with correct rust version(s) and ban disallowed ones (done in https://github.com/stellar/stellar-cli/pull/2022)

graydon:
I think it's more the CLI's job not the SDK's to "work with any rust version". The CLI needs to choose a target to pass to the build based on whether it's on rust 1.81-or-before (use `wasm32-unknown-unknown`) or rust 1.84-or-after (use `wasm32v1-none`). The SDK should probably continue to work with `wasm32-unknown-unknown` but post-rust-1.82 that's a target we don't _want_ to use because it'll generate wasm-extension-using instructions that break on chain.

graydon:
As of https://github.com/stellar/stellar-cli/pull/2022 (which just merged) I don't think there's anything else left to do here.

PR #1453, opened 2025-03-26
Switch to post-rust-1.84 wasm target wasm32v1-none

graydon:
Part of wrapping up bug https://github.com/stellar/rs-soroban-sdk/issues/1428. Mostly this is just:

- Switching rust version to 1.84 which supports `wasm32v1-none`
- Switching the SDK to build with that, not `wasm32-unknown-unknown`

But there are a couple minor subtle bits:

- Some of the crates only build native, not wasm, and we weren't differentiating that in the Makefile right
- There _used_ to be a workaround for a mis-feature of `alloc` that required `std` in order to register an OOM handler; this was actually fixed in March 2023 (https://releases.rs/docs/1.68.0/ which contained https://github.com/rust-lang/rust/pull/102318/) but the workaround was already committed to the SDK by January 2023. Anyway, it's long since fixed and we can just delete the workaround, which we _have_ to in this PR, because `wasm32v1-none` doesn't ship with `std` at all.

Probably don't merge this until I have the other CI bits all switched over to `wasm32v1-none` as described in #1428


leighmcculloch (approved):
Hmm. I wonder if we should continue to build for wasm32-unknown-unknown. There's at least one case where someone will continue to use that target. It's not the intended use of the SDK though, so I guess we need to weigh that with making builds even slower, and increasing the costs of running builds.

leighmcculloch:
If we do something to break the SDK for wasm32-unknown-unknown, it'll likely be the type of thing that's easy to walk back. So I think it's pretty low risk to stop building for it. If we accidentally break it, and someone opens an issue indicating so, we can fix it. 👍🏻

PR #1454, opened 2025-03-31
Muxed address support for Soroban SDK.

dmkozh:
### What

This introduces the `MuxedAddress` type that maps to either `AddressObject` or `MuxedAddressObject`. This also adds a new helper for building token events with muxed addresses as per CAP-67.

### Why

Support for CAP-67.

### Known limitations

N/A


soroban-sdk/src/testutils.rs
```diff
@@ -438,6 +438,12 @@ pub trait Address {
     fn generate(env: &Env) -> crate::Address;
 }
 
+pub trait MuxedAddress {
+    /// Create a new MuxedAddress from a provided account public key and
+    /// multiplexing identifier.
+    fn from_account_id(env: &Env, account_key: &[u8; 32], id: u64) -> crate::MuxedAddress;
```
leighmcculloch, line 444: I think the account key input should be an Address. Users don't usually pass account keys around like this. The naming is also ambiguous to if it is the public key or secret key.
```suggestion
    fn new(env: &Env, address: &Address, muxed_id: u64) -> crate::MuxedAddress;
```

soroban-sdk/src/address.rs
```diff
@@ -55,6 +55,7 @@ impl Debug for Address {
                         let strkey = Strkey::Contract(Contract(contract_id.0));
                         write!(f, "Contract({})", strkey.to_string())?;
                     }
+                    _ => return Err(core::fmt::Error),
```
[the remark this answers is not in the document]
dmkozh: I'm not sure if it's realistic that we will extend the address space, but that wouldn't hurt either, done.

soroban-sdk/src/env.rs
```diff
@@ -354,6 +354,27 @@ impl Env {
...
+    }
+
+    #[doc(hidden)]
+    pub(crate) fn get_id_from_muxed_address(&self, muxed_address: MuxedAddressObject) -> u64 {
+        u64::try_from_val(
+            self,
+            &internal::Env::get_id_from_muxed_address(self, muxed_address).unwrap_infallible(),
+        )
+        .unwrap()
+    }
```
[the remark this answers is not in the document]
dmkozh: Done.

soroban-sdk/src/muxed_address.rs
```diff
@@ -0,0 +1,281 @@
...
+    Address(AddressObject),
+    MuxedAddress(MuxedAddressObject),
+}
+
+/// MuxedAddress is a union type that represents either the regular `Address`,
+/// or a 'multiplexed' address that consists of a regular address and a u64 id
+/// and can be used for representing the 'virtual' accounts that allows for
+/// managing multiple balances off-chain with only a single on-chain balance
+/// entry. The address part can be used as a regular `Address`, and the id
+/// part can be used only in the events for the off-chain processing.
```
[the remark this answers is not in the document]
dmkozh: Done.

soroban-sdk/src/muxed_address.rs
```diff
@@ -0,0 +1,281 @@
...
+///
+/// This type is only necessary in a few special cases, such as token transfers
+/// that support non-custodial accounts (e.g. for the exchange support). Prefer
+/// using the regular `Address` type unless multiplexing support is necessary.
+///
+/// This type is compatible with `Address` at the contract interface level, i.e.
+/// if a contract accepts `MuxedAddress` as an input, then its callers may still
+/// pass `Address` into the call and nothing will break. This means that if a
```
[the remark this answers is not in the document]
dmkozh: Done.

soroban-sdk/src/testutils.rs
dmkozh: That's the most obvious option, but it would require invalid input handling (which is confusing for the user as not every address will be accepted), and (more annoyingly) this would require a generator for the account addresses, which opens a whole new can of worms. I'd like to avoid explicitly adding account address generator for tests for the time being. I've tried a `generate` interface, as well as function for creating addresses with a different mux id. This way account addresses don't need to 'leak' into the usual address test utils.

leighmcculloch: Folks can use the existing `Address::generate`?

soroban-sdk/src/muxed_address.rs
```diff
@@ -0,0 +1,310 @@
...
+impl From<Address> for MuxedAddress {
+    fn from(address: Address) -> Self {
+        address
+            .as_object()
+            .try_into_val(address.env())
+            .unwrap_infallible()
+    }
+}
+
+impl MuxedAddress {
```
leighmcculloch, line 171: The `Address` type has a `from_str` and `from_string` functions. I guess we'd need additional host functions to make that happen. I think it's fine if we skip that until someone asks for it, since adding it is additive, but the APIs of the two types are a little disjoint without it.

Can we add the `from_str` and `from_string` as testutils at least so folks can use them in tests at least?

soroban-sdk/src/muxed_address.rs
```diff
@@ -0,0 +1,310 @@
...
+                ed25519: crate::env::internal::xdr::Uint256(
+                    env.with_generator(|mut g| g.address()),
+                ),
+                id,
+            },
+        ));
+        sc_val.try_into_val(env).unwrap()
+    }
+
+    fn clone_with_id(&self, new_id: u64) -> crate::MuxedAddress {
```
leighmcculloch, line 294: Instead of clone with id, could we add a `new` function with an `Address` and `id`? Then folks could compose the MuxedAddress without the Address having to have already been inside a MuxedAddress. It'll result in more consistent test code that's easier to copy-paste, otherwise the first muxed address has to be created with different code.

soroban-sdk/src/testutils.rs
dmkozh: No, the existing `Address::generate` creates contract addresses only and we can't simply switch it to be account address, as that would break SAC interactions. I don't think it's worth the effort to try and support that, given that what we're trying to do here is only necessary for a narrow use case (custom token implementations) that doesn't need to involve SAC.

soroban-sdk/src/muxed_address.rs
dmkozh: > but the APIs of the two types are a little disjoint without it.

That's actually intentional. Muxed addresses have pretty limited use and I don't think the reasons for why we support strkeys for address (like cross-chain interactions) apply to them, at least for now.

> Can we add the from_str and from_string as testutils at least so folks can use them in tests at least?

It's not hard to add, but I don't see the purpose for that on the other hand. Do you have some use cases in mind? I think a generator for arbitrary addresses is more convenient (arguably it could actually generate ids as well; I don't have too strong opinion on this though).

soroban-sdk/src/muxed_address.rs
dmkozh: As I've mentioned above, this would require us adding support for account addresses, and I don't think it's a good idea at the moment. Even if we had that, it would still be confusing, as contract addresses still wouldn't be supported.

> It'll also result in more consistent test code that's easier to copy-paste, 

Yeah, I realize this is not the prettiest interface and I'm open to suggestions on how we could improve it. But for now I think that whatever interface we come up with should not rely on generating account addresses.

leighmcculloch (approved):
I still think we should add a test function `MuxedAddress::new` and commented inline as I don't understand the limitation preventing us from doing that. But otherwise LGTM.

I think the change as a whole strikes the right balance in adding this functionality, convenience, while preventing mistakes. The comments do a good job of framing the type and calling out what not to do. 👏🏻

dmkozh:
> I'm not sure what that means. The SDK already supports account address (G addresses). The function I'm suggesting is a test function only, the same as the above generate function, it'd just error if the Address passed in wasn't a G address.

But there is no way to _generate_ a test G-account, is there? That's was mostly concerned about - if we only add a way to generate a muxed account from G-account, then for non-snapshot tests it would be necessary to also be able to generate G-address, which I think would be nice to avoid.

> The problem with not adding a way to construct a MuxedAddress from a G address and ID is that it is not possible to write a test that matches an account in an existing snapshot. So the API seems incomplete without it.

That's a fair point. I can add a G+id constructor, but leave the address as is, so whoever has the fully set up G-account can benefit from that, and for the quick unit tests the generate function will be sufficient. Then I guess I can get rid of clone as well.

> Actually, it looks like it's possible to convert from an Address to a MuxedAddress and then call clone_with_id, so still really unclear to me why a ::new can't do the same thing. What am I missing?

The intent was to only clone between 'actually' muxed addresses, but the implementation is also incorrect given that Address is indeed convertible to MuxedAddress (so a user trying to do what you suggested would hit an unhelpful `unreachable` panic). I'll just add `new` for then.

FWIW if we merge this, then well officially switch SDK to pre-protocol-23 mode, so please let me know if you want to merge something before this.

leighmcculloch:
> But there is no way to _generate_ a test G-account, is there?

Ah. Yes, there's no way to create accounts, we dropped that functionality a while ago. But you can create an Address from a G strkey, and if you import a snapshot, accounts can already exist.

> FWIW if we merge this, then well officially switch SDK to pre-protocol-23 mode, so please let me know if you want to merge something before this.

Thanks for calling that out, I think this will be the first v23 PR. We should probably try and merge and release all the other open PRs before this one, if not impractical.

leighmcculloch:
> > But there is no way to _generate_ a test G-account, is there?
> 
> Ah. Yes, there's no way to create accounts, we dropped that functionality a while ago. But you can create an Address from a G strkey, and if you import a snapshot, accounts can already exist.

I think we can get away with not creating accounts, at least for now. Folks can still test that MuxedAddresses work with their contract tokens, and they can still build muxed addresses that match accounts in a snapshot. If folks ask for creating accounts we could revisit that in a minor release I think?

dmkozh:
> If folks ask for creating accounts we could revisit that in a minor release I think?

Yeah, that seems reasonable. I'll update and ping this PR a bit later when I have some time - it doesn't seem like there is urgency to it for now.

soroban-sdk/src/testutils.rs
```diff
@@ -438,6 +438,27 @@ pub trait Address {
...
+    /// Returns a new `MuxedAddress` that has the same `Address` part as the
+    /// provided `address` and the provided multiplexing id.
+    ///
+    /// `address` can be either an `Address` or `MuxedAddress` and it has to
+    /// be an account (non-contract) address.
+    ///
+    /// Note on usage: the simplest way to test `MuxedAddress` is to generate
+    /// an arbitrary valid address with `MuxedAddress::generate`, then
+    /// `MuxedAddress::new` can be used to alter only the multiplexing id part
+    /// of that address.
```
leighmcculloch, line 458: Suggest slightly different instruction:
```suggestion
    /// Note on usage: the simplest way to test with multiple `MuxedAddress` that
    /// have the same underlying address is to generate an arbitrary address with
    /// `Address::generate`, then `MuxedAddress::new` can be used to create
    /// multiple `MuxedAddress` with that base `Address`.
```

soroban-token-sdk/src/event.rs
```diff
@@ -22,6 +22,45 @@ impl Events {
         self.env.events().publish(topics, amount);
     }
 
+    pub fn transfer_with_muxed_address(&self, from: MuxedAddress, to: MuxedAddress, amount: i128) {
```
leighmcculloch, line 25: I think this needs tweaking to change the from to an Address.

soroban-sdk/src/muxed_address.rs
leighmcculloch: > Do you have some use cases in mind?

When writing tests against mainnet state someone may need to be able to paste a muxed address in. From what's here it looks like they'd need to convert it to a G address, Address::from_str with the G, then MuxedAddress::new with the G and the id. So a dev can work around the missing function, it just seems unnecessary complexity. 

> I don't think the reasons for why we support strkeys for address (like cross-chain interactions) apply to them

I think it's possible a bridge might want to deliver to a custodial address, but yeah, there's no evidence of that being a thing right now.

---

I don't see the reason to gate keep this type of functionality. But, I agree, it seems pretty edge case. We can add it when/if someone asks for it, but I don't think we should hesitate to.

leighmcculloch (commented):
Looks good. I think there's just one tweak needed because the event publishing helper publishes with a from MuxedAddress, and that's no longer supported.

leighmcculloch:
For merging this, I think @jayz22 wants to do a minor release with new functionality before we merge p23 changes.

soroban-sdk/src/testutils.rs
dmkozh: The suggested instruction won't work because `Address::generate` generates a contract address.

soroban-token-sdk/src/event.rs
dmkozh: Done.

deny.toml
```diff
@@ -111,7 +111,7 @@ allow = [
     "Apache-2.0",
     "BSD-3-Clause",
     "Apache-2.0 WITH LLVM-exception",
-    "Unicode-DFS-2016",
+    "Unicode-3.0",
```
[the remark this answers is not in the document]
dmkozh: This seems like some kind of a fluke; this appeared due to some dependency bump that is no longer necessary.

dmkozh:
This has been rebased on top of rc1.

soroban-sdk/src/muxed_address.rs
```diff
@@ -0,0 +1,316 @@
...
+        let sc_val = ScVal::Address(crate::env::internal::xdr::ScAddress::MuxedAccount(
+            crate::env::internal::xdr::MuxedEd25519Account {
+                ed25519: crate::env::internal::xdr::Uint256(
+                    env.with_generator(|mut g| g.address()),
+                ),
+                id,
+            },
+        ));
+        sc_val.try_into_val(env).unwrap()
+    }
```
leighmcculloch, line 292: One last feedback on this function: I think we should drop the `id` parameter.

It feels odd to me that you can generate and provide part of the address as a fixed value. 

Conceptually most devs who are working with muxed addresses have a fixed base address, and variable ID, so having the reverse in this function, a random base and fixed ID, is conceptually a bit mind bending.

If folks really need to set the ID, they can call generate, then use `new` to change the id to a fixed value.

soroban-spec-rust/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version.workspace = true
 
 [dependencies]
-stellar-xdr = { workspace = true, features = ["curr", "std", "serde"] }
+stellar-xdr = { workspace = true, features = ["next", "std", "serde"] }
```
leighmcculloch, line 14: This is surprising, why do we need next for the specs? The specs would have been updated in both?

leighmcculloch (commented):
This looks good. I think we should make one tweak.

I also had one question about the xdr inline.

PR #1459, opened 2025-04-07
Remove Env from extern functions

### LINKED ISSUE #1402: Remove Env from the contract extern functions
I noticed today while debugging generated contract functions that there's a warning on the `Env` that is included in the extern contract function signatures saying that:
```
warning: `extern` fn uses type `Env`, which is not FFI-safe
   --> tests/account/src/lib.rs:309:14
    |
309 |         env: soroban_sdk::Env,
    |              ^^^^^^^^^^^^^^^^ not FFI-safe
    |
```

We could omit the Env in this function since we can instantiate it at the start of the function in Wasm builds.

Once upon a time both the test framework and real invocations when built to Wasm went through the same function which is why the Env was a parameter, but since then the extern function now only gets used externally when built to Wasm, so the parameter is unnecessary.

### LINKED ISSUE #1458: wasm32 builds reporting C-ABI warning

any

### What did you do?

Saw this article:
https://blog.rust-lang.org/2025/04/04/c-abi-changes-for-wasm32-unknown-unknown.html

Tried building with `nightly-2024-03-27` which includes warnings that detect when C-ABI will break in a future version:
```
 cargo +nightly-2025-03-27 build --target wasm32-unknown-unknown --release --package test_empty
```

Successful build without warnings.

### What did you see instead?

```
warning: this function definition involves an argument of type `Env` which is affected by the wasm ABI transition
 --> tests/empty/src/lib.rs:7:1
  |
7 | #[contractimpl]
  | ^^^^^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #138762 <https://github.com/rust-lang/rust/issues/138762>
  = help: the "C" ABI Rust uses on wasm32-unknown-unknown will change to align with the standard "C" ABI for this target
  = note: `#[warn(wasm_c_abi)]` on by default
  = note: this warning originates in the attribute macro `contractimpl` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `test_empty` (lib) generated 1 warning
    Finished `release` profile [optimized] target(s) in 1.44s
warning: the following packages contain code that will be rejected by a future version of Rust: test_empty v22.0.7 (/Users/leighmcculloch/Code/rs-soroban-sdk/tests/empty)
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 21`
```

leighmcculloch:
### What
Remove Env parameter from invoke_raw_extern function, that is externed for each callable wasm function.

### Why
The extern functions should need to be C-ABI compatibile for a near future version of Rust which is standardizing the wasm32-unknown-unknown build on C-ABI that up until then wasn't actually C-ABI. On nightly-2025-03-27 we can see a warning already appearing, like that documented at #1458. This impact if not changed is likely that on some future version the build will hard error rather than just warn. The SDK is about to switch to being recommended building for wasm32v1-none anyway, and the warning isn't appearing on it. Presumedly a change like this will or at least may occur to wasm32v1-none too in the future. The change is beneficial in any case as it makes it much clearer as to what is happening contract side during construction. Anyone could not be faulted for wondering what the Env was given the value was not actually a parameter being passed from the host even though it looked like it was.

For anyone wondering why the Env was even included as a parameter, it's a hold over from early contract implementations where we were trying to make the extern functions the actual functions that developers were writing by hand. Those early functions included the Env that in tests would be passed in and be a non-zero sized struct, but then in wasm builds would become a zero sized type which the compiler would omit as a parameter. The SDK no longer needs that hack because the extern functions are code generated via macros and are not the same functions that developers code by hand.

For testing this, other than running all existing tests, I diffed a wasm binary built before and after this change. The only observable difference to the wasm was the embedded sdk version.

Close #1402
Close #1458

### DIFF
--- a/soroban-sdk-macros/src/derive_fn.rs
+++ b/soroban-sdk-macros/src/derive_fn.rs
@@ -174,9 +174,12 @@ pub fn derive_pub_fn(
 
             #[deprecated(note = #deprecated_note)]
             #[cfg_attr(target_family = "wasm", export_name = #wrap_export_name)]
-            pub extern "C" fn invoke_raw_extern(env: #crate_path::Env, #(#wrap_args),*) -> #crate_path::Val {
+            pub extern "C" fn invoke_raw_extern(#(#wrap_args),*) -> #crate_path::Val {
                 #[allow(deprecated)]
-                invoke_raw(env, #(#passthrough_calls),*)
+                invoke_raw(
+                    #crate_path::Env::default(),
+                    #(#passthrough_calls),*
+                )
             }
 
             use super::*;


graydon (approved):
LGTM

graydon:
(For future reference the new / official ABI is https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md -- not much of a spec but enough to be clear that scalar args are scalar args and .. we should really only pass scalar args. Besides weird situations like this one!)

Issue #1097, opened 2023-09-23
Include events in contract interfaces / spec


leighmcculloch:
It's been suggested that events could be a type in Rust contracts, and then those types could become part of the interface, docs, and generated code in clients.

Today events are published like so in contracts:
```rust
env
    .events()
    .publish(
        // Topics
        (
            // Topic 1, any value allowed 
            symbol_short!("COUNTER"),
            // Topic 2, any value allowed
            symbol_short!("increment"),
            // Topic N, more topics if needed
        ),
        // Data, can take any form, may be something simple
        // like an integer, or may be an array or map
        count
    );
```

The existing method to publishing makes it very difficult to capture the schema, because the schema is variable and now described.

leighmcculloch:
There's a discussion in the stellar-protocol repo that signals some value in including the schema of events, or at least their topic first element which are usually a name, into the spec:
- https://github.com/stellar/stellar-protocol/discussions/1596

leighmcculloch:
Here's one idea for how to "structure" an event into a type so that it has some schema.
```rust
#[contractevent(
    // Fixed topics can be specified here.
    "COUNTER",
)]
pub struct CounterEvent {
    #[topic] // There can be multiple topics that get written sequentially to the topic list after any fixed topics.
    pub name: Symbol,
    #[data] // There can only be one data. If multiple data are needed, make it a map or another type.
    pub count: u32,
}
```
My concern with something like this, is that while it improves the schema, it's significantly more work to write that to simply write out `env.events().publish((topic1, topic2), data)`.

leighmcculloch:
Another option is a trait, but I think this would be even more verbose, so I wouldn't lean to this.

In the soroban-sdk:
```rust
// Already in the soroban-sdk.
pub trait Topics: IntoVal<Env, Vec<Val>> { }

// Could be added to the soroban-sdk.
pub trait ContractEvent {
    type Topics: Topics;
    type Data;

    fn topics(&self) -> Self::Topics;
    fn data(&self) -> Self::Data;
}
```

In a contract:
```rust
pub struct MyEvent { /* private fields */ }

#[contracteventimpl]
impl ContractEvent for MyEvent {
    type Topics = (Symbol, Symbol);
    type Data = u32;
    
    fn topics(&self) -> Self::Topics { ... }
    fn data(&self) -> Self::Data { ... }
}
```

sreuland:
[...]

PR #1473, opened 2025-05-30
Add contractevent macro and event support to contract spec

leighmcculloch:
### What

Add `contractevent` macro for defining events, and event support to contract spec.

In it's most basic form, as follows, where an automatic prefix topic based on the name will be published, with any topic fields appended to the topic list, and any other fields emitted as map entries in a map:

```rust
    #[contractevent]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        #[topic]
        pub my_topic: u32,
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

But also supporting more advanced customisable forms, where prefix topics can be overriden and a max of two provided, and where the data section can be defined as either being a `vec` or a `single-value` instead.

### Why

The `contractevent` macro will achieve a couple things for events.

It will allow developers to define the event as a type, in a similar way that they can define an event in Solidity contracts. Defining it as a type makes it easier to reuse the event, and to define the types safely and be clear about what types within the event are being published.

It will provide a foundation that the Rust SDK can use to include in the contract interface (spec) an entry for the event. That entry can then be used by downstream systems to bring more meaning to the raw event data that is visible in transaction / ledger close meta. For an example of what that looks like, see https://github.com/orgs/stellar/discussions/1724#discussioncomment-13118291.

The advanced customisations are a necessity to be able to make the events and the event schema describe existing contract events. The network already has a variety of events and some contracts use two fields as fixed prefix topics, such as the Soroswap contracts. Some use vecs for the data section, and even a single value such as an `amount: i128`, such as the Stellar Asset Contract.

Close https://github.com/stellar/rs-soroban-sdk/issues/1097

### Docs for the `contractevent` attribute macro

Generates conversions from the struct into a published event.

Fields of the struct become topics and data parameters in the published event.

Includes the event in the contract spec so that clients can generate bindings for the type and downstream systems can understand the meaning of the event.

##### Basic Contract Event

Defining a basic contract event.

The event will have a single fixed prefix topic matching the name of the struct in lower snake case. The fixed prefix topic will appear before any topics listed as fields. In the example below, the topics for the event will be:

*   `"my_event"`
*   u32 value from the `my_topic` field

The event’s data will be a [`Map`](struct.Map.html "struct soroban_sdk::Map"), containing a key-value pair for each field with the key being the name as a [`Symbol`](struct.Symbol.html "struct soroban_sdk::Symbol"). In the example below, the data for the event will be:

*   key: my\_event\_data => val: u32
*   key: more\_event\_data => val: u64

```rust
    #![no_std]
    use soroban_sdk::contractevent;
    
    // Define the event using the `contractevent` attribute macro.
    #[contractevent]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        // Mark fields as topics, for the value to be included in the events topic list so
        // that downstream systems know to index it.
        #[topic]
        pub my_topic: u32,
        // Fields not marked as topics will appear in the events data section.
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

##### Prefix Topics

Defining a contract event with a custom set of fixed prefix topics.

The prefix topic list can be set to another value. In the example below, the topics for the event will be:

*   `"my_contract"`
*   `"an_event"`
*   u32 value from the `my_topic` field

```rust
    #![no_std]
    use soroban_sdk::contractevent;
    
    // Define the event using the `contractevent` attribute macro.
    #[contractevent(topics = ["my_contract", "an_event"])]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        // Mark fields as topics, for the value to be included in the events topic list so
        // that downstream systems know to index it.
        #[topic]
        pub my_topic: u32,
        // Fields not marked as topics will appear in the events data section.
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

##### Data Format

Defining a contract event with a different data format. The data format of the event is by default a map, but can alternatively be defined as a `vec` or `single-value`.

###### Vec

In the example below, the data for the event will be a [`Vec`](struct.Vec.html "struct soroban_sdk::Vec") containing:

*   u32
*   u64

```rust
    #![no_std]
    use soroban_sdk::contractevent;
    
    // Define the event using the `contractevent` attribute macro.
    #[contractevent(data_format = "vec")]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        // Mark fields as topics, for the value to be included in the events topic list so
        // that downstream systems know to index it.
        #[topic]
        pub my_topic: u32,
        // Fields not marked as topics will appear in the events data section.
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

###### Single Value

In the example below, the data for the event will be a u32.

When the data format is a single value there must be no more than one data field.

```rust
    #![no_std]
    use soroban_sdk::contractevent;
    
    // Define the event using the `contractevent` attribute macro.
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        // Mark fields as topics, for the value to be included in the events topic list so
        // that downstream systems know to index it.
        #[topic]
        pub my_topic: u32,
        // Fields not marked as topics will appear in the events data section.
        pub my_event_data: u32,
    }
```

##### A Full Example

Defining a contract event, publishing it in a contract, and testing it.

```rust
    #![no_std]
    use soroban_sdk::{contract, contractevent, contractimpl, contracttype, symbol_short, Env, Symbol};
    
    // Define the event using the `contractevent` attribute macro.
    #[contractevent]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct Increment {
        // Mark fields as topics, for the value to be included in the events topic list so
        // that downstream systems know to index it.
        #[topic]
        pub change: u32,
        // Fields not marked as topics will appear in the events data section.
        pub count: u32,
    }
    
    #[contracttype]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct State {
        pub count: u32,
        pub last_incr: u32,
    }
    
    #[contract]
    pub struct Contract;
    
    #[contractimpl]
    impl Contract {
        /// Increment increments an internal counter, and returns the value.
        /// Publishes an event about the change in the counter.
        pub fn increment(env: Env, incr: u32) -> u32 {
            // Get the current count.
            let mut state = Self::get_state(env.clone());
    
            // Increment the count.
            state.count += incr;
            state.last_incr = incr;
    
            // Save the count.
            env.storage().persistent().set(&symbol_short!("STATE"), &state);
    
            // Publish an event about the change.
            Increment {
                change: incr,
                count: state.count,
            }.publish(&env);
    
            // Return the count to the caller.
            state.count
        }
    
        /// Return the current state.
        pub fn get_state(env: Env) -> State {
            env.storage().persistent()
                .get(&symbol_short!("STATE"))
                .unwrap_or_else(|| State::default()) // If no value set, assume 0.
        }
    }
    
    #[test]
    fn test() {
        let env = Env::default();
    
        assert_eq!(client.increment(&1), 1);
        assert_eq!(client.increment(&10), 11);
        assert_eq!(
            client.get_state(),
            State {
                count: 11,
                last_incr: 10,
            },
        );
    }
```

### DIFF
--- a/soroban-sdk-macros/src/derive_event.rs
+++ b/soroban-sdk-macros/src/derive_event.rs
@@ -4,6 +4,7 @@ use crate::{
 };
 use darling::{ast::NestedMeta, Error, FromMeta};
 use heck::ToSnakeCase;
+use itertools::Itertools as _;
 use proc_macro2::Span;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{format_ident, quote};
@@ -214,6 +215,7 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
     let data_params = params
         .iter()
         .filter(|p| p.location == ScSpecEventParamLocationV0::Data)
+        .sorted_by_key(|p| p.name.to_string()) // must be sorted for map_new_from_slices
         .collect::<Vec<_>>();
     let data_params_count = data_params.len();
     let data_idents = data_params


leighmcculloch:
We could consider an alternative way to define topics:

Replace this:
```rust
    #[contractevent]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        #[topic]
        pub my_topic: u32,
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

With:

```rust
    #[contractevent(topics = ["my_event", my_topic]]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        pub my_topic: u32,
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
````

In this way the topic list is defined in a single location, and the full topic list that will be found in the final event is seen in a single list. Fixed topics and the variable topics are clearly shown in a single list.

**Comment by leighmcculloch** [comment]
Yet another alternative way to define topics, which is largely a simplification, although in the default case there is more to do:

Replace this:
```rust
    #[contractevent]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        #[topic]
        pub my_topic: u32,
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
```

With:

```rust
    #[contractevent(topics = ["my_event"])]
    #[derive(Clone, Default, Debug, Eq, PartialEq)]
    pub struct MyEvent {
        #[topic]
        pub my_topic: u32,
        pub my_event_data: u32,
        pub more_event_data: u64,
    }
````

In this way there is no "default" behaviour where a topic is inserted with the type name. Topics are always explicitly defined.

**Comment by leighmcculloch** [comment]
> ```rust
>     #[contractevent(topics = ["my_event"])]
> ```

I gave this approach a go, but it's not particularly enjoyable to have to type the topics out in the cases where the struct name matches the event name, so I'm leaning away from that back to the original solution.

**Comment by leighmcculloch** [comment]
I got feedback offline about the concept of `prefix_topics` being unintuitive.

So I've renamed `prefix_topics` to simply `topics`.

File: soroban-spec-rust/src/types.rs [hunk 1]
```diff
@@ -134,6 +136,39 @@ pub fn generate_error_enum(spec: &ScSpecUdtErrorEnumV0) -> TokenStream {
...
+                    #[topic]
+                    pub #p_ident: #p_type
+                },
+                ScSpecEventParamLocationV0::Data => quote! {
+                    pub #p_ident: #p_type
+                },
+            }
+        });
+        quote! {
+            #[soroban_sdk::contractevent(export = false)]
```
leighmcculloch: TODO:
- [x] Emit fixed topics into the `topics` attribute.

leighmcculloch:
I found a bug in how the map variant of events are passed to the host.

leighmcculloch:
> I found a bug in how the map variant of events are passed to the host.

Fixed in 3f977d4b3fe9fbf30e28be7df34565541673be8d.

dmkozh (approved):

PR #1474, opened 2025-06-17
Allow string macros (env, concat, etc) in contract meta values

leighmcculloch:
### What
  Enable use of string macros like env! and concat! inside contract meta values.

  ### Why
  Allows for the use of things like environment variables, and concatenating data from multiple places inside contract meta values. Within the SDK this makes it simpler to inject the version of the SDK into the meta. Within contracts this will allow folks to inject things like the version of the crate into meta by using a value like `env!("CARGO_PKG_VERSION")`.

### DIFF
--- a/soroban-sdk-macros/Cargo.toml
+++ b/soroban-sdk-macros/Cargo.toml
@@ -14,10 +14,6 @@ rust-version.workspace = true
 proc-macro = true
 doctest = false
 
-[build-dependencies]
-rustc_version = "0.4.0"
-crate-git-revision = "0.0.6"
-
 [dependencies]
 soroban-spec = { workspace = true }
 soroban-spec-rust = { workspace = true }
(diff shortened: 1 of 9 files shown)


leighmcculloch:
The nightly build failures are fixed by:
- https://github.com/stellar/rs-soroban-sdk/pull/1475

dmkozh (approved):

soroban-sdk-macros/Cargo.toml
```diff
@@ -23,7 +23,7 @@ syn = {version="2.0.77",features=["full"]}
 quote = "1.0"
 proc-macro2 = "1.0"
 itertools = "0.10.5"
-darling = "0.20.0"
+darling = "0.20.10"
```
leighmcculloch, line 26: With the use of `darling(with = )` the sdk now requires a newer version of darling that was previously not being required.

PR #1476, opened 2025-06-19
Fix testutils incompatibility with soroban-ledger-snapshot

leighmcculloch:
### What

Add a `testutils` feature to `soroban-ledger-snapshot`, and use it to gate the functionality in ledger snapshots that uses the `soroban-env-host`'s `testutils` features.

### Why

In https://github.com/stellar/rs-soroban-sdk/pull/1467 the `soroban-ledger-snapshot` crate dependency on `soroban-env-host` was changed to include an always-enabled dependency on the `testutils` feature. This appears to have been done because the function `get_stored_entries` of the `Host` was changed from being always available, to only available in `testutils` mode. The `soroban-ledger-snapshot` crate uses the `get_stored_entries` function.

A side-effect of of always pulling in the `soroban-env-host` is that the `soroban-ledger-snapshot` crate became incompatible with any project that also depends on the `soroban-sdk` without its `testutil` feature enabled. This is because the `soroban_env_host::EnvBase` trait type, which is implemented by the `soroban-sdk` crate, has functions that are required when the `testutil` feature is enabled.

To resolve the above challenges with only changes to this repository the most straight-forward thing to do is to move the functionality in `soroban-ledger-snapshot` that now needs `testutils` in `soroban-env-host` behind a `testutils` feature as well, which `soroban-sdk` enables when it uses `soroban-ledger-snapshot`.

### Known Limitations

Features on traits, like `EnvBase`, inherently create situations where dependencies can get stuck in situations where they are incompatible because another dep might be dependent on the trait in the opposite state. We should probably revisit `EnvBase`'s liberal use of features, breaking it up into separate traits. Rust features should always be, where possible, additively compatible. I've opened the following issue for us to explore that sometime:
- https://github.com/stellar/rs-soroban-env/issues/1570

We may also want to revisit whether some simple getters like `get_stored_entries` truly need to be gated behind a testutils flag. Features in general come at a cost in maintenance and rigidity. Typically features in Rust projects are used to reduce dependencies, but in `soroban-env-host` we are using it to limit features in different use cases.

After https://github.com/stellar/rs-soroban-sdk/pull/1467 and the change in this PR, some of the functionality in `soroban-ledger-snapshot` crate is unusable with the `Host`, unless the `Host` is being used in `testutils` mode. Today the only uses of the library that use that functionality is when the `soroban-sdk` is also being used with `testutils` enabled, but it is feasible that we may want to use the functionality in situations where it doesn't make sense to run the `Host` in `testutils` mode. This would probably mean changing the function in `soroban-env-host` that became testutils only, to be available without `testutils` again. We don't need it today, so I'm not proposing it today.

cc @stellar/devx @stellar/contract-committers @fnando

### DIFF
--- a/soroban-sdk/Cargo.toml
+++ b/soroban-sdk/Cargo.toml
@@ -45,7 +45,6 @@ soroban-env-host = { workspace = true, features = ["testutils"] }
 soroban-ledger-snapshot = { workspace = true, features = ["testutils"] }
 stellar-xdr = { workspace = true, features = ["curr", "std"] }
 soroban-spec = { workspace = true }
-soroban-ledger-snapshot = { workspace = true }
 ed25519-dalek = "2.0.0"
 rand = "0.8.5"
 ctor = "0.2.9"


dmkozh:
> We may also want to revisit whether some simple getters like get_stored_entries truly need to be gated behind a testutils flag. Features in general come at a cost in maintenance and rigidity. Typically features in Rust projects are used to reduce dependencies, but in soroban-env-host we are using it to limit features in different use cases.

I would argue that dependency reduction is a side effect, and the primary point of 'features' is to conditionally compile, well, features, as the name suggests. And in case of `testutils` we have a pretty clear cut for what is or is not a test utility. Given that host implements the protocol, I would really hesitate about unconditionally adding any functionality that doesn't contribute into the protocol implementation. Besides making the protocol safer, I think this feature gate is really helpful for being able to ship the SDK features and test utilities without worrying about the protocol behavior changes.

While I don't mind this PR, I don't quite understand the motivation behind not enabling testutils in snapshot, specifically this part:

> today the only uses of the library that use that functionality is when the soroban-sdk is also being used with testutils enabled, but it is feasible that we may want to use the functionality in situations where it doesn't make sense to run the Host in testutils mode.

Isn't the snapshot a test utility as a whole? Also, I can't imagine a scenario where host doesn't run in testutils mode and the snapshot is used. Getting the intermediate stored entries is not something that `Host` is meant to do in the normal operation mode (which is why I've moved the function into test utils in the first place).

dmkozh (approved):

PR #1499, opened 2025-07-20
Support ref fields on contractevent structs

### LINKED ISSUE #1491: Allow for lifetime generics and ref field types in `contractevent` types
We should allow for lifetime generics and ref field types in `contractevent` types.

The `contractevent` macro doesn't allow them today because it shares a lot of the foundation that also support the `contracttype` macro, but the two macros have a big difference in that `contracttype` types have to be unpacked from a Val into a type, where-as events are always one way, only being packed into a `Val`.

It may be helpful for larger types that get put into an event field if the type can be simply reference rather than copied. For most SDK types on their own this doesn't present a problem because most SDK types are just a 64-bit integer handle. But for user-defined `contracttype`s, they could be of arbitrary size, and having to make a copy into another struct seems ashame just to do the publish.

leighmcculloch:
### What
  Add support for reference fields in `contractevent` structs, allowing borrowed values to be used in event fields.

  ### Why
  When publishing events that are made up of user-defined-types, this gives the developer the option of not copying the value into the event, instead referencing it wherever it is. This may present some optimisation opportunities over being forced to copy the value into the struct. Nothing about this change requires the use of references, it merely makes it allowed rather than disallowed.

Close #1491


soroban-sdk-macros/src/derive_event.rs
```diff
@@ -244,7 +250,14 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
         },
         DataFormat::Vec => quote! {
             use #path::IntoVal;
-            (#(&self.#data_idents,)*).into_val(env)
+            (
+                #(
+                    {
+                        let val: #path::Val = self.#data_idents.into_val(env);
```
leighmcculloch, line 256: I changed this conversion to preconvert to a `Val`, so that the macro code didn't need to handle the case where the field is a ref or not, which would otherwise complicate it. There should be no harm in doing this.

leighmcculloch:
A side note that we may want to allow reference fields on `contracttype` for the same reason. Although, any type with reference fields can only be used when writing contract storage, not reading it. So the utility is limited, but may be useful for optimisations.

soroban-sdk-macros/src/derive_event.rs
```diff
@@ -244,7 +250,14 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
...
+                        val
+                    },
```
leighmcculloch: This should be simplified to the following once #1502 is merged.
```suggestion
                    &self.#data_idents,
```

graydon (approved):

PR #1510, opened 2025-08-05
Fix contractimpl macro composability with third party macros

### LINKED ISSUE #1509: `contractimpl` broken for functions with third-party attributes

Bug was introduced in the following commit, because it does not occur for the commit prior.

- e410245cbe04e6eb9df0cf45ed6a1c5ce9fbf780 (#1427)

### What did you do?

Used a third party function attribute macro on a function inside a `contractimpl` block. This is a reduced example using a very simple attribute macro.

```toml
[package]
name = "test"
version = "0.0.0"

[lib]
crate-type = ["cdylib"]

[dependencies.soroban-sdk]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", rev = "e410245cbe04e6eb9df0cf45ed6a1c5ce9fbf780" }
must_be_empty = "0.1.0"
```

```rust
#![no_std]
use must_be_empty::must_be_empty;

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    #[must_be_empty]
    pub fn empty() {}
}
```

A successful compile, and that the attribute would apply to the code under the function.

### What did you see instead?

```
$ stellar contract build
ℹ️ CARGO_BUILD_RUSTFLAGS=--remap-path-prefix=/Users/leighmcculloch/.cargo/registry/src= cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32v1-none --release
   Compiling test_macros v23.0.0-rc.1.1 (/Users/leighmcculloch/Code/rs-soroban-sdk/main/tests/macros)
error: expected `fn`
 --> tests/macros/src/lib.rs:8:1
  |
8 | #[contractimpl]
  | ^^^^^^^^^^^^^^^
  |
```

The attribute macro is actually being applied to all the generated code for the function being exported, including the exported contract spec XDR that is contained in a constant. Because that constant is not a function, the attribute macro is erroring because it is only intended to be used on functions.

### Impact

The actual case I discovered this is using the `stellar-axelar-std` crate (by @axelarnetwork) with new soroban-sdk versions. So at the moment if that crate tried to upgrade to v23 of the SDK, it would be blocked from doing so.

leighmcculloch:
### What
  Add a test vector that demonstrates a composability issue with third party function macros. Apple the existing filtering of which attributes get applied to the generated exported functions also to the generated spec constant.

  ### Why
  In #1427 I moved where the attributes were used as part of fixing another bug, and inadvertently moved the expansion of the attributes to the line before where the attributes are filtered. The filter ensures that only a specific set of attributes get applied to the generated code. The attributes still need to go through the filter for their use on the spec constant and it was an error on my part not to also move the filter.

Close #1509

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -852,6 +852,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "78ca9ab1a0babb1e7d5695e3530886289c18cf2f87ec19a575a0abdce112e3a3"
 
 [[package]]
+name = "must_be_empty"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70a261672ce542b9c17753db9b6d9d921cf8b48ef1dd9952726db5ad9a5f9312"
+dependencies = [
+ "proc-macro-error",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "num-bigint"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


sisuresh (approved):

leighmcculloch:
I pushed a few more changes after @sisuresh's review. The changes are isolated to the testing of the fix, and the fix has not been altered, so I'm not going to wait for a re-review of the change. In the original change the tests used a third party attribute to test the fix. I've changed the test to use a locally defined attribute inside a test vector designed for testing macro interactions. We can build on this test vector to add other macro interactions.

PR #1513, opened 2025-08-05
Remove leading underscores from fn args in specs

leighmcculloch:
### What
  Strip leading underscores from function argument names in the generated contract spec.

  ### Why
  Prevents implementation details from leaking into the spec and ensures consistency across different implementations of the same trait. Underscores are used as prefixes to argument names when a function implementation doesn't make use of the argument. That's an implementation detail, and surfacing that implementation detail into specs causes different implementations of the same trait to be different even though the interface is the same.

I debated about whether it should only strip a single underscore, or any number of underscores. I went with it stripping all underscores because I think it would be more surprising if `__arg` became `_arg` or stayed as `__arg` when `_arg` becomes `arg`.

### DIFF
--- a/soroban-sdk-macros/src/derive_spec_fn.rs
+++ b/soroban-sdk-macros/src/derive_spec_fn.rs
@@ -64,6 +64,16 @@ pub fn derive_fn_spec(
                     "".to_string()
                 };
 
+                // Strip any underscore prefix characters. Implementation's that do not use an
+                // argument will prefix an underscore to the variable name to signal to the
+                // compiler that the developer acknowledges they will not be using the parameter.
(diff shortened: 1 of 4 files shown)


soroban-sdk-macros/src/derive_spec_fn.rs
```diff
@@ -64,6 +64,16 @@ pub fn derive_fn_spec(
                     "".to_string()
                 };
 
+                // Strip any underscore prefix characters. Implementation's that do not use an
```
dmkozh: typo: Implementations

leighmcculloch:
```suggestion
                // Strip any underscore prefix characters. Implementations that do not use an
```

Issue #1487, opened 2025-07-13
String to Bytes conversion


Pfed-prog:
I am struggling to convert string to Bytes because the tutorial example mentions a functionality that does not work unfortunately.

I have created an issue for the tutorial here:
https://github.com/stellar/stellar-docs/issues/1603

I would really appreciate some guidance on how to convert string to bytes for keccak256 implemenation

leighmcculloch:
The BytesN and Bytes types do not support conversions from String in this manner. The conversion will panic because the Val will not be of the correct type. The docs appear incorrect and will need updating.

However, conversions from String to Bytes and vice versa are coming in Protocol 23 as part of CAP-69. They aren't available yet on mainnet, but will be soon:
- https://github.com/stellar/stellar-protocol/discussions/1633

In the interim you'd need to copy the String to a byte slice in memory, then wrap that in the Bytes.

PR #1514, opened 2025-08-05
Add Bytes<->String conversions via From trait.

dmkozh:
### What

Add Bytes<->String conversions via From trait.

### Why

Simplifying the conversions using the new p23 host functions

Resolves https://github.com/stellar/rs-soroban-sdk/issues/1487

### Known limitations

N/A


dmkozh:
> Could we also add:
> 
> * `to_string()` function to `Bytes`
> * `to_bytes()` function to `String`
> 
> ?

Done.

leighmcculloch (approved):

PR #1516, opened 2025-08-05
Generate a seamless conversion from Address to MuxedAddress arguments in clients

dmkozh:
### What

Generate a seamless conversion from Address to MuxedAddress arguments in clients.

Also switch the token interface to use MuxedAddress, as this shouldn't be a breaking change anymore.

### Why

Besides removing the friction from the token interface users, this change corresponds to the protocol semantics of MuxedAddress - any contract that accepts MuxedAddress is going to be interoperable with the contracts that expect that it accepts just Address.

### Known limitations

N/A

### DIFF
--- a/soroban-sdk-macros/src/derive_client.rs
+++ b/soroban-sdk-macros/src/derive_client.rs
@@ -1,8 +1,35 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{Error, FnArg, LitStr, Path, Type, TypePath, TypeReference};
+use syn::{Error, FnArg, LitStr, PatType, Path, Type, TypePath, TypeReference};
 
 use crate::{attribute::pass_through_attr_to_gen_code, symbol, syn_ext};
+
+fn is_muxed_address_type(arg: &FnArg) -> bool {
+    if let FnArg::Typed(pat_type) = arg {
+        if let Type::Path(TypePath { path, .. }) = &*pat_type.ty {
+            let path_str = quote::quote!(#path).to_string();
+            // Check for some common ways to refer to MuxedAddress. The direct
+            // check has a mild risk of a name collision, but the consequences
+            // should be insignificant.
+            return path_str == "MuxedAddress"
+                || path_str == "soroban_sdk :: MuxedAddress"
+                || path_str == ":: soroban_sdk :: MuxedAddress";
+        }
+    }
+    false
+}
+
+fn convert_to_into_muxed_address(arg: &FnArg) -> FnArg {
+    if let FnArg::Typed(pat_type) = arg {
+        return FnArg::Typed(PatType {
+            attrs: pat_type.attrs.clone(),
+            pat: pat_type.pat.clone(),
+            colon_token: pat_type.colon_token,
+            ty: Box::new(syn::parse_quote! { impl Into<MuxedAddress> }),
+        });
+    }
+    arg.clone()
+}
 
 pub fn derive_client_type(crate_path: &Path, ty: &str, name: &str) -> TokenStream {
     let ty_str = quote!(#ty).to_string();
@@ -180,7 +207,7 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
             });
 
             // Map all remaining inputs.
-            let (fn_input_names, fn_input_types): (Vec<_>, Vec<_>) = f
+            let mapped_inputs: Vec<_> = f
                 .inputs
                 .iter()
                 .skip(if env_input.is_some() { 1 } else { 0 })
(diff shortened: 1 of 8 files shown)


soroban-sdk-macros/src/derive_client.rs
```diff
@@ -192,9 +219,34 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
...
+            let fn_input_conversions: Vec<_> = mapped_inputs
+                .iter()
+                .map(|(ident, _, is_muxed_address)| {
+                    if *is_muxed_address {
+                        quote! { #ident.into().into_val(&self.env) }
+                    } else {
+                        quote! { #ident.into_val(&self.env) }
+                    }
                 })
-                .unzip();
```
leighmcculloch, line 197: Fyi you can retain the unzip pattern by importing the `itertools::MultiUnzip` which adds support for unzip doing more than two. We use it in other files in the macros crate, just not here because the stdlib unzip supports two and this code only needed to until now.

soroban-sdk-macros/src/derive_client.rs
```diff
@@ -1,9 +1,36 @@
...
+    if let FnArg::Typed(pat_type) = arg {
+        return FnArg::Typed(PatType {
+            attrs: pat_type.attrs.clone(),
+            pat: pat_type.pat.clone(),
+            colon_token: pat_type.colon_token,
+            ty: Box::new(syn::parse_quote! { impl Into<MuxedAddress> }),
+        });
+    }
+    arg.clone()
+}
```
leighmcculloch: Could `convert_to_into_muxed_address ` be moved into the `syn_ext` module, alongside the `fn_arg_make_ref`, and made into a function for converting a type into a `fn_arg_make_into`? Ideally it does so by reusing the existing `ty` field and just wrapping the `ty` inside the into, e.g.:
```rust
syn::parse_quote! { impl Into<#ty> }
```

The `MuxedAddress` might not be imported into the current module, but this function assumes it does not need to be fully qualified. Rather than having to figure that out, we can pass through the `ty`, however it is written.

soroban-sdk-macros/src/derive_client.rs
```diff
@@ -1,9 +1,36 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{Error, FnArg, LitStr, Path, Type, TypePath, TypeReference};
+use syn::{Error, FnArg, LitStr, PatType, Path, Type, TypePath, TypeReference};
 
 use crate::{attribute::pass_through_attr_to_gen_code, symbol, syn_ext};
 
+fn is_muxed_address_type(arg: &FnArg) -> bool {
```
leighmcculloch, line 11: I think we should use the `map_type` function that's available in this crate. It'll tell you if a type matches as a MuxedAddress, and then we've only got one set of matching logic that we don't have to keep in sync or debug subtle differences. Both functions appear to have used different approaches that have some subtle differences.

leighmcculloch (commented):
Couple asks inline, this looks good though. Thanks for adding tests for all the generated functions 👏🏻.

soroban-sdk/src/token.rs
```diff
@@ -139,9 +139,17 @@ pub trait TokenInterface {
...
+    /// Emits an event with topics
+    /// `["transfer", from: Address, to: Address]
+    /// If `to` is not a muxed address (i.e. has no muxed_id set), then data
+    /// will  have the following format (the `Transfer` event defined below):
+    /// `data = amount: i128`
+    ///
+    /// If `to` is a muxed address (i.e. has a muxed_id set), then data will
+    /// have be emitted as map with Symbol keys in the following format (the
+    /// `TransferMuxed` event defined below):
+    /// `data = { "amount": i128, "to_muxed_id": u64 }`
```
leighmcculloch: Fyi I've opened a couple issues to improve how we communicate about the shape of the published events:
- #1519 
- #1520

soroban-sdk-macros/src/derive_client.rs
dmkozh: Done.

soroban-sdk-macros/src/derive_client.rs
dmkozh: Done.

soroban-sdk-macros/src/derive_client.rs
dmkozh: Done.

soroban-sdk/src/token.rs
dmkozh: Yeah, we can update this then.

leighmcculloch (approved):
This is great.

This is the first time we're using this pattern, so I'd like to cut an rc after this merges and flag it to folks building tokens in hope we can get some early experimentation with it to identify any issues.

PR #1522, opened 2025-08-06
Feat/has impl

willemneal:
### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]


willemneal:
I agree it is unusual for a rust trait to only expose some methods externally. However, this is already present in your example of `Pause`.  The trait has constraint `PauseHooks` which has `pause_before` that while present in `Pause`'s interface is not present in the exposed methods of the contract:

```rust
#[contracttrait]
pub trait Pause : PauseHooks {
    fn pause(env: &Env, paused: bool) {
        Self::pause_before(env);
        env.storage().persistent().set(&symbol_short!("paused"), &paused)
    }

    fn paused(env: &Env) -> bool {
        env.storage().persistent().get(&symbol_short!("paused")).unwrap_or(false)
    }
}

pub trait PauseHooks {
    fn pause_before(env: &Env);
}
```

So this is really the same as:

```rust
#[contracttrait]
pub trait Pause {
    fn pause(env: &Env, paused: bool) {
        Self::pause_before(env);
        env.storage().persistent().set(&symbol_short!("paused"), &paused)
    }

    fn paused(env: &Env) -> bool {
        env.storage().persistent().get(&symbol_short!("paused")).unwrap_or(false)
    }
    #[internal]
    fn pause_before(env: &Env);
}

```

Then using it

```rust
#[contractimpl]
impl Pause for Contract {
    fn pause_before(_: &Env) {
        //...
    }
}
```

The other use case is where an Extension Trait could be used.

```rust
pub trait PauseExt {
    fn assert_paused(env: &Env);
}

impl<T: Paused> PauseExt for T {
    fn assert_paused(env: &Env) {
        assert_with_error!(env, Self::paused(env))
    }
} 
```


This requires the user to import an extra trait. Of course this is a very common pattern in rust and so I would be okay with not having this. But things get very complex if you combine these two cases.

For example the `AccessControl` trait in the OZ repo, combines both uses. It is also very handy for access control macros to not rely on importing an extra Trait to have access to methods like [`assert_has_any_role`](https://github.com/OpenZeppelin/stellar-contracts/pull/350/files#diff-7855560a0491e61c70f6083fcd906e66c6ca72e9ff7038289373e1d99aebf6d8R405).


The biggest pain point though is when combining with the associated type.

For example, it is useful in a constructor to call the following method in `AccessControl`

```rust
    #[internal]
    fn grant_role_no_auth(e: &Env, caller: &Address, account: &Address, role: &Symbol);
```

If this was in an internal trait like `PauseHooks` we would lose the associated type.


```rust
// Also need an extra import if we only use extension traits
use stellar_access::{AccessControl, AccessControlExt, AccessController, InternalAccessControl};

#[contractimpl]
impl AccessControl for Contract {
    type Impl = AccessController;
}

impl InternalAccessControl for Contract {
 // No type impl. Can only use default methods
}

```

So while we could have a default trait method in `InternalAccessControl`, the implementation of `grant_role_no_auth` should be done by the type that implements `AccessControl` and you would lose the ability to change the implementation of  `InternalAccessControl`.

So even though `AccessControler` implements `InternalAccessControl`, you can't use it to help implement it for `Contract`.

I agree that it does feel strange given how traits work in Rust generally, but it is an opt in feature that I found to to be very useful.


As for 2, I had not considered the idea of generating the trait from the spec. However, the same issue you mention would be true in the first example. The trait is constrained by another trait which the spec would not convey. 

Personally since we are defining the external API while being able to reuse parts of the internal API, it is a flexibility that seems worth it.  Most of the time I've used it is for a version of an extension trait, which is then very useful for assertions. But I found it most useful for things like `metadata_set` or `mint` in the tokens traits. It allows the traits to define the required interface that will be needed internally say by a constructor or a mint method. This way we can reuse the same composability and swap out the base implementation of the trait and not need to update the constructor or leak which implementation was used.

Consider the change in the OZ code. This way the implementation can remain abstracted and not leak implementation details and you can swap out the implementation or extend it.

```diff
#[contractimpl]
impl ExampleContract {
    pub fn __constructor(e: &Env, owner: Address, initial_supply: i128) {
-        Base::set_metadata(e, 18, String::from_str(e, "My Token"), String::from_str(e, "TKN"));
-        Base::mint(e, &owner, initial_supply);
-        e.storage().instance().set(&OWNER, &owner);
+        Self::set_metadata(e, 18, String::from_str(e, "My Token"), String::from_str(e, "TKN"));
+        Self::set_owner(e, &owner);
+        Self::internal_mint(e, &owner, initial_supply);
    }
```

Issue #1517, opened 2025-08-06
Expose new host function `get_address_executable`


leighmcculloch:
[CAP-68] in p23 address a new host function `get_address_executable`.

This needs exposing in the SDK.

Suggestion for where to expose it:

```rust
impl Address {
    pub fn executable(&self) -> Option<ContractExecutable> { ... }
}
```

The `ContractExecutable` type needs expanding to support the additional cases noted in CAP-68. The type should probably be moved from where it currently lives in the `auth` module, into the `address` module. I think it would also be reasonable for it to be renamed to `Executable`.

[CAP-68]: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0068.md

leighmcculloch:
I think it would also be reasonable to surface another function that discards the returned value, to save some CPU instructions for the common case where you just want to know if something exists:

```rust
impl Address {
    pub fn exists(&self) -> bool { ... }
}
```

I've seen at least one case where someone needed to check that a Stellar Asset Contract was already deployed, and ideally they don't have to pay for the enum value to be constructed on the guest side if they aren't going to use it.

PR #1525, opened 2025-08-06
Add a getter function for Address executable.

dmkozh:
### What

Add a getter function for Address executable.

Also added exists() function for convenience/small optimization.

### Why

Exposing new host function introduced in p23.

Resolves https://github.com/stellar/rs-soroban-sdk/issues/1517

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/address.rs
+++ b/soroban-sdk/src/address.rs
@@ -177,8 +177,8 @@ impl From<Address> for ScAddress {
         (&v).into()
     }
 }
-#[cfg(not(target_family = "wasm"))]
 
+#[cfg(not(target_family = "wasm"))]
 impl TryFromVal<Env, ScAddress> for Address {
     type Error = ConversionError;
     fn try_from_val(env: &Env, val: &ScAddress) -> Result<Self, Self::Error> {


soroban-sdk/src/address.rs
```diff
@@ -192,6 +192,14 @@ impl TryFromVal<Env, ScAddress> for Address {
     }
 }
 
+#[contracttype(crate_path = "crate", export = false)]
+#[derive(Clone)]
```
leighmcculloch: Can we add `Debug, PartialEq, Eq, ...`? It would be useful to be able to compare, and Debug is needed for use with `assert_eq!` too.

soroban-sdk/src/address.rs
```diff
@@ -285,6 +293,32 @@ impl Address {
         }
     }
 
+    /// Returns the executable type of this address, if any.
+    ///
+    /// For Wasm contracts, this also returns the hash of the contract code.
+    /// Otherwise, this just returns which kind of 'built-in' executable this is
+    /// (StellarAsset or Account).
```
leighmcculloch, line 301: It's implied but worth calling this out:
```suggestion
    /// (StellarAsset or Account).
    ///
    /// Returns None when the contract or account does not exist.
```

soroban-sdk/src/tests/address.rs
```diff
@@ -55,3 +55,44 @@ fn test_contract_address_conversions() {
...
+
+#[test]
+fn test_get_existing_contract_address_executable() {
+    const EXAMPLE_WASM: &[u8] =
+        include_bytes!("../../../target/wasm32v1-none/release/test_udt.wasm");
+
+    let env = Env::default();
+    let contract_address = env.register(EXAMPLE_WASM, ());
+    let contract_executable = contract_address.executable();
+    assert!(matches!(contract_executable, Some(Executable::Wasm(_))));
```
leighmcculloch: What does the value look like for a native contract deployment without wasm?

soroban-sdk/src/address.rs
```diff
@@ -285,6 +293,32 @@ impl Address {
...
+    pub fn executable(&self) -> Option<Executable> {
+        let executable_val: Val =
+            Env::get_address_executable(&self.env, self.obj).unwrap_infallible();
+        if executable_val.is_void() {
+            return None;
+        }
+        Some(executable_val.into_val(&self.env))
```
leighmcculloch: Does the following not work?
```suggestion
    pub fn executable(&self) -> Option<Executable> {
        let executable_val: Val =
            Env::get_address_executable(&self.env, self.obj).unwrap_infallible();
        executable_val.into_val(&self.env)
```

leighmcculloch (commented):
Couple small tweaks suggested

soroban-sdk/src/address.rs
leighmcculloch: Option<T> is implemented as void -> None, otherwise T -> Some.

soroban-sdk/src/address.rs
dmkozh: I wasn't sure about adding these as this is a contracttype and we don't seem to add anything to the other contract types; I'm not sure if the generated code size is of concern. But added for now.

soroban-sdk/src/address.rs
dmkozh: Done.

soroban-sdk/src/tests/address.rs
dmkozh: It's represented as empty Wasm. Updated the test respectively.

soroban-sdk/src/address.rs
dmkozh: Done.

soroban-sdk/src/address.rs
leighmcculloch: I think we just missed it on the other contract types in hte sdk. There's no code size concern, as Rust is good at not compiling in anything that isn't used.

leighmcculloch (approved):

PR #1526, opened 2025-08-07
fix: ensure that the trait derive macro generate ctors

willemneal:
The contractimpl macro doesn't have all the info needed. So this makes it the responsibility of the trait macro to generate the ctor function.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -291,7 +291,7 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
                     crate_path,
                     ty,
                     trait_ident,
-                    pub_methods.into_iter(),
+                    pub_methods.iter().map(|m| &m.sig),
                 )
             });
             output.extend(quote! { #cfs });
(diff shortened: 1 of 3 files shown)


leighmcculloch:
Thanks, this is a bug in the implementation that's still a WIP. I've got it in the list to look at as part of that work. Cheers!

soroban-sdk-macros/src/lib.rs
```diff
@@ -285,12 +285,15 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
             };
-            let cfs = derive_contract_function_registration_ctor(
-                crate_path,
-                ty,
-                trait_ident,
-                pub_methods.into_iter(),
-            );
+            // derive_contractimpl_trait will handle this since we don't know all the methods here
+            let cfs = trait_ident.is_none().then(|| {
+                derive_contract_function_registration_ctor(
```
leighmcculloch, line 290: I'm not following this part of the change. It wasn't part of the change I did when addressing the issue. Ctors are only setup here for functions implemented manually, and ctors are only published in the other location for default functions not implemented, so I don't _think_ they should collide. What case do they collide?

PR #1527, opened 2025-08-07
Update the migration guide for p22->23.

dmkozh:
### What

Update the migration guide for p22->23.

### Why

Improving documentation.

### Known limitations

N/A


soroban-sdk/src/token.rs
```diff
@@ -288,6 +288,33 @@ pub struct Clawback {
     pub amount: i128,
 }
 
+/// This is a helper function to publish either a `Transfer` or `TransferMuxed`
+/// event based on whether the `to` address is a muxed address or not (i.e.
+/// whether it has non-None ID).
+pub fn publish_transfer_to_muxed_address_event(
+    env: &Env,
+    from: &Address,
+    to: &MuxedAddress,
```
leighmcculloch, line 297: I like how this isn't an `impl Into<MuxedAddress>`, which pushes devs towards supporting `MuxedAddress` instead of calling this new function with an `Address`.

soroban-sdk/src/token.rs
```diff
@@ -288,6 +288,33 @@ pub struct Clawback {
     pub amount: i128,
 }
 
+/// This is a helper function to publish either a `Transfer` or `TransferMuxed`
+/// event based on whether the `to` address is a muxed address or not (i.e.
+/// whether it has non-None ID).
+pub fn publish_transfer_to_muxed_address_event(
```
leighmcculloch, line 294: A function like this used to live in the `soroban-token-sdk`. For example:

https://github.com/stellar/rs-soroban-sdk/blob/f46e9e0610213bbb72285566f9dd960ff96d03d8/soroban-token-sdk/src/event.rs#L20-L23

Maybe we should bring that back, mark it as deprecated, and have it call through to here.

leighmcculloch, line 294: I think the name can be shortened and still retain the same meaning.
```suggestion
pub fn publish_transfer_event(
```

leighmcculloch (approved):
I pushed a change in e1a7add that expands on the contract event example, and because of how long these are, I moved each migration instructions into its own page.  I'd like to move the other v21tov22, etc pages to separate too, so that he main `_migrating` becomes a succinct short list.

Lmkwyt

Approving so we can merge this as is, but I'd also like to keep iterating on the migrate docs, as well as the event docs, early next week.

soroban-sdk/src/token.rs
dmkozh: Renamed.

dmkozh: Should we also add the deprecation/removal of token events in the migration guide then? I would be inclined to do that as a separate PR in any case.

soroban-sdk/src/token.rs
dmkozh: Yeah, I assume the other instances of transfer events should just build the event struct explicitly; at least that's the approach that we seem to promote for all the token events.

soroban-sdk/src/token.rs
leighmcculloch: I'll follow up in:
- https://github.com/stellar/rs-soroban-sdk/issues/1531

leighmcculloch: I don't think the change for this got pushed before merging. I'm renaming it in:
- https://github.com/stellar/rs-soroban-sdk/pull/1533

PR #1530, opened 2025-08-08
Add tests comparing token events to SAC events

leighmcculloch:
### What
  Add tests comparing token events to SAC events. Fixes two bugs in how the events were inconsistently defined.

  ### Why
We've noticed a couple bugs both now and in the long past of the token interface in the SDK. There is currently nothing that validates ti is correct and aligns with the Stellar Asset Contract's or SEP-41. This change checks that the new events are consistent with the SAC's events, minus the fact that the SAC includes an additional topic.

### DIFF
--- a/soroban-sdk/src/tests/token_events.rs
+++ b/soroban-sdk/src/tests/token_events.rs
@@ -8,7 +8,7 @@ use soroban_sdk::{
     testutils::{Address as _, Events as _, MuxedAddress as _},
     token::StellarAssetClient,
     token::{Approve, Burn, Clawback, Mint, Transfer, TransferMuxed},
-    vec, xdr, Address, Env, Event, IntoVal, Map, MuxedAddress, Symbol, Val,
+    vec, xdr, Address, Env, IntoVal, Map, MuxedAddress, Symbol, Val,
 };
 
 #[contract]


soroban-sdk/src/testutils.rs
```diff
@@ -576,4 +585,17 @@ impl StellarAssetContract {
     pub fn issuer(&self) -> StellarAssetIssuer {
         self.issuer.clone()
     }
+
+    pub(crate) fn asset(&self) -> xdr::Asset {
```
leighmcculloch, line 589: I've made these helper functions pub(crate) because we do not expect them to used by devs, and so as not to expand the surface area of these types.

soroban-sdk/src/token.rs
```diff
@@ -551,7 +551,7 @@ pub trait StellarAssetInterface {
     ///
     /// # Events
     ///
-    /// Emits an event with topics `["mint", admin: Address, to: Address], data
+    /// Emits an event with topics `["mint", to: Address], data
```
leighmcculloch, line 554: Doc bug.

soroban-sdk/src/token.rs
```diff
@@ -573,6 +573,8 @@ pub trait StellarAssetInterface {
 
 #[contractevent(crate_path = "crate", topics = ["set_admin"], data_format = "single-value", export = false)]
 pub struct SetAdmin {
+    #[topic]
+    pub admin: Address,
```
leighmcculloch, line 577: Code bug fixed.

soroban-sdk/src/token.rs
```diff
@@ -263,7 +263,7 @@ pub struct TransferMuxed {
     pub from: Address,
     #[topic]
     pub to: Address,
-    pub to_muxed_id: u32,
+    pub to_muxed_id: u64,
```
leighmcculloch, line 266: Code bug that @dmkozh had found which prompted the addition of these comparison tests with the SAC.

soroban-sdk/src/tests/token_events.rs
leighmcculloch, line 1: The tests in this file are pretty verbose, I've opted not to dry them, because afew of them have unique differences in annoying spots that would make drying the tests worse imo.

PR #1533, opened 2025-08-11
Bring back token event fns but deprecated

leighmcculloch:
### What
  Bring back the token event fns but marked deprecated instead of deleted.

  ### Why
  I deleted the event functions in #1489 to encourage use of the new event types, however they should be at least marked deprecated to soften the upgrade path, and help developers immediately find through meaningful deprecated notices what they need to use instead.

### DIFF
--- a/soroban-token-sdk/src/event.rs
+++ b/soroban-token-sdk/src/event.rs
@@ -1,9 +1,10 @@
-use soroban_sdk::{symbol_short, Address, Env, EnvBase, IntoVal, MuxedAddress, Symbol};
+use soroban_sdk::{symbol_short, Address, Env, Symbol};
 
 pub struct Events {
     env: Env,
 }
 
+#[allow(deprecated)]
 impl Events {
     #[inline(always)]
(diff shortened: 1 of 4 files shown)


soroban-token-sdk/src/event.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl Events {
+    #[inline(always)]
+    pub fn new(env: &Env) -> Events {
+        Events { env: env.clone() }
+    }
+
+    #[deprecated = "use soroban_sdk::token::Approve::publish"]
+    pub fn approve(&self, from: Address, to: Address, amount: i128, expiration_ledger: u32) {
```
leighmcculloch, line 15: For each of the functions I have simply brought back the old functionality that was deleted in https://github.com/stellar/rs-soroban-sdk/pull/1489, and haven't rewritten the functions to use the new types. I think it'd be better to not introduce change here, and in the next major lets delete these functions.

dmkozh (approved):

PR #1534, opened 2025-08-11
Lean into Transfer events with data as a map

leighmcculloch:
### What

Remove the `token::publish_transfer_event` function.

Rename Transfer to TransferLegacy.

Rename TransferMuxed to Transfer.

### Why

To encourage developers publishing the newer and easier to extend Transfer event.

While it's true that the Stellar Asset Contract, for largely legacy reasons, continues to publish a `transfer` event with the amount as the data, there's no reason for the SDK to encourage other token implementations to do the same.

Rather than giving developers a function to call to publish the transfer event across these two different data forms, we should just point them to publishing the newer form that is extendable.

### Merging

This PR is intended to merge to `main` after #1533. It is marked as a draft PR until #1533 is merged.


soroban-sdk/src/_migrating/v23_token_transfer.rs
```diff
@@ -55,9 +51,13 @@
 //!    let to = muxed_to.address();
 //!    // Token-specific implementation of balance movement (same as before).
 //!    token_impl::move_balance(&env, &from, &to, amount);
-//!    // Publish an appropriate transfer event that includes the muxed ID
-//!    // when it's non-None.
-//!    token::publish_transfer_event(&env, &from, &muxed_to, amount);
+//!    // Publish the transfer event.
+//!    token::Transfer {
```
dmkozh, line 55: My main concern about this is that this would lead to confusion when emitting transfer events not from the transfer. Though probably the helper function would lead to the same sort of confusion, so maybe it's fine.

dmkozh (approved):

soroban-sdk/src/_migrating/v23_token_transfer.rs
leighmcculloch: I think there's a high chance of confusion whatever we do here.

PR #1535, opened 2025-08-11
Allow ref (&) parameters in contract fns

leighmcculloch:
### What
Allow contract functions parameters to be immutable references, rather than only owned values.

### Why

This change is primarily about making it more convenient to write contract functions like typical Rust functions. In Rust function parameters can be of a few variations, they can be an owned value or a reference/borrow, and can be immutable or mutable. Today contract function parameters must be owned and immutable. 

Reference parameters are useful when needing to pass around a value that the caller would like to retain ownership over. Mutable parameters are helpful when the function will mutate the parameter making the mutated value available to the caller.

It is common when writing small composable Rust functions to use ref and mut as needed.

When writing contract functions the composability of the functions being written is restricted. Owned values must be used, and this leads to situations where functions must received cloned values.

In contracts cloning SDK types when passing them around is not a performance concern since all SDK types are a small integer value. But it is inconvenient and it makes contracts look less like regular Rust code when there is an abundance of `.clone()` for no apparent reason other than the contract functions don't let you.

Is this a big issue? No, not particularly. But in general the soroban-sdk tries to let people write Rust how they'd write Rust elsewhere. And this change lifts a limitation that isn't really required.

### Why not

This change introduces variance into the trait-compatible interface that implementations can take. For a contract interface that is intended to be standard, one implementation might use all owned values, another might use all refs, and another might use any combination. This doesn't affect a contracts ability to be called because this information does not make it into the exported interface. But if someone was to share a Rust trait to match one implementation, there's a good chance it would fail to match other implementations. Technically variance already exists today. The `env: Env` parameter is already optional, and can already take either an owned `Env` or a ref `&Env`. So the ship already sailed from day one.

You could argue that the top-level entry point value in a program should always be owned. Someone has to own every value. The fact that contract functions do not need to own the value is because there's actually an outer function that gets exported that is truly the entry point and that function can own the value. But in some ways that's a bit odd.

### Merging

This change can be merged at anytime and can be released in a minor release. It does not change or limit any existing behaviour and only adds new behaviour that is disallowed today.

### Thanks

Thanks to @willemneal contributing #1521 that this change is derived.

### Todo
- [x] Test the change on soroban-examples. https://github.com/stellar/rs-soroban-sdk/pull/1591
- [x] Test the change on several complex ecosystem contracts.
- [x] Test with #1544 
- [x] #1606


leighmcculloch:
> but it would be amazing if `&mut Env` would be able to determine whether a method could write to storage/publish events

Agreed, but I think that's a bigger thing to tackle, and the timing isn't right. Lots of complexity, and with unclear benefits. There's a discussion over here discussing the idea, but I am not chasing it in the near future:
- https://github.com/stellar/stellar-protocol/discussions/1454

soroban-sdk-macros/src/derive_fn.rs
```diff
@@ -82,7 +82,14 @@ pub fn derive_pub_fn(
                     ty: Box::new(Type::Verbatim(quote! { #crate_path::Val })),
                 });
                 let passthrough_call = quote! { #ident };
+
+                let call_prefix = match *pat_ty.ty {
+                    Type::Reference(TypeReference { mutability: Some(_), .. }) => quote!(&mut),
```
leighmcculloch: 🤔 Do we need to insert the lifetime in here?

leighmcculloch: No I guess not. Because the lifetime shouldn't be included at the call site: 2046b000c9c13be06b14c9715bf976b015a7952b

leighmcculloch:
@stellar/contract-committers This change is ready for review. I'd like to get more than one set of eyes on this feature, both its soundness and the benefits before merging, because at this stage it's involved very few people and there hasn't been many asks driving this. When reviewing please take a look at the why not section in the PR. All things considered I still think the change is worth making because it removes what is a somewhat arbitrary limitation, but I'm interested in hearing everyones thoughts.

dmkozh:
I'm a bit hesitant about this change. While I agree that trait consistency already might be an issue, there are just 3 options and realistically only 2 of them are 'real' (`Env` and `&Env`, specifying a trait without `Env` seems really problematic). Adding this functionality creates an exponential number of different ways of exposing the same thing.

But I'm even more concerned about the contract interface usage patterns. `&mut` arguments seem particularly problematic to me. Contracts can not modify the input values (and host values are in general immutable). This creates a wrong impression that it's possible to use contract arguments for returning the values from the contract call (which is how `&mut` args are typically used in general), and to make things worse this will _sometimes_ be the case (e.g. if you pass a value from the unit test to a client), which might result in incorrect developer expectations in terms of contract interoperability and interface design.

`mut` arguments seem less problematic, but also it's a very minor syntactic sugar that only matters for the contract implementation, so I'm really not sure if this really provides much benefit vs the added complexity. Also, given that contract client interface is by definition public, I find it a bit weird for it to 'leak' the implementation details this way. 

As for the const references, I agree that they do solve a more significant real issue without adding too much confusion (there is still _some_ confusion though due to unclear ownership in case of root-level contract calls).  However, I think a more appropriate solution for this issue would have been making the SDK types `Copy`, though I realize this has its own set issues.

willemneal:
From my experience with OZ's codebase and others, they have a pattern of having two sets of interfaces to avoid the cloning:

```rust
trait Trait {
    fn foo(e: &Env, faa: String){
        ImplementingType::foo(&faa)
    }
}

```
This leads to types that don't actually implement the outer trait. 

I agree that adding mutable refs is a strange pattern and doesn't add much.

As for adding Copy, I have often considered it, but while cheap there is a cost to cloning a host reference, so it is best not to hid it for new devs and encourage references instead.

leighmcculloch:
The motivation for this change is writing natural rust, composable functions, and avoiding having the compiler say "you're not allowed to use refs" when as a Rust devs that's what I might naturally otherwise do.

leighmcculloch:
> But I'm even more concerned about the contract interface usage patterns. `&mut` arguments seem particularly problematic to me. Contracts can not modify the input values (and host values are in general immutable). This creates a wrong impression that it's possible to use contract arguments for returning the values from the contract call (which is how `&mut` args are typically used in general), and to make things worse this will _sometimes_ be the case (e.g. if you pass a value from the unit test to a client), which might result in incorrect developer expectations in terms of contract interoperability and interface design.

👍🏻 This is a good observation. Thanks.

> given that contract client interface is by definition public, I find it a bit weird for it to 'leak' the implementation details this way.

This change doesn't affect the client interface. The client interfaces only have parameters as references, they are never passed as owned values, or mut values. So as a point, the client interface already differs from the implementation interfaces. They have never matched and there's little value in making them match. Clients offer additional functions like the `try_` variants.

dmkozh:
> Copy is not an option, because most of the SDK types embed the Env which is not Copy in the test environment. It can't be Copy when built for wasm and not Copy when built for test.

I'm not sure I understand this, what prevents us from adding Copy to Env,? I think that would make sense semantically given that SDK types are indeed very cheap to copy, as you've pointed out in the next post.

> The client interfaces only have parameters as references, they are never passed as owned values, or mut values. So as a point, the client interface already differs from the implementation interfaces. They have never matched and there's little value in making them match. Clients offer additional functions like the try_ variants.

True, I forgot that this mismatch is already the case. I think this difference is a bit unfortunate. While the client interface indeed isn't be identical to the contractimpl used to generate it at least due to `try_` versions of methods, the `contractimpl` or trait definition itself is probably the most obvious thing to look at in order to understand the contract interface without involving rust-analyzer. So maybe it would have been better to try to at least keep the argument types consistent. At this point, maybe `&` and `mut` are alright, but I still really dislike the `&mut` as it's misleading for the reader.

dmkozh (approved):

PR #1536, opened 2025-08-12
Separate token/asset specs and events from SDK into dedicated crates

### LINKED ISSUE #1532: Token event specs will not be exported in contracts that use the new event types
Token contract events are not included in contract spec if merely used from the SDK.

Specs are written to the wasm file by code injecting a static variable containing the XDR bytes, marked with an attribute to indicate which custom section the bytes should be appended to.

So whether specs are included or not is decided at the time when the type is defined, and the macro on the type executes at compile time, irrespective of whether the type gets used or not.

Types inside the soroban-sdk are marked with `export = false`, to tell the macro not to mark the spec XDR variable to be included in the wasm custom section. This is because a large variety of applications can import the SDK, and it would not be appropriate for an application not using the types to have them included in their spec. So when contract functions expose an SDK type it's a bit of a hack. The type isn't actually exposed in the spec like it would be if the contract had defined the type itself. In all cases where this matters the SDK can recognise the referenced type in a spec even if it's type is not included. It's a hack, but it works around the limitations.

But, this presents a problem where events defined don't get referenced in a contract function definition, so the SDK doesn't surface the use or presence of the events any other way.

We need to rethink about how the token contract events are defined so that they get surfaced into the specs of any tokens that use them.

When posting ideas for how to solve this, please label them with a number for easy and not-confusing reference.

cc @dmkozh @stellar/contract-committers

leighmcculloch:
### What
Move token and asset specs into dedicated crates. Move token events into the soroban-token-sdk crate.

### Why
Accessing utilities for publishing token events have always lived in the soroban-token-sdk. The soroban-token-sdk is what folks can import if they want utilities for building a token. The token utilities in the soroban-sdk are intended to only support calling tokens.

We want the generated token and asset spec to include the events, and since the events aren't defined in the soroban-sdk the spec gen needs to live outside the soroban-sdk because the soroban-sdk can't depend on the soroban-token-sdk. Creating new crates to collate and expose the generated spec XDR makes a lot of sense in any case, since it isolates that functionality into crates dedicated for that purpose rather than clouding the API of the main soroban-sdk.

Tools that need the spec, such as the stellar-cli, can now call `soroban_token_spec::xdr()` or `stellar_asset_spec::xdr()` instead of calling `soroban_sdk::token::StellarAssetSpec::spec_xdr()`.

Token developers using token event publishing utilities continue to use `soroban-token-sdk`.

Close #1532


.github/workflows/rust.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       with:
         name: cargo-semver-checks
         version: 0.42.0
-    - run: cargo semver-checks
+    - run: cargo semver-checks --exclude '*-spec'
```
leighmcculloch: There's some funky stuff with the `soroban-token-spec` crate, because it's technically been published before a long time ago, and the tool is messing up looking it up. After the next stable release this can be removed.

soroban-token-sdk/Cargo.toml
```diff
@@ -11,8 +11,10 @@ edition = "2021"
 rust-version.workspace = true
 
 [lib]
-crate-type = ["cdylib", "rlib"]
```
leighmcculloch, line 14: The cdylib is a hold over from when we used to use this lib as a way to build a wasm to get the contract spec, which we don't do anymore. The 'rlib' is unnecessary.

soroban-token-sdk/src/event.rs
```diff
@@ -11,45 +11,59 @@ impl Events {
...
+        #[contractevent(data_format = \"single-value\")]\n\
+        pub struct SetAdmin {\n\
+            #[topic]\n\
+            admin: Address,\n\
+            new_admin: Address,\n\
+        }"]
     pub fn set_admin(&self, admin: Address, new_admin: Address) {
         let topics = (symbol_short!("set_admin"), admin);
         self.env.events().publish(topics, new_admin);
     }
```
leighmcculloch, line 64: The `set_authorized` and `set_admin` events should never have been in the token sdk, as they're events published only by the Stellar Asset Contract, not part of SEP-41 Tokens. So the deprecation notice tells people how to define their own events that produce the same output.

dmkozh (approved):

PR #1550, opened 2025-08-19
Update copilot-instructions.md

leighmcculloch:
### What

Remove unused deps from Copilot instructions: rust-src, cargo-watch, cargo-nextest.

### Why

They aren't required tools.

Telling Copilot to use rust-src is likely to send it on tangents we don't want it to take. Recompiling rust-src and relying on features that Rust provides only through rust-src would harm the developer experience of folks using the soroban-sdk, requiring them to also use rust-src which is a not a common Rust developer experience.

cargo-watch is a deprecated tool and not recommended to use. Unclear to me why Copilot would benefit from it since it is more likely to benefit from choosing to run appropriate test commands after each change, which the instructions already detail.

cargo-nextest is not needed. It may speed up test runs, but we don't use it in the Makefile and Copilot will benefit from using the test targets in the Makefile to ensure that the test vectors build before tests.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -11,15 +11,11 @@ Always reference these instructions first and fallback to search or bash command
   - `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
   - `rustup install stable`
   - `rustup +stable target add wasm32v1-none`
-  - `rustup +stable component add rust-src`
 - Install Rust nightly (required for documentation):
   - `rustup install nightly`
   - `rustup +nightly target add wasm32v1-none`
-  - `rustup +nightly component add rust-src`
 - Install required cargo tools:
   - `cargo install --locked cargo-hack`
-  - `cargo install --locked cargo-nextest`
-  - `cargo install --locked cargo-watch`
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -11,15 +11,11 @@ Always reference these instructions first and fallback to search or bash command
   - `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
   - `rustup install stable`
   - `rustup +stable target add wasm32v1-none`
-  - `rustup +stable component add rust-src`
```
dmkozh, line 14: Copilot didn't actually hallucinate these instructions: we do list these deps in the CONTRIBUTING.md (e.g. https://github.com/stellar/rs-soroban-sdk/blob/daa17248a5266223bc3d5a0986ff33b739316731/CONTRIBUTING.md?plain=1#L29). Shall we clean that up as well?

leighmcculloch: b021350407390d4b1bd4c5a5cc8cc30cad2ae9a5

anupsdf (approved):

Issue #1554, opened 2025-08-26
Complete event definitions for tokens and SAC


leighmcculloch:
The token event definitions included in `soroban-token-sdk`, `soroban-token-spec`, and `stellar-asset-spec` crates are missing the event definitions for:

- Mint when the data is a map containing a muxed ID. This event should be something contracts can create, so should be visible in the `soroban-token-sdk`.
 
- Mint and transfer with other memo types. These should only be included in the `-spec` crates and not `soroban-token-sdk` because contracts shouldn't have a need to create them.

PR #1555, opened 2025-08-27
Add mint event with mux support and other memo types to specs

leighmcculloch:
### What
  Add a new Mint event with muxed address support. Add events to specs for other memo types that can occur with payments from classic.

  ### Why
  The mint event should be defined so that contract developers can choose to emit a mint event to a destination with a muxed id.

The specs should contain all of the event structures that an ingester could potentially see, including when other memo types are in use.

### Tests

It wasn't possible to add comparison tests for the mint muxed event, or the mint memo string and bytes to_muxed_id types, to confirm the events are identical. This is because the soroban-env doesn't allow us to trigger those events because they are only generated by the classic transaction processing.

For this reason @dmkozh @sisuresh please take great care to review the events being added. I've written them based on CAP-67, but we've had bugs in the past arise out of out-dated CAPs, so your help in validating they match core's implementation is much appreciated.

Close https://github.com/stellar/rs-soroban-sdk/issues/1554


soroban-token-spec/src/lib.rs
```diff
@@ -2,6 +2,56 @@
...
+pub struct TransferWithMuxedString {
+    #[topic]
+    pub from: Address,
+    #[topic]
+    pub to: Address,
+    pub to_muxed_id: Option<String>,
+    pub amount: i128,
+}
+
+/// Mint event published when a classic payment uses the MEMO_HASH or MEMO_RETURN memo type.
```
leighmcculloch:
```suggestion
/// Transfer event published when a classic payment uses the MEMO_HASH or MEMO_RETURN memo type.
```

soroban-token-sdk/src/events.rs
```diff
@@ -37,9 +37,17 @@ pub struct Burn {
 }
 
 #[contractevent(topics = ["mint"], data_format = "single-value")]
+pub struct MintLegacy {
```
dmkozh: I don't think this should be called legacy, at least not with the current way SEP-41 is phrased. I think at least for now we could just leave mint as is and have the classic-related specs defined outside of the SDK.

soroban-token-spec/src/lib.rs
```diff
@@ -2,6 +2,56 @@
...
+use soroban_sdk::{contractevent, Address, BytesN, String};
+
+// Events that are a special case generated by the Stellar Asset Contract, but are part of SEP-41,
+// but are not expected to be generated by other contracts. Defined here privately only for the
+// purpose of generating the complete contract spec.
+
+/// Transfer event published when a classic payment uses the MEMO_TEXT memo type.
+#[contractevent(topics = ["transfer"], data_format = "map", export = false)]
+#[doc(hidden)]
+pub struct TransferWithMuxedString {
```
dmkozh, line 14: Could these go into stellar-asset-spec instead to signify that they only concern Stellar assets?

leighmcculloch: As much as possible for the movement of value there needs to be no fragmentation for how to ingest and understand the movement of value. Someone should be able to understand the movement of value using a parser built from the token spec, which is why these events need to be defined in SEP-41. We can say in SEP-41 that these events, when to muxed id can be String or Bytes, exist in the standard for legacy applications, and contracts should not use them, whilst consumers of events will need to understand them.

The other events defined only in the asset spec are about the admin and access and nothing to do with the movement of value.

soroban-token-sdk/src/events.rs
leighmcculloch: I followed the same pattern as what we did with transfers. Ideally there's one way we suggest to devs to emit a mint event, and any other representations are there for legacy use cases. This narrative would be easier if we had chosen fro the SAC to use a single mint and transfer event structure and hadn't retained the old way for some use cases.

I'm open to other names, but similar to my other comment below we can't remove it from the token spec, as any parser of tokens needs to understand the event.

dmkozh: But calling this 'legacy' creates a wrong impression that mint events should support memos, which is not what we wanted to achieve. The situation is different from the transfer here, because transfers should ideally support muxed destinations.

leighmcculloch: Why can't contract tokens support mint to custodial accounts?

leighmcculloch: The main reason to refer to the amount data events as legacy is that they aren't extendable. Where-as the data variant is.

soroban-token-spec/src/lib.rs
dmkozh: I don't fully understand the parser argument. If a parser is going to process the asset events, then it will need to have some code that's not entirely compatible with the SDK (because of the different address space), i.e. it can't just use these Rust definitions as is in Rust. If this is just used as a template for generating the XDR spec, then I'm not sure there is a strong argument for this to be comprehensive in the library used by the contracts - the consumer will need to import the XDR specs, not the Rust specs, so I'm not sure it matters where exactly the specs come from. Is there some issue with building parser for SAC events using the SAC contract spec?

> The other events defined only in the asset spec are about the admin and access and nothing to do with the movement of value.

I understand that it's the case now, but since the spec contains the events now, it seems logical to me to define the SAC-specific events in the SAC spec.

Given that docs are hidden for these events, I would expect the chance of tripping up the token developers with the definitions is rather low, but I would like to at least understand the issue of not being able to use the SAC spec for parsing.

leighmcculloch: The argument for why these events should be part of the token interface, not bespoke events for the asset interface, is the same reason for why we added the mint event to SEP-41. An integrator consuming value events can't get a unified picture of the movement of value without ingesting both asset and token events, and the token events should be the common ground to track value.

soroban-token-sdk/src/events.rs
dmkozh: I guess they can, but the 'legacy' name implies that they can't _not_ support it. Maybe it's fine to make this the default mint event spec in SEP-41, though it seems inconsistent with the SAC interface that we have now. 
We specifically wanted to limit the muxed id surface to the minimum necessary use case, which is why only specifically transfer supports it. With 'why could they not support muxed address in X operation' argument we could basically arrive at switching each address to be muxed (which is basically what has been done for Classic). I'm not saying that this philosophy is necessarily wrong, but we've agreed on a different philosophy for CAP-67, and it seems a bit weird to diverge from it now.
I don't want to block this much, but I would've been happier with something more neutral like `Mint` and `MintMuxed` or something like that, i.e. not explicitly deprecating the existing spec just yet.

soroban-token-spec/src/lib.rs
dmkozh: Memos are not necessary for the value tracking, and I don't think a generic integrator should look at memos at all, as these are only defined for a specific consumer that knows what they mean. If we move towards map-based value payload, then I think it would be wise for integrator to only look at a specific field that defines the value, and for SEP-41 to specify that value is always in `amount` field of map payload and there may be other optional fields, defined for a specific token implementation. The integrator can then choose what to do with these fields, but in any case they can track the value generically. I don't know though if that's possible to express now with the token schema, but that's a different issue.
Standardizing implementation specifics of SAC (which are basically implementation specifics of the classic protocol) as a part of SEP-41 seems to point at poor abstraction choices.

leighmcculloch: > Standardizing implementation specifics of SAC (which are basically implementation specifics of the classic protocol) as a part of SEP-41 seems to point at poor abstraction choices.

This is the reason we added the clawback event to SEP-41, even though SEP-41 has no self defined mechanics for clawback. Without defining clawback in SEP-41 it is impossible to get a complete picture of the movement of value of a Stellar asset when implementing a consumer of SEP-41. The same principle applies here? The muxed ID is part of the destination identity.

I think what we're debating is more what should be in SEP-41 than an implementation detail here, so lets move conversation to:
- https://github.com/stellar/stellar-protocol/pull/1775

soroban-token-sdk/src/events.rs
leighmcculloch: We need a better name instead of legacy, and a way to differentiate. Should we call the legacy variant "AmountOnly"? e.g. `TransferWithAmountOnly`, `MintWithAmountOnly`?

dmkozh: This seems fine to me as well.

leighmcculloch: Renamed in fd278e2.

dmkozh (approved):

PR #1556, opened 2025-08-28
Scope generated fn code mod by contract type name

leighmcculloch:
### What
  Change the name of the generated code module that contains the generated functions for calling the contract, so that the name includes the contract type's name.

  ### Why
  Prevents naming conflicts between functions with the same name across different contract implementations. This is particularly helpful in tests and examples where a developer may legitimately define two contracts, that are intended to be used independently, but exist in the same namespace. Today those contracts will see a compile error if the contracts share the same name, which is much more of a common problem now that constructors exist and have a common name.

This problem was reported by @ozgunozerk.

### Trying it out

Edit your Cargo.toml to include the patch:

```toml
[patch.crates-io.soroban-sdk]
git = "https://github.com/stellar/rs-soroban-sdk"
branch = "allow-coexisting-contractimpls-with-same-names"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1003,7 +1003,7 @@ dependencies = [
 
 [[package]]
 name = "proc_macros"
-version = "23.0.0-rc.3"
+version = "23.0.1"
 dependencies = [
  "quote",
  "syn 2.0.103",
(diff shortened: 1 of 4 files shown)


dmkozh (approved):

leighmcculloch:
@ozgunozerk this fix is available in 23.0.2.

PR #1558, opened 2025-08-28
Break circular dependency between soroban-sdk and soroban-token-sdk.

dmkozh:
### What

Break circular dependency between soroban-sdk and soroban-token-sdk.

We only introduced it in order to be able to run the doctest in migration guide soroban-sdk. I've moved the migration guide to the respective crate (soroban-token-sdk). This may slightly reduce discoverability, but I've left a reference to it in the main SDK migration guide.

### Why

It's apparently impossible to publish crates with circular dependency between them, even the dependency only occurs in dev-dependencies (or at least I wasn't able to quickly find a way around that).

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/_migrating.rs
+++ b/soroban-sdk/src/_migrating.rs
@@ -3,7 +3,7 @@
 //! 1. [`contractevent` replaces `Events::publish`][v23_contractevent]
 //!   
 //! 2. [`MuxedAddress` replaces `Address` as the `to` of the `TokenInterface::transfer`].
-//!    This only `soroban-token-sdk`` and is documented in detail in
+//!    This change concerns `soroban-token-sdk` and is documented in detail in
 //!    `soroban-token-sdk` crate migration guide.
 //!
 //! # Migrating from v21 to v22


soroban-sdk/src/_migrating.rs
```diff
@@ -2,7 +2,9 @@
 //!
 //! 1. [`contractevent` replaces `Events::publish`][v23_contractevent]
 //!   
-//! 2. [`MuxedAddress` replaces `Address` as the `to` of the `TokenInterface::transfer`][v23_token_transfer]
+//! 2. [`MuxedAddress` replaces `Address` as the `to` of the `TokenInterface::transfer`].
+//!    This only `soroban-token-sdk`` and is documented in detail in
```
graydon: "This only" .. exists? is found in? Not sure what this means to say.

graydon (changes requested):
Minor nit about the comment text otherwise looks ok.

soroban-sdk/src/_migrating.rs
dmkozh: Must have messed up during editing, updated.

graydon (approved):

Issue #1544, opened 2025-08-19
Test Improvements


leighmcculloch:
Recently we had some bugs get merged into the SDK:

- https://github.com/stellar/rs-soroban-sdk/pull/1530
- https://github.com/stellar/rs-soroban-sdk/pull/1528
- https://github.com/stellar/rs-soroban-sdk/pull/1518
- https://github.com/stellar/rs-soroban-sdk/pull/1510

The bugs were detected while the SDK was in release-candidate status.

We have already improved testing for the first bug. For the first we added tests that compare the code with the SAC, as a way to ensure consistency.

The other two point to a theme where generated code is not as easy to test, because of the huge variety in ways that generated code can interact and compose with other code that the developer is writing.

I suggest we take on a few improvements to testing, some which will improve the above cases, and others that we've talked about at some point and just never done.

- [x] 1. Integration test every PR of the soroban-sdk with the soroban-examples repo, but patching in the sdk into the soroban-examples repo and then running the repos tests and builds ignoring deprecated warnigns. The sdk is _always_ intended to be backwards compatible with at least one major version, so this test should _always_ pass no matter what protocol change is coming downstream. There will probably be exceptions to this where a break will be required in a single version change, but that should be rare. https://github.com/stellar/rs-soroban-sdk/pull/1591

- [x] 2. Integration test on some schedule the soroban-sdk with ecosystem contracts that help identify issues early. For example, integration testing with the OpenZeppelin contracts, and the Axelar libraries, and the Blend and Soroswap contracts, and more. Some of these libraries are already the first place we go to test substantial changes to the macros and automating that will benefit. https://github.com/stellar/rs-soroban-sdk/pull/1595

- [x] 3. Store generated code snapshots alongside test vectors. Test vectors exist in this repo and we could store snapshots of what they look like with their macros fully expanded. Committing those and reviewing those in PRs will give us greater confidence as to how macros work in a variety of contexts. https://github.com/stellar/rs-soroban-sdk/pull/1572

- [x] 4. Consider new test vectors to add to better represent the landscape of the contexts macros get used in. For example:
     - [x] https://github.com/stellar/rs-soroban-sdk/pull/1598
     - [x] https://github.com/stellar/rs-soroban-sdk/pull/1600
     - [x] https://github.com/stellar/rs-soroban-sdk/pull/1601

[...]

PR #1591, opened 2025-10-13
Add workflow to test with soroban-examples

leighmcculloch:
### What
  Add a GitHub Actions workflow that tests rs-soroban-sdk changes against the soroban-examples repository by patching the SDK dependency, building, and running tests.

  ### Why
  This ensures SDK changes don't break existing example contracts and provides early detection of compatibility issues before release.

The soroban-examples are a good extended set of test vectors to test upcoming changes in the sdk. I often manually test PRs against soroban-examples looking for breakages. This codifies that process.

Deprecations are allowed via an env var when runnin the builds and tests in soroban-examples, because it is permitted for the sdk to deprecate functionality and later remove that functionality. The soroban-examples should always stay within at least one major version, and the sdk should generally try to give deprecations at least one full major version before being removed. Meaning that the sdk should generally never break soroban-examples as long as they are within one major version.

The new workflow does not block pull request merges, giving us an opportunity to see how practical it is to run these tests with sdk changes. We can make it a required build when we're happy with how the initial period goes.

For #1544

### DIFF
--- a/.github/workflows/test-with-soroban-examples.yml
+++ b/.github/workflows/test-with-soroban-examples.yml
@@ -62,6 +62,7 @@ jobs:
 
     - name: Patch SDK versions
       run: |
+        # TODO: Update this patch logic to use `cargo add` once this issue is resolved: https://github.com/rust-lang/cargo/issues/16101
         crates=$(cd ${{ github.workspace }}/rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
         find . -name Cargo.toml | while read file; do
           echo Patching "$file" ...


.github/workflows/test-with-soroban-examples.yml
```diff
@@ -0,0 +1,102 @@
...
+        crates=$(cd ${{ github.workspace }}/rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
+        find . -name Cargo.toml | while read file; do
+          echo Patching "$file" ...
+          dir=$(dirname "$file")
+          for crate in $crates; do
+            rel_path=$(realpath --relative-to="$dir" ${{ github.workspace }}/rs-soroban-sdk/$crate)
+            sed -i 's|'"$crate"' = "\([^"]*\)"|'"$crate"' = { path = "'"$rel_path"'" }|g' "$file"
+            sed -i 's|'"$crate"' = {.*version = "[^"]*"\(.*\)|'"$crate"' = { path = "'"$rel_path"'" \1|g' "$file"
+          done
+        done
```
leighmcculloch, line 75: TODO: Update this patch logic to use `cargo add` once this issue is resolved: https://github.com/rust-lang/cargo/issues/16101

PR #1593, opened 2025-10-15
Add compile error for event names exceeding limit

### LINKED ISSUE #1592: `#[contractevent]` macro panics on `unwrap()`
`23.0.2` for `soroban-sdk`

### What did you do?
`#[contractevent]` macro panics  on `unwrap()` when event struct's name exceeds `32` characters(the limit for `Symbol`)

A macro invocation error as it happens for fields of `#[contracttype]` types, e.g.:
`struct field name is too long: 39, max is 30`

### What did you see instead?

Image

Image

leighmcculloch:
### What
  Add compile-time validation that event names do not exceed 32 characters. Return a descriptive error message showing the actual length and the limit when validation fails.
  
  #### Before
```
error: custom attribute panicked
  |
8 |     #[contractevent]
  |     ^^^^^^^^^^^^^^^^
  |
```

#### After
```
error: event name has length 44 greater than length limit of 32
  |
9 |     pub struct MyEventIsTooLongAndLongAndLongAndLongAndLong {
  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

  ### Why
  Event names are constrained by the ScSymbol type to 32 characters maximum, but previously this constraint was only enforced by a call to .unwrap(), which emits a very unhelpful message.

### Testing
The SDK doesn't currently have the infra setup to do compile error testing. It used to. We removed it because it was crazy slow. It might come back as part of #1544. So no test is being included with this fix. However, in the first commit (1445bae4d9f8ce960ce478d6fe4b04b4b99e93e5) I am including a test that will fail at compile. Look at the logs for that commit and you'll see it. That test will be removed before merging.

Fix #1592

### DIFF
--- a/soroban-sdk/src/tests/contract_event.rs
+++ b/soroban-sdk/src/tests/contract_event.rs
@@ -4,16 +4,6 @@ use crate::{
 };
 
 #[test]
-fn test_name_too_long() {
-    #[contractevent]
-    pub struct MyEventIsTooLongAndLongAndLongAndLongAndLong {
-        #[topic]
-        name: Symbol,
-        value: Symbol,
-    }
-}
-
-#[test]
 fn test_defaults() {
     let env = Env::default();
 


soroban-sdk-macros/src/derive_event.rs
```diff
@@ -86,6 +86,19 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
     let (gen_impl, gen_types, gen_where) = input.generics.split_for_impl();
     let path = &args.crate_path;
 
+    // Check event name length
+    const EVENT_NAME_LENGTH: u32 = 32;
+    let event_name = input.ident.to_string();
+    let event_name_len = event_name.len();
+    let event_name: StringM<EVENT_NAME_LENGTH> = errors
```
leighmcculloch, line 93: Fyi we do this dance of defining the size, using it on a `StringM`, then wrapping it in the `ScSymbol`, as a way to include the max length in the error string while getting the compilers help to make sure that the number written here, `32`, always stays consistent with the `SySymbol`'s max length on its `StringM`. If `ScSymbol` ever changes max length the code here will fail to compile because `StringM<32>` will not be assignable to `StringM<NEW_MAX>`.

PR #1597, opened 2025-10-17
Run tests on macos and windows only on merge and push

leighmcculloch:
### What
  Add a setup job that dynamically generates the system matrix for the rust ci workflow. Use only ubuntu-latest for pull requests and the full set of targets (Linux, macOS, Windows across x86_64 and aarch64) for other events (merge group, push).

  ### Why
  Pull requests only need validation on a single platform to provide fast feedback, while the test runs on merge (the merge_group event) can catch any rare issues we create that are platform specific, and same with releases and main branch builds triggerd by push events.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,7 +65,7 @@ jobs:
             }
           ]'
         fi
-        echo "sys=$sys" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


graydon (approved):
fascinating -- I did not know this is possible. you're really pushing the envelope with gh actions!

PR #1618, opened 2025-11-13
Implement Poseidon, Poseidon2

jayz22:
### What

Implements Poseidon, Poseidon:
- two new hash functions under `Crypto` which performs Poseidon/Poseidon2 hashing. Underneath they wrap the permutation host function call with a sponge implementation
- provide reference parameters for Poseidon/Poseidon2 for bn254
- results match both circom (for Poseidon) and noir (for Poseidon2) with test cases
- also expose the two permutation host functions in `CryptoHazmat`, to support custom sponge implementation / parameter sets.

### Why

### Known limitations

For Poseidon only supports `hash2` (t=3, hashing two inputs into one output) for now. 
Only support BN254 for now.
To add support for the rest, need to generate and import those parameters, which will be done in a followup.

### DIFF
--- a/soroban-sdk/src/crypto/poseidon2_sponge.rs
+++ b/soroban-sdk/src/crypto/poseidon2_sponge.rs
@@ -16,6 +16,8 @@ pub struct Poseidon2Sponge {
 
 impl Poseidon2Sponge {
     pub fn new(env: &Env, iv: U256, field: Symbol) -> Poseidon2Sponge {
+        // TODO: we need to set up parameters for BLS12-381
+        assert_eq!(field, soroban_sdk_macros::internal_symbol_short!("BN254"));
         let mut state = vec![env];
         for _ in 0..RATE {
             state.push_back(U256::from_u32(env, 0));
--- a/soroban-sdk/src/crypto/poseidon_sponge.rs
+++ b/soroban-sdk/src/crypto/poseidon_sponge.rs
@@ -16,6 +16,8 @@ pub struct PoseidonSponge {
 
 impl PoseidonSponge {
     pub fn new(env: &Env, iv: U256, field: Symbol) -> PoseidonSponge {
+        // TODO: we need to set up parameters for BLS12-381
+        assert_eq!(field, soroban_sdk_macros::internal_symbol_short!("BN254"));
         // the 0-th element is reserved for capacity
         let mut state = vec![env, iv];
         for _ in 0..RATE {


soroban-sdk/src/crypto/poseidon2_sponge.rs
```diff
@@ -0,0 +1,100 @@
...
+    squeezed: bool,
+    cache: Vec<U256>,
+    state: Vec<U256>,
+    m_diag: Vec<U256>,
+    rc: Vec<Vec<U256>>,
+}
+
+impl Poseidon2Sponge {
+    pub fn new(env: &Env, iv: U256, field: Symbol) -> Poseidon2Sponge {
+        // TODO: we need to set up parameters for BLS12-381
```
leighmcculloch, line 19: TODO

soroban-sdk/src/crypto/poseidon2_sponge.rs
```diff
@@ -0,0 +1,100 @@
...
+    rate: u32,
+    squeezed: bool,
+    cache: Vec<U256>,
+    state: Vec<U256>,
+    m_diag: Vec<U256>,
+    rc: Vec<Vec<U256>>,
+}
+
+impl Poseidon2Sponge {
+    pub fn new(env: &Env, iv: U256, field: Symbol) -> Poseidon2Sponge {
```
leighmcculloch, line 18: Is `field` like an enum? Can we make it an enum? The enum can be converted into a symbol val internally when communicating to the host.

soroban-sdk/src/crypto/poseidon2_params.rs
```diff
@@ -0,0 +1,346 @@
...
+
+#[rustfmt::skip]
+pub fn get_rc4(e: &Env) -> Vec<Vec<U256>>
+{
+    vec![e,
+        vec![e, U256::from_be_bytes(e, &bytesn!(e, 0x19b849f69450b06848da1d39bd5e4a4302bb86744edc26238b0878e269ed23e5).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x265ddfe127dd51bd7239347b758f0a1320eb2cc7450acc1dad47f80c8dcf34d6).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x199750ec472f1809e0f66a545e1e51624108ac845015c2aa3dfc36bab497d8aa).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x157ff3fe65ac7208110f06a5f74302b14d743ea25067f0ffd032f787c7f1cdf8).into()),
+        ],
```
leighmcculloch, line 29: It's fine to skip formatting on these lines to organise them in a way that is more readable. But can we keep some indentation consistency?

For example:
```suggestion
        vec![e,
        U256::from_be_bytes(e, &bytesn!(e, 0x19b849f69450b06848da1d39bd5e4a4302bb86744edc26238b0878e269ed23e5).into()),
        U256::from_be_bytes(e, &bytesn!(e, 0x265ddfe127dd51bd7239347b758f0a1320eb2cc7450acc1dad47f80c8dcf34d6).into()),
        U256::from_be_bytes(e, &bytesn!(e, 0x199750ec472f1809e0f66a545e1e51624108ac845015c2aa3dfc36bab497d8aa).into()),
        U256::from_be_bytes(e, &bytesn!(e, 0x157ff3fe65ac7208110f06a5f74302b14d743ea25067f0ffd032f787c7f1cdf8).into()),
        ],
```

Or:

```suggestion
        vec![e,
            U256::from_be_bytes(e, &bytesn!(e, 0x19b849f69450b06848da1d39bd5e4a4302bb86744edc26238b0878e269ed23e5).into()),
            U256::from_be_bytes(e, &bytesn!(e, 0x265ddfe127dd51bd7239347b758f0a1320eb2cc7450acc1dad47f80c8dcf34d6).into()),
            U256::from_be_bytes(e, &bytesn!(e, 0x199750ec472f1809e0f66a545e1e51624108ac845015c2aa3dfc36bab497d8aa).into()),
            U256::from_be_bytes(e, &bytesn!(e, 0x157ff3fe65ac7208110f06a5f74302b14d743ea25067f0ffd032f787c7f1cdf8).into()),
        ],
```

soroban-sdk/src/crypto/poseidon2_params.rs
```diff
@@ -0,0 +1,346 @@
...
+        e,
+        U256::from_be_bytes(e, &bytesn!(e, 0x10dc6e9c006ea38b04b1e03b4bd9490c0d03f98929ca1d7fb56821fd19d3b6e7).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x0c28145b6a44df3e0149b3d0a30b3bb599df9756d4dd9b84a86b38cfb45a740b).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x00544b8338791518b2c7645a50392798b21f75bb60e3596170067d00141cac15).into()),
+        U256::from_be_bytes(e, &bytesn!(e, 0x222c01175718386f2e2e82eb122789e352e105a3b8fa852613bc534433ee428b).into()),
+    ]
+}
+
+#[rustfmt::skip]
+pub fn get_rc4(e: &Env) -> Vec<Vec<U256>>
```
leighmcculloch, line 22: My questions about this are likely to be naive. But wow there will be a lot of host function calls for this fn, that gets called once for each new `Poseidon2Sponge`. Is it likely or possible someone will need to create multiple `Poseidon2Sponge`? Maybe we should find a way to do `get_rc4` once only.

leighmcculloch, line 22: Did we consider introducing a host function that would provide a vec prefilled with this data?

soroban-sdk/src/crypto/poseidon2_sponge.rs
```diff
@@ -0,0 +1,100 @@
...
+        // TODO: we need to set up parameters for BLS12-381
+        assert_eq!(field, soroban_sdk_macros::internal_symbol_short!("BN254"));
+        let mut state = vec![env];
+        for _ in 0..RATE {
+            state.push_back(U256::from_u32(env, 0));
+        }
+        state.push_back(iv);
```
leighmcculloch, line 25: I think it would be more performant to construct the vec using `Vec::from_array` that does the construction of the vec with a single host fn.

soroban-sdk/src/crypto/poseidon2_sponge.rs
```diff
@@ -0,0 +1,100 @@
...
+            self.cache.push_back(U256::from_u32(&self.env, 0));
+        }
+        // add the cache into sponge state
+        for i in 0..RATE {
+            let elem = self
+                .state
+                .get_unchecked(i)
+                .add(&self.cache.get_unchecked(i));
+            self.state.set(i, elem);
+        }
```
leighmcculloch, line 55: Same comments here regarding vec construction requires less host fns if you use from_array.

soroban-sdk/src/crypto.rs
```diff
@@ -3,11 +3,19 @@
...
+    Bytes, BytesN, ConversionError, Env, IntoVal, Symbol, TryFromVal, TryIntoVal, Val, Vec, U256,
 };
 
 pub mod bls12_381;
 pub mod bn254;
+pub(crate) mod poseidon2_params;
+pub mod poseidon2_sponge;
+pub(crate) mod poseidon_params;
+pub mod poseidon_sponge;
+pub use bn254::Fr as BnScalar;
```
leighmcculloch, line 15: Is this type rename necessary? Any reason folks can't do `bn254::Fr`?

soroban-sdk/src/crypto.rs
```diff
@@ -3,11 +3,19 @@
...
     env::internal::{self, BytesObject},
     unwrap::UnwrapInfallible,
-    Bytes, BytesN, ConversionError, Env, IntoVal, TryFromVal, Val,
+    Bytes, BytesN, ConversionError, Env, IntoVal, Symbol, TryFromVal, TryIntoVal, Val, Vec, U256,
 };
 
 pub mod bls12_381;
 pub mod bn254;
+pub(crate) mod poseidon2_params;
+pub mod poseidon2_sponge;
```
leighmcculloch, line 12: Can we move the sponge and params inside a `poseidon2` module? Same with `poseidon`?

soroban-sdk/src/crypto.rs
```diff
@@ -189,6 +197,28 @@ impl Crypto {
...
+        let iv = U256::from_u32(&self.env, 0);
+        let mut sponge = PoseidonSponge::new(&self.env, iv, field);
+        for input in inputs.iter() {
+            sponge.absorb(input);
+        }
+        sponge.squeeze()
+    }
+
+    /// Performs a poseidon2 hash with a sponge construction equivalent to the one in the Barretenberg proving system
+    pub fn poseidon2_hash(&self, inputs: &Vec<U256>, field: Symbol) -> U256 {
```
leighmcculloch, line 213: Would it be appropriate to name these fns simple `poseidon` and `poseidon2`? Looking at the other functions in here we do not typically suffix hashing functions with `_hash`.

Or are there other things you can do with poseidon that would result in us needing to distinguish a hash from something else?

leighmcculloch (commented):
Some questions and feedback inline.

Note: I did not review the soundness of the crypto in any capacity, I'm not an expert or familiar with poseidon or poseidon2 hashing fns or primitives, and only the exported interface and general use of sdk types, etc.

jayz22:
Thanks @leighmcculloch , I will address your comments in a follow-up PR.

Issue #1448, opened 2025-03-01
Improve the in SDK fork testing experience


leighmcculloch:
If we are to make the SDKs fork testing a better experience, it needs a fast data source of a stable ledger. The way we have addressed the need for historical state today in the Soroban Rust SDK and stellar-cli's snapshot testing is for the stellar-cli to download a history archive, which takes considerable time and requires significant bandwidth.

This issue is to explore improving the fork testing experience by having the SDK act independently to retrieve ledger entries from data sources as it needs them. That data source is not yet defined, but ideally it would be an RPC, or some RPC-compatible service, or possibly but not preferred continued use of history archives.

The way the SDK fork testing works today is:

```mermaid
sequenceDiagram
    user->>+cli: create snapshot with<br/>accountIDs and contract IDs
    cli->>+historyarchive: req full ledger
    historyarchive->>-cli: full ledger (GBs, slow)
    cli->>cli: filter ledger down to<br/>accountIDs and contract IDs<br/>(painful dev experience)
    cli->>-user: snapshot (KBs)
    
    user->>+sdk: write test importing snapshot for fork testing
    sdk->>sdk: loads snapshot and uses it as storage starting point
    sdk->>-user: 
```

The way it would ideally work if the SDK could independently load a stable ledger state from a data-source. RPC is listed as the data source, but imagine it could be any data-source that provides random access of a stable ledger state, via an RPC-compatible API:
```mermaid
sequenceDiagram
    user->>+sdk: write test fork testing ledger N
    loop 
    sdk->>sdk: test running
    alt local snapshot exists (cache)
    sdk->>sdk: load ledger entry for key K from snapshot
    else local snapshot does not exist
    sdk->>+rpc: req ledger entry for key K ledger N
    rpc->>-sdk: entry for K (Bytes)
    end
    end
    opt if new entries were downloaded
    sdk->>sdk: write local snapshot for starting point
    end
    sdk->>-user: 
```

This idea was first discussed here:
- https://github.com/stellar/quickstart/issues/625#issuecomment-2686362940

Dependent on:
- https://github.com/stellar/stellar-rpc/issues/375 - This issue was opened to explore the possibility that RPC, or an RPC-compatible service, could serve stable recent historical state in a fashion that could be randomly accessed.
- https://github.com/stellar/stellar-core/issues/4397
   - https://github.com/stellar/stellar-core/pull/4623

leighmcculloch:
Next steps:
[...]

PR #1619, opened 2025-11-14
Extract default ledger info to testutils function

leighmcculloch:
### What
  Extract hardcoded LedgerInfo initialization in Env::default() into a new testutils::default_ledger_info() function.

  ### Why
  I'm writing more code that needs to set a default ledger info and it's helpful if this is outside of the Env code and I can call it in the new places too.

For #1448

### DIFF
--- a/soroban-sdk/src/testutils.rs
+++ b/soroban-sdk/src/testutils.rs
@@ -256,7 +256,7 @@ pub trait ContractFunctionSet {
 pub use crate::env::internal::LedgerInfo;
 
 /// Returns a default `LedgerInfo` suitable for testing.
-pub fn default_ledger_info() -> LedgerInfo {
+pub(crate) fn default_ledger_info() -> LedgerInfo {
     LedgerInfo {
         protocol_version: 23,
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

soroban-sdk/src/testutils.rs
```diff
@@ -255,6 +255,20 @@ pub trait ContractFunctionSet {
 #[doc(inline)]
 pub use crate::env::internal::LedgerInfo;
 
+/// Returns a default `LedgerInfo` suitable for testing.
+pub(crate) fn default_ledger_info() -> LedgerInfo {
+    LedgerInfo {
+        protocol_version: 23,
```
mootz12, line 261: p24?

mootz12 (approved):

soroban-sdk/src/testutils.rs
leighmcculloch: That should get updated in the next major release. The sdk is still supports a min of p23.

PR #1620, opened 2025-11-14
Accept custom snapshot sources in Env::from_ledger_snapshot

leighmcculloch:
### What
  Accept custom snapshot source input types as ledger snapshots through a new SnapshotSourceInput struct that wraps source, ledger_info, and snapshot parameters. Implement From<LedgerSnapshot> for SnapshotSourceInput to maintain backward compatibility. Cache anything loaded from the snapshot source and write it out for reuse.

  ### Why
  Enables Env to be initialized from different snapshot source types beyond just the LedgerSnapshot file, providing more flexibility for testing scenarios while keeping the existing API intact. The snapshot source is only used to load the data once, then it is loaded from the cache file for test reproducibility and speed.

This will support hooking in more snapshot sources such as RPCs, history archives, and the meta lake which I am working on.

For #1448 

### Merging

Targeting `main`, but dependent on:
- https://github.com/stellar/rs-soroban-sdk/pull/1623

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -334,10 +334,7 @@ impl Env {
     /// incorrectly can introduce security vulnerabilities. Use [Crypto] if
     /// possible.
     #[cfg_attr(any(test, feature = "hazmat-crypto"), visibility::make(pub))]
-    #[cfg_attr(
-        feature = "docs",
-        doc(cfg(any(feature = "hazmat", feature = "hazmat-crypto")))
-    )]
+    #[cfg_attr(feature = "docs", doc(cfg(feature = "hazmat-crypto")))]
     #[inline(always)]
     pub(crate) fn crypto_hazmat(&self) -> crate::crypto::CryptoHazmat {
         crate::crypto::CryptoHazmat::new(self)
--- a/.github/workflows/test-with-soroban-examples.yml
+++ b/.github/workflows/test-with-soroban-examples.yml
@@ -75,7 +75,14 @@ jobs:
       run: |
         # TODO: Update this patch logic to use `cargo add` once this issue is resolved: https://github.com/rust-lang/cargo/issues/16101
         crates=$(cd ${{ github.workspace }}/rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
-        find . -name Cargo.toml | while read file; do
+
+        # Find Cargo.toml files to patch
+        local_files=$(find . -name Cargo.toml)
+        workspace_root=$(cargo metadata --format-version 1 --no-deps | jq -r '.workspace_root')
+        files=$(echo "$local_files"; echo "$workspace_root/Cargo.toml")
+
+        # Patch local files and workspace root (sort -u to deduplicate)
+        echo "$files" | sort -u | while read -r file; do
           echo Patching "$file" ...
           dir=$(dirname "$file")
           for crate in $crates; do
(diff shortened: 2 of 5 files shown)


soroban-sdk/src/env.rs
```diff
@@ -577,6 +578,40 @@ impl Env {
...
+        let snapshot = snapshot.or_else(|| {
+            // Try to read from test_snapshots_before file for the current test.
+            if let Some(test_name) = test_name.as_ref() {
+                // Construct path similar to to_test_ledger_snapshot_before_file.
+                let test_name_path = test_name
+                    .split("::")
+                    .map(|p| std::path::Path::new(p).to_path_buf())
+                    .reduce(|p0, p1| p0.join(p1))
+                    .expect("test name to not be empty");
+                let dir = std::path::Path::new("test_snapshots_before");
```
mootz12: [picky] not entirely sure I like the naming of this, operating on the assumption these cache files are almost always checked in to source control to ensure reproducibility.

Ideally, we could keep all snapshots in the same folder

```
test_snapshots/
    source/
        test_hello.1.json
    results/
        test_hello.1.json
```

I think ^ would be best, but realize this is a breaking change for existing uses of `test_snapshots` for validation.

Two potential alternatives that wouldn't cause breaking changes:

```
test_snapshots/
    test_hello.1.source.json
    test_hello.1.json
    ...
```
or simply
```
test_snapshots_source/
    test_hello.1.json
```

mootz12 (commented):
Small comment on naming, looks good otherwise

soroban-sdk/src/env.rs
leighmcculloch: Naming the new files was something I was struggling with. I like the `source` vocab, that's good, it also directly aligns well with `SnapshotSource`, which is better than introducing the new term `before`. I'll rename it to `test_snapshots_source/`. I like the separate directory option because it means one can git ignore either or do something on bulk on the different kinds of files.

leighmcculloch:
@mootz12 I've simplified how this works so that it doesn't write out any new files and this is just an interface change to allow providing custom implementations for snapshots. Any file writing will be the job of the snapshot implementation itself if it wants to do caching on its own terms.

soroban-sdk/src/testutils.rs
```diff
@@ -605,3 +620,75 @@ impl StellarAssetContract {
...
+///
+/// This struct enables [`Env::from_ledger_snapshot`] to accept custom snapshot
+/// source types beyond [`LedgerSnapshot`], providing flexibility for testing
+/// scenarios that load ledger state from different sources such as RPC endpoints,
+/// history archives, or in-memory data structures.
+///
+/// # Fields
+///
+/// * `source` - A snapshot source implementing the [`SnapshotSource`] trait.
+///   This is used to load ledger entries on demand during test execution.
```
mootz12, line 634: So, if we assume `source` pulls from an RPC, and the test being executed has multiple steps, and performs a token transfer first, does `source.get` need to do anything to fetch the updated token balance? Or is that managed by the test env itself, and `source.get` is only used for initialization of ledger entries?

leighmcculloch: That's a great question, and this wasn't obvious to me when I started working with snapshot source either. The snapshot source only loads state when it's first needed, but once state is loaded, the rest of the transaction is updating state in memory, and then at the very end of the transaction, everything is written out.

So, in tests, snapshot source collects data the very first time it's needed, it's then cached, and any subsequent requests reading or mutating that ledger entry don't re-query the snapshot source.

This is also why the snapshot source doesn't have a set function, only a get function. It's only for that initial loading of the data into the runtime.

PR #1621, opened 2025-11-14
Add RPC-based snapshot source for fast but low consistency fork tests

leighmcculloch:
### What
  Add RPC-based snapshot source implementation that fetches ledger entries from stellar-rpc server. Implement SnapshotSource trait using synchronous HTTP requests to getLedgerEntries. Add test_fork example demonstrating RPC snapshot usage.

### Why
Enables contract testing against live or local RPC endpoints by forking network state on a rolling basis. This allows developers to test contract interactions with actual on-chain data without deploying to test networks, improving test realism and debugging capabilities. This implementation of the snapshot source leans into getting data fast, but with no guarantees to consistency. The ledger entries loaded may span more than a single ledger.

For #1448 

### Merging

Targeting `main`, but dependent on:
- https://github.com/stellar/rs-soroban-sdk/pull/1620
- https://github.com/stellar/rs-soroban-sdk/pull/1623

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -578,9 +578,9 @@ impl Env {
             1
         };
 
-        // Apply fallback to read from test_snapshots_before if snapshot is None
+        // Apply fallback to read from test_snapshots_source if snapshot is None
         let snapshot = snapshot.or_else(|| {
-            // Try to read from test_snapshots_before file for the current test.
+            // Try to read from test_snapshots_source file for the current test.
             if let Some(test_name) = test_name.as_ref() {
                 // Construct path similar to to_test_ledger_snapshot_before_file.
                 let test_name_path = test_name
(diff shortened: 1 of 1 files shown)


soroban-snapshot-source-rpc/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+    }
+}
+
+impl From<RpcSnapshotSource> for SnapshotSourceInput {
+    fn from(source: RpcSnapshotSource) -> Self {
+        Self {
+            source: Rc::new(source),
+            ledger_info: None,
```
mootz12, line 113: this should pull from the cache file, or fetch `latestLedger` to apply timestamp and sequence number from the RPC

Makes it quite hard to test any contracts that rely on things like ledger sequence or timestamp otherwise

soroban-snapshot-source-rpc/src/lib.rs
mootz12, line 1: Curious why you opted to make this a separate crate? I figured this would all live under the existing `soroban-ledger-snapshot` crate, given the overlap.

leighmcculloch: Just because this is all hacking about right now, trying to find patterns and experiences. Nothing about the organisation of crates will stick.

soroban-snapshot-source-rpc/src/lib.rs
leighmcculloch: Agreed. I had a commit that did this but dropped it because trying to figure this out is tricky with the current organisation of code. Right now ledger info is needed upfront, which means RpcSnapshotSource always makes a call. But ideally RpcSnapshotSource only makes a call if the file isn't loaded. I think I need to rearrange the SnapshotSourceInput so the components are lazily gettable.

PR #1623, opened 2025-11-17
Store test name and env number in env itself

leighmcculloch:
### What
  Store test name and environment number directly in EnvTestState instead of tracking them at the point where test snapshots are written.

  ### Why
  Snapshot file numbering is confusing and difficult to follow when multiple Env instances existed per test. It's because the order of the env numbers is based on the order of the drops, not the order of the envs being defined. This is difficult to reason about, and doesn't work with new features I want to add where we'll write and read from other types of files for each Env in a test.

This is technically a breaking change, but multiple env's in tests is super rare and is supported as an edge case to ensure things still keep working. I think it's okay to "fix this" with a break given the rareness of its use. Anyone using it who does see a break will see there test files flip. 1 -> 2, 2 -> 1.

This change is a bit of a clean up, but it is also required for the work on improving fork testing (#1448). The improved fork testing will start sourcing network data from a variety of sources but will need to cache that data for subsequent test runs to ensure tests stay fast and deterministic. The logic used to assign test names and numbers to envs will be used to separate the caches for envs in tests. Even if we don't end up using this in the fork testing flows I think this change stands alone and is worth it for the other reasons above.

For #1448

### DIFF
--- a/soroban-sdk/src/tests/env.rs
+++ b/soroban-sdk/src/tests/env.rs
@@ -150,7 +150,7 @@ fn test_snapshot_file() {
         assert!(!p1.exists());
         assert!(!p2.exists());
         assert!(p3.exists());
-    } // Env dropped, written to p2.
+    } // Env dropped, written to p1.
     assert!(p1.exists());
     assert!(!p2.exists());
     assert!(p3.exists());


soroban-sdk/src/tests/env.rs
leighmcculloch, line 111: There are two changes to this test that do a good job of illustrating how the behaviour is changing.

1️⃣ There are now three paths that get checked. That's because this test has three Envs. Every Env is assigned a number now, instead of only those that wrote a test snapshot. Prior to this change Env 2 never writes a snapshot and so it is never assigned a number. Now every Env has a number.

2️⃣ The assignment of the numbers to the Envs is different. Before they were assigned based on the order that they are dropped. The third Env is dropped first and so was assigned the p1 file. Drops are implied in Rust code, so there's no keyword or command that indicates its happening, you just have to remember and know when it happens. After this change the number assignment happens when the Envs are defined, which is explicit and more predictable.

soroban-sdk/src/tests/env.rs
```diff
@@ -115,38 +115,48 @@ fn test_snapshot_file() {
...
         {
             let _ = Env::default(); // When dropped won't be written because empty.
         } // Env dropped, nothing written.
         assert!(!p1.exists());
         assert!(!p2.exists());
+        assert!(!p3.exists());
         {
             let e3 = Env::default(); // When dropped will be written to p1.
             let _ = e3.register(Contract, ());
         } // Env dropped, written to p1.
```
mootz12: [picky] written to p3*

soroban-sdk/src/tests/env.rs
```diff
@@ -115,38 +115,48 @@ fn test_snapshot_file() {
...
-        assert!(p1.exists());
+        assert!(!p1.exists());
         assert!(!p2.exists());
+        assert!(p3.exists());
         e2.as_contract(&c, || {});
-        assert!(p1.exists());
+        assert!(!p1.exists());
         assert!(!p2.exists());
+        assert!(p3.exists());
     } // Env dropped, written to p2.
```
mootz12: [picky] written to p1*

soroban-sdk/src/tests/env.rs
leighmcculloch:
```suggestion
            let e3 = Env::default(); // When dropped will be written to p1.
            let _ = e3.register(Contract, ());
        } // Env dropped, written to p3.
```

soroban-sdk/src/tests/env.rs
leighmcculloch:
```suggestion
    } // Env dropped, written to p1.
```

PR #1633, opened 2025-12-02
Issues/1566 simplify event testing

mootz12:
### What

Updates `testutils::Events` read functions to use the `xdr::ContractEvent` type and deprecates the current `all` function. Adds helpers to `contractevent` structs to easily transform event structs into `xdr::ContractEvent`'s for easy comparison during tests.

Finally, this also adds function to `testutils::Events` to allow test writers to easily pull events published by a specific `contract_id`.

As a side effect, using xdr also allows the assertion debug info to display what is different between events.

For example:
```
left: [ContractEvent { ext: V0, contract_id: Some(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000001))), type_: Contract, body: V0(ContractEventV0 { topics: VecM([Symbol(ScSymbol(StringM(transfer))), Address(Contract(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000002)))), Address(Account(AccountId(PublicKeyTypeEd25519(Uint256(0000000000000000000000000000000000000000000000000000000000000003)))))]), data: Map(Some(ScMap(VecM([ScMapEntry { key: Symbol(ScSymbol(StringM(amount))), val: I128(Int128Parts { hi: 0, lo: 1 }) }, ScMapEntry { key: Symbol(ScSymbol(StringM(to_muxed_id))), val: U64(1) }])))) }) }]
 right: [ContractEvent { ext: V0, contract_id: Some(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000001))), type_: Contract, body: V0(ContractEventV0 { topics: VecM([Symbol(ScSymbol(StringM(transfer))), Address(Contract(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000002)))), Address(Account(AccountId(PublicKeyTypeEd25519(Uint256(0000000000000000000000000000000000000000000000000000000000000003)))))]), data: Map(Some(ScMap(VecM([ScMapEntry { key: Symbol(ScSymbol(StringM(amount))), val: I128(Int128Parts { hi: 0, lo: 0 }) }, ScMapEntry { key: Symbol(ScSymbol(StringM(to_muxed_id))), val: U64(1) }])))) }) }]
 ```
 instead of
 ```
 left: Vec(Ok((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Vec(Ok(Symbol(transfer)), Ok(Address(obj#35)), Ok(Address(obj#37))), Map(obj#43))))
 right: Vec(Ok((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Vec(Ok(Symbol(transfer)), Ok(Address(obj#11)), Ok(Address(obj#29))), Map(obj#53))))
 ```

### Why

Helps test write easier event comparison.

For a lighter solution, we can also consider #1634 . This just adds a few utils to make building events in tests easier.

### Known limitations

None

### DIFF
--- a/soroban-sdk/src/address.rs
+++ b/soroban-sdk/src/address.rs
@@ -388,9 +388,9 @@ impl crate::testutils::Address for Address {
     }
 }
 
-#[cfg(not(target_family = "wasm"))]
+#[cfg(any(not(target_family = "wasm"), test, feature = "testutils"))]
 impl Address {
-    pub(crate) fn contract_id(&self) -> ContractId {
+    pub fn contract_id(&self) -> ContractId {
         let sc_address: ScAddress = self.try_into().unwrap();
         if let ScAddress::Contract(c) = sc_address {
             c
@@ -399,7 +399,7 @@ impl Address {
         }
     }
 
-    pub(crate) fn from_contract_id(env: &Env, contract_id: [u8; 32]) -> Self {
+    pub fn from_contract_id(env: &Env, contract_id: [u8; 32]) -> Self {
         Self::try_from_val(env, &ScAddress::Contract(ContractId(Hash(contract_id)))).unwrap()
     }
 }
--- a/soroban-sdk/src/events.rs
+++ b/soroban-sdk/src/events.rs
@@ -59,7 +59,7 @@ impl Debug for Events {
 }
 
 #[cfg(any(test, feature = "testutils"))]
-use crate::{testutils, xdr, Address, TryIntoVal};
+use crate::{testutils, xdr, Address, TryFromVal, TryIntoVal};
 
 pub trait Event {
     fn topics(&self, env: &Env) -> Vec<Val>;
@@ -69,14 +69,18 @@ pub trait Event {
         env.events().publish_event(self);
     }
 
-    /// Convert this event into a [`testutils::ContractEvent`] object for testing purposes.
+    /// Convert this event and the given contract_id into a [`xdr::ContractEvent`] object.
     #[cfg(any(test, feature = "testutils"))]
-    fn to_contract_event(
-        &self,
-        env: &Env,
-        contract_id: &crate::Address,
-    ) -> testutils::ContractEvent {
-        testutils::ContractEvent::new(env, contract_id.clone(), self.topics(env), self.data(env))
+    fn to_contract_event(&self, env: &Env, contract_id: &crate::Address) -> xdr::ContractEvent {
+        xdr::ContractEvent {
+            ext: xdr::ExtensionPoint::V0,
+            type_: xdr::ContractEventType::Contract,
+            contract_id: Some(contract_id.contract_id()),
+            body: xdr::ContractEventBody::V0(xdr::ContractEventV0 {
+                topics: self.topics(env).into(),
+                data: xdr::ScVal::try_from_val(env, &self.data(env)).unwrap(),
+            }),
+        }
     }
 }
 
@@ -162,8 +166,7 @@ impl testutils::Events for Events {
         vec
     }
 
-    fn contract_events(&self) -> std::vec::Vec<testutils::ContractEvent> {
-        let env = self.env();
+    fn contract_events(&self) -> std::vec::Vec<xdr::ContractEvent> {
         self.env()
             .host()
             .get_events()
@@ -171,21 +174,11 @@ impl testutils::Events for Events {
             .0
             .into_iter()
             .filter_map(|e| {
-                if e.failed_call {
-                    None
-                } else if let xdr::ContractEvent {
-                    type_: xdr::ContractEventType::Contract,
-                    contract_id: Some(contract_id),
-                    body: xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, data }),
-                    ..
-                } = e.event
+                if !e.failed_call
+                    && e.event.type_ == xdr::ContractEventType::Contract
+                    && e.event.contract_id.is_some()
                 {
-                    Some(testutils::ContractEvent::new(
-                        &env,
-                        Address::from_contract_id(env, contract_id.0 .0),
-                        topics.try_into_val(env).unwrap(),
-                        data.try_into_val(env).unwrap(),
-                    ))
+                    Some(e.event)
                 } else {
                     None
                 }
@@ -196,10 +189,11 @@ impl testutils::Events for Events {
     fn contract_events_for(
         &self,
         contract_id: &crate::Address,
-    ) -> std::vec::Vec<testutils::ContractEvent> {
+    ) -> std::vec::Vec<xdr::ContractEvent> {
+        let contract_id = Some(contract_id.contract_id());
         self.contract_events()
             .into_iter()
-            .filter(|e| &e.contract_id == contract_id)
+            .filter(|e| e.contract_id == contract_id)
             .collect()
     }
 }
--- a/soroban-sdk/src/testutils.rs
+++ b/soroban-sdk/src/testutils.rs
@@ -484,18 +484,21 @@ pub trait Events {
     /// Events are returned in the order they were published, with the
     /// last published event being the last in the list.
     ///
-    /// Returns a [`std::vec::Vec`] of [`ContractEvent`]s
-    fn contract_events(&self) -> std::vec::Vec<ContractEvent>;
+    /// Returns a [`std::vec::Vec`] of [`xdr::ContractEvent`]s
+    fn contract_events(&self) -> std::vec::Vec<xdr::ContractEvent>;
 
-    /// Returns all contract events that have been published by the given contract id
+    /// Returns all contract events that have been published by the given contract
     /// during the last contract invocation. If the last contract invocation failed,
     /// no events are returned.
     ///
     /// Events are returned in the order they were published, with the
     /// last published event being the last in the list.
     ///
-    /// Returns a [`std::vec::Vec`] of [`ContractEvent`]s
-    fn contract_events_for(&self, contract_id: &crate::Address) -> std::vec::Vec<ContractEvent>;
+    /// Returns a [`std::vec::Vec`] of [`xdr::ContractEvent`]s
+    fn contract_events_for(
+        &self,
+        contract_id: &crate::Address,
+    ) -> std::vec::Vec<xdr::ContractEvent>;
 }
 
 /// Test utilities for [`Logs`][crate::logs::Logs].
--- a/soroban-token-sdk/src/tests/events.rs
+++ b/soroban-token-sdk/src/tests/events.rs
@@ -8,15 +8,16 @@ use crate::events::{
 };
 use soroban_sdk::{
     contract, symbol_short,
-    testutils::{Address as _, ContractEvent, Events as _, MuxedAddress as _},
+    testutils::{Address as _, Events as _, MuxedAddress as _},
     token::StellarAssetClient,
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: soroban-sdk/src/testutils.rs [hunk 1]
```diff
@@ -413,15 +415,60 @@ pub mod budget {
     }
 }
 
+#[derive(Clone)]
+pub struct ContractEvent {
+    pub env: Env,
```
mootz12: Curious what the best practice is for using `env` here would be.

Decided to store a copy on the struct, but I could also use the env from one of the fields as well

tupui:
Great to see some work on that!!

I personally would prefer your alternate option 1. Would there be a way to just not care about the contract ID? I would suspect that in lot of cases people would not care much about asserting that no? In basic instances, I actually might just want to check that I have an instance of a specific event and don't need to check the inner values.

tupui:
> > Would there be a way to just not care about the contract ID?
> 
> You would need to compare the parts of the event you care about manually within the test.
> ...
> 
> I don't think it would be a good idea for contract IDs to be omitted from the `env.events()` in general.
> 

Fair enough, your example is a simple enough syntax 👍 

> One advantage of the `ContractEvent` struct, is that you can do things like:
> 
> ```rust
>         assert!(env.events().all().contains(
>             &Transfer {
>                 from: from.clone(),
>                 to: to.clone(),
>                 amount,
>                 to_muxed_id: None,
>             }
>             .to_contract_event(&env, &contract_id)
>         ));
> ```
> 
> This is neat for composed contracts where their event might be one of many in the test transaction.

This would be great to be able to do some contains like that 💯 

One question, is there actually an advantage in having `env.events().all()` not return directly a vec of struct? Or maybe we could have a `env.events_test().all()` which would do the handy manipulation for us.

soroban-sdk/src/testutils.rs
leighmcculloch: We typically store the env explicitly, even if there's also the env available via another field. Because this is an internal type, I would be inclined to make all of the fields non-public so that people aren't reaching in here.

I think there is somewhere else in the SDK where we made that mistake and it became a little bit painful to ever change that type in the future. Any change, including adding a field, would be a breaking change.

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +415,60 @@ pub mod budget {
...
+
+impl PartialEq for ContractEvent {
+    fn eq(&self, other: &Self) -> bool {
+        if self.contract_id != other.contract_id || self.topics != other.topics {
+            false
+        } else if self.data.is_object() || other.data.is_object() {
+            self.env.obj_cmp(self.data, other.data).unwrap_infallible() == 0
+        } else {
+            self.data.get_payload() == other.data.get_payload()
+        }
```
leighmcculloch, line 436: At some point in the past, @graydon and I discussed the idea of having a type that contained an env and a val, and we could use that in places where we wanted it to be comparable via the obj_cmp function. This could be another good use case for where that would be usable.

The data and the topics would be Vec of that type that wraps the env and the val instead of just a val. So then this entire type would just be comparable as well automatically without having to write these functions.

I wonder if that would simplify anything about this solution.

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +415,60 @@ pub mod budget {
...
+
+impl Debug for ContractEvent {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(
+            f,
+            "ContractEvent {{ contract_id: {:?}, topics: {:?}, data: {:?} }}",
+            self.contract_id, self.topics, self.data
+        )
+    }
+}
```
leighmcculloch, line 448: This implementation looks like it matches pretty closely to what the default implementation would be if we just derived debug on the type, which I think would be reasonable. Is there a reason that we wouldn't use that?

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +415,60 @@ pub mod budget {
...
 pub trait Events {
     /// Returns all events that have been published by contracts.
     ///
-    /// Returns a [`Vec`] of three element tuples containing:
-    /// - Contract ID
-    /// - Event Topics as a [`Vec<Val>`]
-    /// - Event Data as a [`Val`]
-    fn all(&self) -> Vec<(crate::Address, Vec<Val>, Val)>;
+    /// Returns a [`std::vec::Vec`] of [`ContractEvent`]s
+    fn all(&self) -> std::vec::Vec<ContractEvent>;
```
leighmcculloch: Because this is a breaking change, I'm thinking: how could we do this so that it's not a breaking change? And/or how could we do this so that at least when people are using it to assert against some tuples, they don't have to change anything or aren't met with a whole bunch of errors and then have to figure out how to rewrite it all away from the tuples?

soroban-token-sdk/src/tests/events.rs
```diff
@@ -42,10 +42,12 @@ fn test_approve() {
     let token_events = env.events().all();
     assert_eq!(
         token_events,
-        vec![
+        std::vec![ContractEvent::new(
```
leighmcculloch: If we made it possible to compare a Soroban vec with a listed vec, would the need for this change disappear? I'm thinking about whether that change worked, and if it would soften the breaking nature of this change.

tupui:
> assert!(event.matches(&env, &id, &published));

This is nice and simpler, I like it!

I am not sure about the wording `matches`, as at first I was thinking about a Rust `match`. It's more an inclusion so `contains` above was more obvious on what was going on. Maybe `event.published`, `event.published_in` or something like that would be clearer and directly link to `event.publish`?

soroban-sdk/src/testutils.rs
mootz12: Yup, that would simplify things considerably.

It's likely we wouldn't even need the custom type, as the existing tuple `(Address, Vec<Val>, Val)` could be directly compared, given all it's components now would implement partial eq, and it's easy enough to create this tuple from a `contractevent`.

I thought about this but opted against it as I figured there is a reason it wasn't done, but I think this would be the simplest breaking change we could introduce to fix this.

soroban-sdk/src/testutils.rs
mootz12: The `env` field doesn't implement debug

soroban-sdk/src/testutils.rs
mootz12: This was the thought behind #1634. It gets most of the way there in terms of testing semantics, but IMO does it in a less obvious way to prevent old tests from breaking.

soroban-token-sdk/src/tests/events.rs
mootz12: It wanted me to make `ContractEvent` implement TryFromVal, which I thought was not necessary.

This also was inline with how other test-only vecs are returned, like `Logs`.

soroban-sdk/src/address.rs
```diff
@@ -388,9 +388,9 @@ impl crate::testutils::Address for Address {
     }
 }
 
-#[cfg(not(target_family = "wasm"))]
+#[cfg(any(not(target_family = "wasm"), test, feature = "testutils"))]
```
leighmcculloch: Why did this line need changing? I don't remember the full backstory but the way fns are behind cfg is split, there's stuff behind it for testutils, and all the scval stuff is behind the non-wasm target.

soroban-sdk/src/events.rs
```diff
@@ -58,13 +58,30 @@ impl Debug for Events {
...
     fn topics(&self, env: &Env) -> Vec<Val>;
     fn data(&self, env: &Env) -> Val;
 
     fn publish(&self, env: &Env) {
         env.events().publish_event(self);
     }
+
+    /// Convert this event and the given contract_id into a [`xdr::ContractEvent`] object.
+    #[cfg(any(test, feature = "testutils"))]
+    fn to_contract_event(&self, env: &Env, contract_id: &crate::Address) -> xdr::ContractEvent {
```
leighmcculloch:
```suggestion
    /// Convert this event into a [`xdr::ContractEvent`] that is published by the given contact id.
    #[cfg(any(test, feature = "testutils"))]
    fn to_contract_event(&self, env: &Env, contract_id: &crate::Address) -> xdr::ContractEvent {
```

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +415,90 @@ pub mod budget {
...
+        write!(
+            f,
+            "ContractEvent {{ contract_id: {:?}, topics: {:?}, data: {:?} }}",
+            self.contract_id, self.topics, self.data
+        )
+    }
+}
+
+impl ContractEvent {
+    pub fn new(
```
leighmcculloch: Is this internally created only? In the interest of keeping the public interface exposed tighter and just expose things as we need, can it be pub crate or doc hidden?

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +415,90 @@ pub mod budget {
...
+}
+
 /// Test utilities for [`Events`][crate::events::Events].
 pub trait Events {
-    /// Returns all events that have been published by contracts.
+    /// Returns all contract events that have been published by the last contract
+    /// invocation. If the last contract invocation failed, no events are returned.
+    ///
+    /// Events are returned in the order they were published, with the
+    /// last published event being the last in the list.
```
leighmcculloch: 👏🏻

stellar-asset-spec/src/tests/events.rs
```diff
@@ -10,6 +10,7 @@ use soroban_sdk::{
 struct Contract;
 
 #[test]
+#[allow(deprecated)]
```
leighmcculloch: Before merging once we're settled on the impl can we update all the tests so that they call both `all` and the new fn while both are being supported?

tests/events/src/lib.rs
```diff
@@ -62,28 +63,14 @@ mod test {
...
-                        (
-                            Symbol::new(&env, "to_muxed_id"),
-                            <_ as IntoVal<Env, Val>>::into_val(&to.id().unwrap(), &env)
-                        ),
-                    ]
-                    .to_val()
-                ),
-            ],
+            env.events().contract_events(),
+            std::vec![Transfer {
```
leighmcculloch: You shouldn't need a std::vec here, you should be able to use an array. It should mean you don't need the extern crate std either.
```suggestion
            [Transfer {
```
If that doesn't work, you should be able to use a slice:
```suggestion
            &[Transfer {
```

tests/events/src/lib.rs
```diff
@@ -100,25 +87,14 @@ mod test {
         client.transfer(&from, &to, &amount);
 
         assert_eq!(
-            env.events().all(),
```
leighmcculloch: Can we preserve the use of `all` in this test while both fns exist in the sdk? So keep the existing assertion, and then add the new assertion behind it.

soroban-sdk/src/events.rs
leighmcculloch: I think we should drop the `env` and get the `env` out of the `Address`.

```suggestion
    fn to_contract_event(&self, contract_id: &crate::Address) -> xdr::ContractEvent {
```

mootz12:
closing in favor of #1638

Issue #1566, opened 2025-09-13
Change env.events().all() to return list comparable with contractevent types for easier definition of assert values


leighmcculloch:
I propose we change `env.events().all()` to return list that is comparable with `contractevent` struct types.

Today defining events to assert on is cumbersome. So cumbersome that it takes an AI several attempts to get close and I haven't seen success yet.

Doing so would make it easier to define the events being asserted on, for humans an machines.

How to make this change needs some exploration because the SDK can't know what the type is. Maybe we can use generic types to specify the event type. Maybe we can just support the PartialEq trait between the existing values and the event type. 

Alternative: Make the contractevent structs comparable with the tuple returned by the all fn.

japui-coder:
Change env.events().all() to return list comparable with contractevent types for easier definition of assert values #1634

PR #1634, opened 2025-12-03
Issues/1566 simplify event testing 2

mootz12:
### What

Adds test functions to the `Event` trait to help construct event tuples and to `testutils::Events` to pull only events for a given contract_id.

Adds a function `

### Why

This serves as an alternate, minimal solution to #1566, as opposed to #1633.

### Known limitations

None

### DIFF
--- a/tests/events_ref/src/lib.rs
+++ b/tests/events_ref/src/lib.rs
@@ -60,15 +60,18 @@ mod test {
 
         client.transfer(&from, &to, &amount);
 
-        let event = Transfer {
-            from: &from,
-            to: &to.address(),
-            amount: &amount,
-            to_muxed_id: Some(&to.id().unwrap()),
-        };
         assert_eq!(
             env.events().from_contract(&contract_id),
-            vec![&env, event.to_event_tuple(&env, &contract_id)]
+            vec![
+                &env,
+                Transfer {
+                    from: &from,
+                    to: &to.address(),
+                    amount: &amount,
+                    to_muxed_id: Some(&to.id().unwrap()),
+                }
+                .to_event_tuple(&env, &contract_id)
+            ]
         );
     }
 
@@ -84,15 +87,18 @@ mod test {
 
         client.transfer(&from, &to, &amount);
 
-        let event = Transfer {
-            from: &from,
-            to: &to,
-            amount: &amount,
-            to_muxed_id: None,
-        };
         assert_eq!(
             env.events().from_contract(&contract_id),
-            vec![&env, event.to_event_tuple(&env, &contract_id)]
+            vec![
+                &env,
+                Transfer {
+                    from: &from,
+                    to: &to,
+                    amount: &amount,
+                    to_muxed_id: None,
+                }
+                .to_event_tuple(&env, &contract_id)
+            ]
         );
     }
 


tests/events/src/lib.rs
```diff
@@ -61,30 +60,13 @@ mod test {
...
-                ),
-            ],
-        );
+        let event = Transfer {
+            from: from.clone(),
+            to: to.address(),
+            amount,
+            to_muxed_id: to.id(),
+        };
+        assert!(env.events().contains(&contract_id, &event));
```
dmkozh: I would generally recommend against using non-exhaustive assertions where possible. Using `contains` both misses the unintentionally published events, and also misses the event order for the cases when there is more than one event present. I can see this being _sometimes_ a more concise option for testing complex multi-contract interactions, but generally I think we still need better tools for validating the whole event array (i.e. another PR in some shape), and promote that as the 'best practice' for the event testing.

tupui: Yes but then we really need better tooling for that. For instance if you do some calls to transfer within your contract, you will get the events from these transfers. You don't necessarily care about these and since you don't have the event trait itself and are not the one raising the event. You need to hunt it down or play around until you get the structure right in your tests. It's proven to be quite tedious to test on my side.

dmkozh: I think a contract_id filter makes sense for when you only care about 1 contract, would that make sense for your case?

BTW, in case of transfers, the spec actually exists in the SDK. I understand that not every test needs this, but OTOH I can see how at least some tests would want this. Transfers are actually pretty important side effects, and it would make sense for at least a fraction of tests to make sure that the expected amounts have actually been transferred. So even if there is a way to narrow down the event assertion scope, I would still argue that we should also have a simple way to cover the common token events.

tupui: > Transfers are actually pretty important side effects, and it would make sense for at least a fraction of tests to make sure that the expected amounts have actually been transferred. 

Totally agree, but I prefer to check balances vs rely on events, as this is the actual thing that matters.

tupui: Regardless, if this is the way, we should then have easier way to pull in events which are not ours. Transfer as you said is ok to find, but for cross contracts I then have to go on GitHub or else to try to find the specs.

soroban-sdk/src/tests/contract_event.rs
```diff
@@ -542,3 +543,258 @@ fn test_ref_fields() {
...
+        name: symbol_short!("hi"),
+        value: symbol_short!("hello"),
+    };
+    env.as_contract(&id, || {
+        event.publish(&env);
+    });
+
+    let published = env.events().all().get_unchecked(0);
+    assert!(!event.matches(&env, &id_2, &published));
+    assert!(event.matches(&env, &id, &published));
```
leighmcculloch: One thing to consider is that these style of assertions are difficult to inspect when things aren't working in comparison to using `assert_eq!` where the two values will be displayed in test output.

tupui: Could we do some test first on the trait type maybe? This way we could say if the event trait is event present in the reported events or not.

tests/events/src/lib.rs
mootz12: This was primarily added out of personal annoyance, trying to improve testing like is done here: https://github.com/blend-capital/blend-contracts-v2/blob/main/test-suites/tests/test_pool.rs#L83-L108

Transfers are verified via `balance` checks, which IMO is the way things like this should be tested. IMO I don't think testing the event logic of other contracts is necessary. In this case, I only care that the correct event I intended to emit during this call was emitted.

If we want to avoid recommending this, I can remove the `contains` fn. The primary lift of this change is the `matches` function, and you could just do `env.events().all().any(...` yourself. No strong opinions either way, as long as I can directly compare `contractevent` to the tuple event type.

soroban-sdk/src/tests/contract_event.rs
mootz12: This is a good point, good catch.

We could add an assertion macro that helps with this `assert_event_eq!` that essentially calls `event.match` and prints debug info for both events if false.

tests/events/src/lib.rs
dmkozh: > If we want to avoid recommending this, I can remove the contains fn. 

I don't really mind that as one of the tools, but I also think an assertion on the full event list with a contract id filter would be useful. Does that sound like a reasonable compromise between coverage and usability?

> Transfers are verified via balance checks, which IMO is the way things like this should be tested. IMO I don't think testing the event logic of other contracts is necessary

That's a fair point, though the same can be told about the balances - if you are concerned with the side effects beyond your contract, then you either need to test the state of the other contracts, or events of the other contracts. I'm not sure which one is more 'correct', but for now just checking the balance is clearly more convenient. Maybe we could think of something to make standard token events equally convenient?

mootz12: > I also think an assertion on the full event list with a contract id filter would be useful. 

I agree. Added a function to `testutils::Events` that allows users to just grab a vec of events emitted by a specific contract_id.

mootz12:
closing in favor of: #1638

PR #1638, opened 2025-12-11
Simplify contract event testing

mootz12:
### What

Adds a testutils struct `ContractEvents` to be returned by `env.events().all()` that now returns the XDR variants of the emitted contract events. To simplify testing, Events defined by `contractevents` can be converted to the XDR variant with `to_contract_event`.

Additional partial eq impl were added to maintain backwards compatibility with users who test with existing methods. This will impact tests where users directly check the length of the events array, via `all().len()`, who will now need to write `all().events.len()`.

Lastly, this also allows the debug information when testing events to display actual values, making it easier to determine what went wrong during test failures:

Before:
```
left: Vec(Ok((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Vec(Ok(Symbol(transfer)), Ok(Address(obj#35)), Ok(Address(obj#37))), Map(obj#43))))
right: Vec(Ok((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Vec(Ok(Symbol(transfer)), Ok(Address(obj#11)), Ok(Address(obj#29))), Map(obj#53))))
```

After:
```
left: [ContractEvent { ext: V0, contract_id: Some(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000001))), type_: Contract, body: V0(ContractEventV0 { topics: VecM([Symbol(ScSymbol(StringM(transfer))), Address(Contract(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000002)))), Address(Account(AccountId(PublicKeyTypeEd25519(Uint256(0000000000000000000000000000000000000000000000000000000000000003)))))]), data: Map(Some(ScMap(VecM([ScMapEntry { key: Symbol(ScSymbol(StringM(amount))), val: I128(Int128Parts { hi: 0, lo: 1 }) }, ScMapEntry { key: Symbol(ScSymbol(StringM(to_muxed_id))), val: U64(1) }])))) }) }]
right: [ContractEvent { ext: V0, contract_id: Some(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000001))), type_: Contract, body: V0(ContractEventV0 { topics: VecM([Symbol(ScSymbol(StringM(transfer))), Address(Contract(ContractId(Hash(0000000000000000000000000000000000000000000000000000000000000002)))), Address(Account(AccountId(PublicKeyTypeEd25519(Uint256(0000000000000000000000000000000000000000000000000000000000000003)))))]), data: Map(Some(ScMap(VecM([ScMapEntry { key: Symbol(ScSymbol(StringM(amount))), val: I128(Int128Parts { hi: 0, lo: 0 }) }, ScMapEntry { key: Symbol(ScSymbol(StringM(to_muxed_id))), val: U64(1) }])))) }) }]
```

An example test:

```rust
    #[test]
    fn test_event() {
        let env = Env::default();

        let from = Address::generate(&env);
        let to = MuxedAddress::generate(&env);
        let amount = 1i128;

        client.transfer(&from, &to, &amount);

        assert_eq!(
            env.events().all(),
            std::vec![Transfer {
                from: from.clone(),
                to: to.address(),
                amount,
                to_muxed_id: to.id(),
            }
            .to_contract_event(&env, &contract_id)],
        );
    }
```

### Why

Solves #1566. Makes it easier to validate events during tests. 

### Known limitations

None


soroban-sdk/src/address.rs
```diff
@@ -388,9 +388,9 @@ impl crate::testutils::Address for Address {
     }
 }
 
-#[cfg(not(target_family = "wasm"))]
+#[cfg(any(not(target_family = "wasm"), test, feature = "testutils"))]
 impl Address {
-    pub(crate) fn contract_id(&self) -> ContractId {
+    pub fn contract_id(&self) -> ContractId {
```
leighmcculloch: Can we mark these as doc hidden since they're primarily intended for use by generated code?

soroban-sdk/src/events.rs
```diff
@@ -58,13 +58,30 @@ impl Debug for Events {
...
     fn topics(&self, env: &Env) -> Vec<Val>;
     fn data(&self, env: &Env) -> Val;
 
     fn publish(&self, env: &Env) {
         env.events().publish_event(self);
     }
+
+    /// Convert this event and the given contract_id into a [`xdr::ContractEvent`] object.
+    #[cfg(any(test, feature = "testutils"))]
+    fn to_contract_event(&self, env: &Env, contract_id: &crate::Address) -> xdr::ContractEvent {
```
leighmcculloch: I'm inclined to be explicit here with the name:
```suggestion
    fn to_xdr(&self, env: &Env, contract_id: &crate::Address) -> xdr::ContractEvent {
```

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +413,103 @@ pub mod budget {
...
 }
 
+#[derive(Clone)]
+pub struct ContractEvents {
+    env: Env,
+    pub events: std::vec::Vec<xdr::ContractEvent>,
+}
+
+impl ContractEvents {
+    pub fn new(env: &Env, events: std::vec::Vec<xdr::ContractEvent>) -> Self {
```
leighmcculloch: Can we make this function pub(crate)?

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +413,103 @@ pub mod budget {
...
+        ContractEvents {
+            env: env.clone(),
+            events,
+        }
+    }
+
+    /// Creates a new ContractEvents struct that only includes events emitted
+    /// by the provided contract_id.
+    pub fn filter_by_contract_id(&self, contract_id: &crate::Address) -> Self {
```
leighmcculloch: Suggest:
```suggestion
    pub fn filter_by_contract(&self, addr: &crate::Address) -> Self {
```

soroban-sdk/src/testutils.rs
```diff
@@ -413,15 +413,103 @@ pub mod budget {
...
+        if len != other.len() {
+            return false;
+        }
+
+        for i in 0..len {
+            let (contract_id, topics, data) = match other.get(i) {
+                Some(obj) => obj,
+                None => return false,
+            };
+            let obj = self.events[i as usize].clone();
```
leighmcculloch: Shouldn't need to do this clone I think?

tests/events/src/lib.rs
```diff
@@ -63,27 +64,13 @@ mod test {
...
-                            <_ as IntoVal<Env, Val>>::into_val(&1i128, &env)
-                        ),
-                        (
-                            Symbol::new(&env, "to_muxed_id"),
-                            <_ as IntoVal<Env, Val>>::into_val(&to.id().unwrap(), &env)
-                        ),
-                    ]
-                    .to_val()
-                ),
-            ],
```
leighmcculloch, line 86: Can we keep the old assertions to continue to confirm that those types of assertions continue to work?

And do the new assertions too, just do both.

soroban-sdk/src/tests/contract_event.rs
```diff
@@ -18,34 +19,33 @@ fn test_defaults() {
...
-                    &env,
-                    (
-                        symbol_short!("value"),
-                        <_ as IntoVal<Env, Val>>::into_val(&symbol_short!("hello"), &env),
-                    ),
-                ]
-                .into_val(&env)
-            ),
-        ],
-    );
```
leighmcculloch, line 48: Can we retain the old assertions too?

leighmcculloch (commented):
👏🏻 I like this approach. It breaks the API, but softens that break for the common use cases. Few comments inline. Some of copilots comments seem like worth addressing as minor improvements too.

soroban-sdk/src/tests/contract_event.rs
mootz12: This is covered by `test_event_comparison_tuple_vec`, I didn't think there was much value in retaining the additional check.

No strong opinion here, so lmk if you'd still like to see this.

brozorec:
> * Since you folks have already been building event assertions utilities, what do you think of the change in this PR that would expose utilities for asserting on events using contractevent types and their xdr counterparts instead of the tuple list?

I'm supporting this change. Being able to directly use `contractevent` is more intuitive than getting and asserting the event data from the tuple.

> * If we introduced the breakage, changing the return value of `.all()`, how disruptive would that be? We can of course make it a new function, `.all2()` or something else.

I'd say no need to have both variants. We'll refactor once that's out.

soroban-sdk/src/address.rs
leighmcculloch: Looking at the latest diff, it doesn't look to me like there's any need for contract_id to be exposed anymore. The only uses are internal, not in generated code. What am I missing, is there a need to expose it?

leighmcculloch (approved):
Looks good.

I pushed a couple tweaks that aren't critical. I had tried them locally because I wasn't sure if they were possible and they worked so pushed them instead of posting comments with them. If they cut across anything you were planning though we can revert them.

leighmcculloch (approved):

Issue #1636, opened 2025-12-09
Add methods to extract/construct Address from raw 32-byte payloads


leighmcculloch:
### What problem does your feature solve?

Currently there's no way to extract the raw 32-byte payload (contract hash or Ed25519 public key) from an `Address`, or to construct an `Address` from such a payload. This functionality is useful for:

- **Cross-chain interoperability**: Working with raw public keys/contract hashes for bridges and cross-chain protocols that need to pass addresses across different systems

### What would you like to see?

Add methods to the `Address` type to extract and construct from raw 32-byte payloads:

```rust
pub enum AddressPayloadType {
    AccountEd25519PublicKey,
    ContractHash,
}

impl Address {
    /// Extracts the 32-byte payload from the address.
    /// Returns the payload type and raw bytes.
    pub fn payload(&self, env: &Env) -> Option<(AddressPayloadType, Bytes)>;

    /// Constructs an Address from a payload type and 32-byte payload.
    pub fn from_payload(env: &Env, payload_type: AddressPayloadType, payload: &Bytes) -> Address;
}
```

An experimental implementation exists at [`soroban-address-payload-ext`](https://github.com/leighmcculloch/soroban-address-payload-ext) that demonstrates how this could be implemented using XDR serialization/deserialization.

It might be suggested that the protocol should be changed to provide host functions that return this data. However, porting in this implementation using XDR doesn't prevent us from optimizing it in the future to use protocol functions should it ever become apparent they are needed.

### What alternatives are there?

- Manually serialize/deserialize through XDR - verbose and error-prone
- Use strkeys and parse them - requires a base32 decoder/encoder
- Do nothing

mootz12:
+1 to seeing this get added to SDK

> I think one thing we can do to try to prevent misuse is instead of adding these functions directly onto the Address type, we can keep them behind an extension trait

I personally don't think this is necessary. If a dev has the idea to manually implement ED25519 auth, I don't think the requirement of importing a trait is gonna stop them. Documentation on this fn should make it clear that using the ED25519 payload directly for custom verification is not safe.

This being said, if we opt to put this behind a trait something along the lines of `internals` would make more sense for naming.

leighmcculloch:
> I wonder if we can make this not appear in the autocompletion by default.

Another option is to put the functions behind the hazmat feature.
[...]

PR #1639, opened 2025-12-11
Split hazmat feature into granular sub-features

### LINKED ISSUE #1637: Split `hazmat` feature into granular sub-features

Currently, enabling the `hazmat` feature exposes all hazardous functionality in the SDK at once. This creates two problems:

1. **Accidental exposure**: A developer who needs to use just one specific hazmat function (e.g., `secp256k1_recover` for cross-chain interoperability) inadvertently gains access to all other hazmat functionality. During development, they might accidentally use other exposed hazmat features without realizing they are hazardous.

2. **Varying levels of risk**: When deciding whether something should be behind the `hazmat` feature, we need to ensure it meets the same bar as everything else already in `hazmat`. But over time, we're likely to have different types of hazmat functionality: some that are truly dangerous, and others that are less risky but still not something we want exposed to general contract developers by default. A single `hazmat` feature forces us to treat all of these the same way.

This concern was raised during the discussion in #1636 about adding address payload extraction methods, where the question of how to appropriately guard potentially dangerous functionality came up.

Split the `hazmat` feature into granular sub-features, allowing developers to enable only the specific hazmat functionality they need. The last component should use underscores to exactly match the function name. For example:

- `hazmat-crypto-secp256k1_recover`
- `hazmat-crypto-secp256r1_verify`
- `hazmat-crypto-ed25519_verify`

Additional hazmat sub-features could be added over time as new hazmat functionality is introduced.

The existing `hazmat` feature would remain and would enable all hazmat sub-features, maintaining backwards compatibility. However, the recommended best practice going forward would be for developers to enable only the specific hazmat sub-feature they need.

- **Keep the current single `hazmat` feature**: Simpler, but exposes developers to more functionality than they need

### Considerations

Every additional feature adds another combination of builds that need to be tested. An acceptable approach may be to only test the `hazmat` feature as a whole in CI, rather than testing every individual sub-feature combination.

---

**Related:** This option was discussed in https://github.com/stellar/rs-soroban-sdk/issues/1636#issuecomment-3638897949

leighmcculloch:
### What
Add granular hazmat feature (`hazmat-crypto`) for cryptographic hazmat functions. The umbrella `hazmat` feature remains and enables all hazmat sub-features.

### Why
Developers needing hazmat crypto functions previously had to enable all hazmat functionality, risking accidental use of other hazardous APIs. Granular features let developers enable only what they need, reducing exposure to unintended risks.

As we expand the types of hazardous functionality in the SDK we make it easier to treat some functionality as hazardous even though it doesn't fit the traditional super-hazardous threshold of crypto functions. For example: https://github.com/stellar/rs-soroban-sdk/issues/1636

Close #1637


soroban-sdk/Cargo.toml
```diff
@@ -61,8 +62,17 @@ sha2 = "0.10.7"
 [features]
 alloc = []
 testutils = ["soroban-sdk-macros/testutils", "soroban-env-host/testutils", "soroban-ledger-snapshot/testutils", "dep:ed25519-dalek", "dep:arbitrary", "dep:derive_arbitrary", "dep:ctor", "dep:soroban-ledger-snapshot"]
-hazmat = []
 docs = []
 
+# Umbrella feature that enables all hazmat sub-features (backwards compatible)
+hazmat = [
+  "hazmat-crypto-secp256k1_recover",
+  "hazmat-crypto-secp256r1_verify",
```
jayz22: I'm not sure about one sub feature for each function, as this list might get bloated quickly (I'm going to add some soon for poseidon/poseidon2). I feel grouping them under `hazmat-crypto` (and another for "non-crypto") is sufficient now, but willing to go with it if you feel strongly.

leighmcculloch: Or maybe even areas, such as `hazmat-crypto-sec256k1` and then `hazmat-crypto-podeidon`. Wdyt? cc @dmkozh

tests-expanded/test_account_tests.rs
leighmcculloch, line 1: I'm not sure why the diff happens with the changes in this PR, but it seems harmless and seems consistent.

leighmcculloch: These changes happened because I used a newer version of cargo-expand.

soroban-sdk/Cargo.toml
jayz22: technically `secp256k1` and `secp256r1` are two different areas, so I don't think that'll help much.

leighmcculloch: That's a good point. I think doing as you say will be sufficient, `hazmat-crypto` and `hazmat-address`. I'll update the PRs to do that.

leighmcculloch: Updated in f9f1fcab8d5bdb257513267e794fe0cd568a5545. I'll keep this open for @dmkozh to take another look too.

soroban-sdk/Cargo.toml
```diff
@@ -61,8 +62,15 @@ sha2 = "0.10.7"
 [features]
 alloc = []
 testutils = ["soroban-sdk-macros/testutils", "soroban-env-host/testutils", "soroban-ledger-snapshot/testutils", "dep:ed25519-dalek", "dep:arbitrary", "dep:derive_arbitrary", "dep:ctor", "dep:soroban-ledger-snapshot"]
-hazmat = []
 docs = []
 
+# Umbrella feature that enables all hazmat sub-features (backwards compatible)
+hazmat = [
```
dmkozh, line 68: Are we planning to add hazmat-address here as well? If so, it seems like it would be quite easy to pick up all the new hazmat features as they come - maybe it should be reserved as a legacy alias for hazmat-crypto?

soroban-sdk/Cargo.toml
dmkozh: I'm fine with less granular separation, it seems like a reasonable compromise.

soroban-sdk/Cargo.toml
leighmcculloch: Do you mean have `hazmat` as an alias only for crypto? Yeah that's not a bad idea. I'm also thinking we just get rid of it too in the next major release.

dmkozh: Right, that would be the plan. Make a non-breaking change for the minor release, but only use it for crypto for compatibility. Then make a breaking change in the major release. This way users of the address functionality won't be affected at all.

Makefile
```diff
@@ -14,8 +14,15 @@ doc: fmt
...
 
+# Run tests. 
+# The docs feature is excluded because it is a market for docs builds. The
+# hazmat granular features are excluded because all hazmat features are tested
+# together with the umbrella hazmat feature.
 test-only:
-	cargo hack --feature-powerset --ignore-unknown-features --features testutils --exclude-features docs test
+	cargo hack --feature-powerset --ignore-unknown-features --features testutils \
+		--exclude-features docs \
+		--exclude-features hazmat-crypto \
```
jayz22, line 24: Nit: all new hazmat sub types need to be included here in the future? would it be better to exclude 'hazmat' instead (test with sub features combinations), since this is easier to maintain and also more aligned with what cargo hack is intended for?

soroban-sdk/Cargo.toml
leighmcculloch: Actually, we may not want to do this and instead keep the umbrella feature because it's not possible with cargo hack to group features unless the features are used on all the crates. It's a limitation of its `--group-features` and `--ignore-unknown-features` options. And these features are only used on the Soroban SDK crate so we have to ignore the features for other  crates.

This means the only way we can run the tests for all the hazmat things in a single group is to have a feature that supersets them all and then exclude the sub-features. The reason we'd want to group them is so that we don't blow out test times with yet another feature combination.

Makefile
leighmcculloch: The reason we do this isn't to exclude Soroban, it's to group the features together. See this comment above. 
https://github.com/stellar/rs-soroban-sdk/pull/1639#discussion_r2612583311

soroban-sdk/Cargo.toml
leighmcculloch: I'll keep it as an umbrella for now we can still delete it at the next major if anyone thinks the tradeoff isn't right.

PR #1640, opened 2025-12-11
Add Address to_payload and from_payload

### LINKED ISSUE #1636: Add methods to extract/construct Address from raw 32-byte payloads

leighmcculloch:
### What
Add `to_payload` and `from_payload` methods on `Address` to extract and construct raw 32-byte payloads (Ed25519 public keys or contract hashes). Add a new `address_payload` module with the `AddressPayload` enum. Gate behind a `hazmat-address` feature flag to discourage use outside careful intentional situations that absolutely require them.

### Why
Cross-chain interoperability protocols need to pass raw address payloads between systems. Previously this required manual XDR serialization which was verbose and error-prone. Feature flags prevent accidental misuse by developers who might incorrectly use extracted Ed25519 keys for signature verification, as account master keys may not be active signers.

### API

```rust
// With feature = "hazmat-address" or feature = "hazmat"
use soroban_sdk::address_payload::AddressPayload;

// Extract payload from address
let payload: Option<AddressPayload> = address.to_payload();

// Construct address from payload
let address = Address::from_payload(&env, payload);

// AddressPayload methods
let address = payload.to_address(&env);
let payload = AddressPayload::from_address(&address);
```

Close #1636

### Merging

Intended to be merged to `main` after:
- #1639

### DIFF
--- a/soroban-sdk/src/address_payload.rs
+++ b/soroban-sdk/src/address_payload.rs
@@ -52,26 +52,28 @@ impl AddressPayload {
     pub fn to_address(&self, env: &Env) -> Address {
         use crate::xdr::FromXdr;
         // Build XDR header and get payload bytes based on payload type:
-        let (header, payload_bytes): (&[u8], &BytesN<32>) = match self {
+        let (header, payload_bytes) = match self {
             AddressPayload::AccountIdPublicKeyEd25519(bytes) => (
-                &[
+                [
                     0, 0, 0, 18, // ScVal::Address
                     0, 0, 0, 0, // ScAddress::Account
                     0, 0, 0, 0, // PublicKey::PublicKeyTypeEd25519
-                ],
-                bytes,
+                ]
+                .as_slice(),
+                bytes.as_bytes(),
             ),
             AddressPayload::ContractIdHash(bytes) => (
-                &[
+                [
                     0, 0, 0, 18, // ScVal::Address
                     0, 0, 0, 1, // ScAddress::Contract
-                ],
-                bytes,
+                ]
+                .as_slice(),
+                bytes.as_bytes(),
             ),
         };
 
         let mut xdr = Bytes::from_slice(env, header);
-        xdr.append(&Bytes::from(payload_bytes.clone()));
+        xdr.append(payload_bytes);
 
         Address::from_xdr(env, &xdr).unwrap_optimized()
     }
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -110,6 +110,10 @@ macro_rules! impl_bytesn_repr {
                 Self(bytes)
             }
 
+            pub fn into_bytes(self) -> BytesN<$size> {
+                self.0
+            }
+
             pub fn to_bytes(&self) -> BytesN<$size> {
                 self.0.clone()
             }
@@ -1110,14 +1114,14 @@ impl<const N: usize> From<BytesN<N>> for Val {
 impl<const N: usize> From<BytesN<N>> for Bytes {
     #[inline(always)]
     fn from(v: BytesN<N>) -> Self {
-        v.0
+        v.into_bytes()
     }
 }
 
 impl<const N: usize> From<&BytesN<N>> for Bytes {
     #[inline(always)]
     fn from(v: &BytesN<N>) -> Self {
-        v.0.clone()
+        v.to_bytes()
     }
 }
 
@@ -1156,6 +1160,18 @@ impl<const N: usize> BytesN<N> {
 
     pub fn env(&self) -> &Env {
         self.0.env()
+    }
+
+    pub fn as_bytes(&self) -> &Bytes {
+        &self.0
+    }
+
+    pub fn into_bytes(self) -> Bytes {
+        self.0
+    }
+
+    pub fn to_bytes(&self) -> Bytes {
+        self.0.clone()
     }
 
     pub fn as_val(&self) -> &Val {


soroban-sdk/src/tests/address_payload.rs
```diff
@@ -0,0 +1,88 @@
...
+            0xd7, 0x92, 0x8b, 0x72, 0xc2, 0x70, 0x3c, 0xcf, 0xea, 0xf7, 0xeb, 0x9f, 0xf4, 0xef,
+            0x4d, 0x50, 0x4a, 0x55, 0xa8, 0xb9, 0x79, 0xfc, 0x9b, 0x45, 0x0e, 0xa2, 0xc8, 0x42,
+            0xb4, 0xd1, 0xce, 0x61,
+        ],
+    );
+
+    let address = Address::from_payload(&env, AddressPayload::ContractHash(payload));
+    assert_eq!(
+        address.to_string().to_string(),
+        "CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC"
```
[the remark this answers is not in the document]
leighmcculloch:
```suggestion
    assert_eq!(
        address.to_string(),
        "CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC"
```

soroban-sdk/src/address_payload.rs
```diff
@@ -0,0 +1,142 @@
...
+//! For account addresses, the Ed25519 public key corresponds to the account's
+//! master key, which depending on the configuration of that account may or may
+//! not be a signer of the account. Do not use this for custom Ed25519 signature
+//! verification as a form of authentication.
+
+use crate::{unwrap::UnwrapOptimized, Address, Bytes, BytesN, Env};
+
+/// The payload contained in an [`Address`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum AddressPayload {
```
dmkozh, line 26: Just an alternative to consider: since this is meant to be used as data payload (and the name suggests the same), is there any reason to be specific about the meaning of the 32 bytes?
I think something like
`Account(BytesN<32>)/Contract(BytesN<32>)` would make more sense semantically. We can tell what these are in comments, but ultimately these should be treated as opaque 32 bytes for the purpose of this struct. 

I've also been thinking about using
```
struct AddressPayload {
  address_kind: Account|Contract,
  payload: BytesN<32>
}
```
This has a downside of not being extensible to non-32-byte payloads, though I'm not sure if that's something worth planning around. The upside is that it resembles the message on the wire better.
Any thoughts on this?

leighmcculloch: I'd prefer to be transparent. This is a hazmat API intended for "you know what you're doing" and so less abstractions and more access to the raw. There's also no guarantee that future address types will be 32-bytes in the account branch and it's possible a future address type may be a different size and need representing with a new arm.

leighmcculloch: Another reason is that the enum is providing a namespace / scope around a payload, and if the network was to introduce a new 32-byte public key type, even though it fits into that same size, it would be a new namespace because the identity on network is not simply the 32-byte payload, it is the 32-byte payload plus the public key discriminant and that public key discriminant creates a new namespace.

dmkozh: >  so less abstractions and more access to the raw.

I'm just not sure if this information is relevant for the functionality that is only meant for building the addresses out of parts. It's a bit weird to name something 'ed25519 key' and then have a disclaimer everywhere that it's actually not safe to use it as ed25519, i.e. it's supposed to be treated as opaque payload.

> Another reason is that the enum is providing a namespace / scope around a payload,

Yeah, the struct approach was just an idea to consider, I don't feel strongly about it.

leighmcculloch: > It's a bit weird to name something 'ed25519 key' and then have a disclaimer everywhere that it's actually not safe to use it as ed25519

It's not safe to blindly assume that a ed25519 public key in the address is the signer. I don't think the warning was being specific enough about that so I modified it a bit. The payload is still an ed25519 public key though, and I think someone building cross contract functionality is probably going to be aware that the payload within the G is an ed25519 public key, so I don't think obscuring that information is serving much of a purpose is it?

leighmcculloch: I renamed `ContractHash` to `ContractIDHash` to better match how the XDR describes the terms because it is capture as an `ScAddress::Contract(ContractID(Hash(...)))`.

leighmcculloch: The names are a clear mapping:
`AddressPayload::AccountIdPublicKeyEd25519` <> `ScAddress::Account(AccountId(PublicKey::Ed25519))`
`AddressPayload::ContractIdHash` <> `ScAddress::Contract(ContractId(Hash))`

PR #1648, opened 2025-12-16
Issues/1310 shrink snapshot file

### LINKED ISSUE #1310: Ledger snapshots store both the key and values unnecessarily
The ledger snapshot crate stores the key and value for each ledger entry, but that's unnecessary since the key data is store inside the entry, so more data is being stored than is necessary.

This would be a breaking change, but we could drop the key and rematerialize it, make the existing key field optional in the next major release.

Ideally this change would be made in a way it is not breaking.

mootz12:
### What

Add a compact format for `LedgerSnapshot` to prevent having to write duplicate `LedgerKey` data to files.

On serialization, the `LedgerKey` is derived from the `LedgerEntry` to keep the public interface the same.

### Why

Fixes #1310 .

Tested against a large snapshot (82mb). File was reduced to 53mb, or ~35% reduction.

Roundtrip to serialize the large snapshot and write it back takes 60s with the old way, and 22s for the new way. A good half of this optimization comes from using `BufReader`.

The memory footprint does increase by the size of the snapshot due to conversions between the compact and ledgerSnapshot type, and up to 2x the snapshot if we are reading a snapshot of the old format. I assume this is due to how serde manages the deserialization retry logic for enums.

### Known limitations

N/A

### DIFF
--- a/soroban-ledger-snapshot/src/tests.rs
+++ b/soroban-ledger-snapshot/src/tests.rs
@@ -30,7 +30,9 @@ fn test_snapshot_from_v1() {
     snapshot.write(&mut written).unwrap();
     let written_str = String::from_utf8(written).unwrap();
 
-    assert_eq!(written_str, expected_str);
+    let expected_normalized = expected_str.replace("\r\n", "\n");
+    let written_normalized = written_str.replace("\r\n", "\n");
+    assert_eq!(written_normalized, expected_normalized);
 }
 
 #[test]
@@ -49,5 +51,7 @@ fn test_snapshot_roundtrip() {
     snapshot.write(&mut written).unwrap();
     let written_str = String::from_utf8(written).unwrap();
 
-    assert_eq!(written_str, expected_str);
+    let expected_normalized = expected_str.replace("\r\n", "\n");
+    let written_normalized = written_str.replace("\r\n", "\n");
+    assert_eq!(written_normalized, expected_normalized);
 }


soroban-ledger-snapshot/src/lib.rs
```diff
@@ -22,10 +25,56 @@ pub enum Error {
...
+}
+
+/// Internal structure used to represent a [`LedgerSnapshot`] in a more compact form. The
+/// [`LedgerKey`] is omitted from the ledger_entries tuple since it is duplicated within
+/// the [`LedgerEntry`].
 #[serde_as]
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "snake_case")]
-pub struct LedgerSnapshot {
+struct LedgerSnapshotV1Compact {
```
leighmcculloch: I think we can call this file format V2. What do you think?

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -22,10 +25,56 @@ pub enum Error {
...
+    pub protocol_version: u32,
+    pub sequence_number: u32,
+    pub timestamp: u64,
+    #[serde_as(as = "serde_with::hex::Hex")]
+    pub network_id: [u8; 32],
+    pub base_reserve: u32,
+    pub min_persistent_entry_ttl: u32,
+    pub min_temp_entry_ttl: u32,
+    pub max_entry_ttl: u32,
+    pub ledger_entries: Vec<(Box<LedgerEntry>, Option<u32>)>,
```
leighmcculloch: If we think about this as a V2 format and also address some things which are maybe not great about this format, do you think we should, instead of making the ledger entries a tuple, which means that there's this integer value in the option in the second field that isn't particularly intuitive, replace that with a struct so that the fields are named? I don't feel strongly about this, and one downside of doing this is it would make the file larger.

leighmcculloch (approved):
👏🏻 Love it. A couple of thoughts inline, but this looks fantastic. Good use of the manual serde impls to retain the same exported interface that supports use with serde and makes all this an implementation detail not exposed to others.

dmkozh (commented):
LGTM, but there seems to be some issue with loading the snapshots in CI

leighmcculloch:
> The memory footprint does increase by the size of the snapshot due to conversions between the compact and ledgerSnapshot type

This is surprising. Are we cloning in places that's unnecessary, maybe? Can we try to understand this a little bit better? Because there may be some simple things we can do to address this. I expected that the memory footprint, once processed, should be the same as what it is today.

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -135,23 +184,46 @@ impl LedgerSnapshot {
     }
 }
 
+impl serde::Serialize for LedgerSnapshot {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let compact: LedgerSnapshotV1Compact = self.into();
+        compact.serialize(serializer)
```
leighmcculloch: It's a shame that we have to clone the entire structure to do the serialize. We could actually write that implementation manually instead of relying on serde-derive and a struct. What do you think?

leighmcculloch: Another option is we might be able to use `serde_with` or `serde_as` on the field itself. So instead of having to write the manual serde for the entire struct, we write a custom serializer for the one field that contains the Vec.

leighmcculloch:
```rust
    #[serde_as(as = "OmitAndRehydrateKeySerDer")]
    pub ledger_entries: Vec<(Box<LedgerKey>, (Box<LedgerEntry>, Option<u32>))>,
```

mootz12: Good call, used `serde_as` and the solution looks simpler.

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -136,22 +200,23 @@ impl LedgerSnapshot {
...
+    /// Read in a [`LedgerSnapshot`] from a reader.
     pub fn read(r: impl Read) -> Result<LedgerSnapshot, Error> {
         Ok(serde_json::from_reader::<_, LedgerSnapshot>(r)?)
     }
 
-    // Read in a [`LedgerSnapshot`] from a file.
+    /// Read in a [`LedgerSnapshot`] from a file.
     pub fn read_file(p: impl AsRef<Path>) -> Result<LedgerSnapshot, Error> {
-        Self::read(File::open(p)?)
+        let reader = BufReader::new(File::open(p)?);
```
leighmcculloch, line 219: Nice catch 👍🏻

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -35,11 +39,71 @@ pub struct LedgerSnapshot {
...
+        S: serde::Serializer,
+    {
+        let entries: Vec<LedgerEntryExt> = source
+            .iter()
+            .map(|(_, (entry, live_util))| LedgerEntryExt {
+                entry: entry.clone(),
+                live_util: *live_util,
+            })
+            .collect();
+        entries.serialize(serializer)
```
leighmcculloch: To reduce the memory footprint here you could serialize each element one at a time. Something like:
```
        use serde::ser::SerializeSeq;
        let mut seq = serializer.serialize_seq(Some(source.len()))?;
        for (_, (entry, live_util)) in source {
            seq.serialize_element(&LedgerEntryExtRef {
                entry,
                live_util: *live_util,
            })?;
        }
        seq.end()
```
Assuming you also had something like:
```
#[derive(serde::Serialize)]
struct LedgerEntryExtRef<'a> {
    entry: &'a LedgerEntry,
    live_util: Option<u32>,
}
```

mootz12: Good call!

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -35,11 +39,80 @@ pub struct LedgerSnapshot {
...
     pub ledger_entries: Vec<(Box<LedgerKey>, (Box<LedgerEntry>, Option<u32>))>,
 }
 
+/// Extended ledger entry that includes the live util ledger sequence. Provides a more compact
+/// form of the tuple used in [`LedgerSnapshot::ledger_entries`], to reduce the size of the snapshot
+/// when serialized to JSON.
+#[derive(Debug, Clone, serde::Deserialize)]
+struct LedgerEntryExt {
+    entry: Box<LedgerEntry>,
+    live_util: Option<u32>,
```
leighmcculloch: There's at typo here.
```suggestion
    live_until: Option<u32>,
```

soroban-ledger-snapshot/src/lib.rs
```diff
@@ -35,11 +39,80 @@ pub struct LedgerSnapshot {
...
+}
+
+/// Extended ledger entry that includes the live util ledger sequence, and the entry by reference.
+/// Used to reduce memory usage during serialization.
+#[derive(serde::Serialize)]
+struct LedgerEntryExtRef<'a> {
+    entry: &'a Box<LedgerEntry>, // Reference = no clone
+    live_util: Option<u32>,
```
leighmcculloch: And here.
```suggestion
    live_until: Option<u32>,
```

leighmcculloch (approved):
👏🏻 Looks great. There's a typo worth fixing because it makes its way into the JSON files. Otherwise LGTM.

soroban-ledger-snapshot/src/lib.rs
leighmcculloch: 16038d1

soroban-ledger-snapshot/src/lib.rs
leighmcculloch: 16038d1

PR #1662, opened 2026-01-02
feat: add try_as_contract test function

### LINKED ISSUE #1434: Add `try_as_contract` function

Currently we only provide `as_contract` that unwraps the return value of any invocation that happens inside. Thus if a function inside `as_contract` is expected to fail, the only way to make an assertion for the failure is to use `should_panic`. However `should_panic` is rather inconvenient (only one allowed per test) and brittle (there is a risk of a matching panic happening before the expected line. This is relevant to the library functions that operate on `env`, but not on any particular contract.

`try_as_contract` function that behaves just like `as_contract`, but returns a `Result` of the invocation instead of unwrapping it

Keep using `should_panic`

mootz12:
### What

Adds a `try_as_contract` fn that gracefully returns errors like `try_invoke_contract`.

### Why

Fixes #1434 

### Known limitations

- [x] Marked as draft until the necessary env change is made: https://github.com/stellar/rs-soroban-env/pull/1628

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -1572,7 +1572,7 @@ impl Env {
     ///     client.set_storage(&key, &val);
     ///
     ///     // Successfully read the storage key
-    ///     let result = env.as_contract::<Symbol>(&contract_id, || {
+    ///     let result = env.as_contract(&contract_id, || {
     ///         env.storage()
     ///             .persistent()
     ///             .get::<Symbol, Symbol>(&key)
@@ -1635,21 +1635,19 @@ impl Env {
     ///             .get::<Symbol, Symbol>(&key)
     ///             .unwrap()
     ///     });
-    ///     assert!(result.is_ok());
-    ///     assert_eq!(result.unwrap(), val);
+    ///     assert_eq!(result, Ok(val));
     ///
     ///     // Attempting to extend TTL of a non-existent key throws an error
     ///     let new_key = Symbol::new(&env, "baz");
-    ///     let result = env.try_as_contract::<_, Error>(&contract_id, || {
+    ///     let result = env.try_as_contract(&contract_id, || {
     ///         env.storage().persistent().extend_ttl(&new_key, 1, 100);
     ///     });
-    ///     assert!(result.is_err());
     ///     assert_eq!(
-    ///         result.unwrap_err(),
-    ///         Ok(Error::from_type_and_code(
+    ///         result,
+    ///         Err(Ok(Error::from_type_and_code(
     ///             ScErrorType::Storage,
     ///             ScErrorCode::MissingValue
-    ///         ))
+    ///         )))
     ///     );
     /// }
     /// ```
--- a/soroban-sdk/src/tests/env.rs
+++ b/soroban-sdk/src/tests/env.rs
@@ -241,8 +241,7 @@ fn test_try_as_contract() {
             .get::<Symbol, Symbol>(&key)
             .unwrap()
     });
-    assert!(result.is_ok());
-    assert_eq!(result.unwrap(), val);
+    assert_eq!(result, Ok(val));
 }
 
 #[test]
@@ -258,14 +257,13 @@ fn test_try_as_contract_host_error() {
         // should error as key doesn't exist in storage
         env.storage().persistent().extend_ttl(&key, 1, 100);
     });
-    assert!(result.is_err());
     assert_eq!(
-        result.unwrap_err(),
-        Ok(Error::from_type_and_code(
+        result,
+        Err(Ok(Error::from_type_and_code(
             ScErrorType::Storage,
             ScErrorCode::MissingValue
-        ))
-    )
+        )))
+    );
 }
 
 #[test]
@@ -281,8 +279,7 @@ fn test_try_as_contract_host_error_contract_error_expected() {
         // should error as key doesn't exist in storage
         env.storage().persistent().extend_ttl(&key, 1, 100);
     });
-    assert!(result.is_err());
-    assert_eq!(result.unwrap_err(), Err(InvokeError::Abort));
+    assert_eq!(result, Err(Err(InvokeError::Abort)));
 }
 
 #[test]
@@ -295,8 +292,7 @@ fn test_try_as_contract_contract_error() {
     let result = env.try_as_contract::<_, ContractError>(&addr, || {
         panic_with_error!(&env, ContractError::AnError);
     });
-    assert!(result.is_err());
-    assert_eq!(result.unwrap_err(), Ok(ContractError::AnError));
+    assert_eq!(result, Err(Ok(ContractError::AnError)));
 }
 
 #[test]
@@ -309,8 +305,7 @@ fn test_try_as_contract_contract_error_unexpected_error() {
     let result = env.try_as_contract::<_, ContractError>(&addr, || {
         panic_with_error!(&env, Error::from_contract_error(99));
     });
-    assert!(result.is_err());
-    assert_eq!(result.unwrap_err(), Err(InvokeError::Contract(99)));
+    assert_eq!(result, Err(Err(InvokeError::Contract(99))));
 }
 
 #[test]
@@ -323,12 +318,11 @@ fn test_try_as_contract_panic() {
     let result = env.try_as_contract::<_, Error>(&addr, || {
         panic!("please don't do this when writing contracts");
     });
-    assert!(result.is_err());
     assert_eq!(
-        result.unwrap_err(),
-        Ok(Error::from_type_and_code(
+        result,
+        Err(Ok(Error::from_type_and_code(
             ScErrorType::WasmVm,
             ScErrorCode::InvalidAction
-        ))
+        )))
     );
 }


soroban-sdk/src/env.rs
```diff
@@ -85,6 +85,8 @@ pub use internal::TryFromVal;
 pub use internal::TryIntoVal;
 pub use internal::Val;
 pub use internal::VecObject;
+#[cfg(any(test, feature = "testutils"))]
+use soroban_env_host::HostError;
```
leighmcculloch: Can you move this `use` to be adjacent to the other testutil-only imports here and import it from internal just to match the existing pattern?
https://github.com/stellar/rs-soroban-sdk/blob/bd21054b0369924249826290ba15a7301084e0f6/soroban-sdk/src/env.rs#L490-L491

But, if we do https://github.com/stellar/rs-soroban-sdk/pull/1662#discussion_r2659910726 then it shouldn't need importing and exporting anyway.

soroban-sdk/src/env.rs
```diff
@@ -1550,6 +1552,26 @@ impl Env {
         t.unwrap()
     }
 
+    /// Run the function as if executed by the given contract ID. Returns an
+    /// error if the function execution fails for any reason.
+    ///
+    /// Used to write or read contract data, or take other actions in tests for
+    /// setting up tests or asserting on internal state.
+    pub fn try_as_contract<T>(&self, id: &Address, f: impl FnOnce() -> T) -> Result<T, HostError> {
```
leighmcculloch: For the return value I think this function should return an `InvokeError` and the same `Result<Result<>, Result<>>` structure for consistency with the `try_invoke_contract` function.
https://github.com/stellar/rs-soroban-sdk/blob/9415ceeb60bb8eabaa60227e57ef73bc7dd0cd07/soroban-sdk/src/env.rs#L407-L412

Why: So the dev has the same error handling experience as for invoking. The as_contract fn is basically like calling invoke_contract on a contract function. Also, like the other try_ functions the case where T is the wrong type should be handable as well.

leighmcculloch (commented):
Some feedback inline.

soroban-sdk/src/env.rs
mootz12: `*_as_contract` extracts the result of the closure without converting it to a Val. Thus, I kept the Ok result as `T`, otherwise we need to convert it to a val and back which felt weird.

Good call on the error, updated it to match the `try_invoke` method and added tests.

leighmcculloch (approved):

soroban-sdk/src/env.rs
```diff
@@ -1550,6 +1550,34 @@ impl Env {
         t.unwrap()
     }
 
+    /// Run the function as if executed by the given contract ID. Returns an
+    /// error if the function execution fails for any reason.
+    ///
+    /// Used to write or read contract data, or take other actions in tests for
+    /// setting up tests or asserting on internal state.
```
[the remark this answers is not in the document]
mootz12: What is our take on doctests? I see it's split for env functions if they exist or not.

Should we be doing it for env functions?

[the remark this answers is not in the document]
leighmcculloch: I would lean into adding them, executable ones though that will self-test, not examples that will rot over time.

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+    env.register_at(&addr, Contract, ());
+
+    let key = Symbol::new(&env, "foo");
+
+    let result = env.try_as_contract::<_, ContractError>(&addr, || {
+        // should error as key doesn't exist in storage
+        env.storage().persistent().extend_ttl(&key, 1, 100);
+    });
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err(), Err(InvokeError::Abort));
```
leighmcculloch: Does this work?
```suggestion
    assert_eq!(result, Err(Err(InvokeError::Abort)));
```

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+    let env = Env::default();
+
+    let addr = Address::generate(&env);
+    env.register_at(&addr, Contract, ());
+
+    let result = env.try_as_contract::<_, ContractError>(&addr, || {
+        panic_with_error!(&env, ContractError::AnError);
+    });
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err(), Ok(ContractError::AnError));
```
leighmcculloch: Does this work?
```suggestion
    assert_eq!(result, Err(Ok(ContractError::AnError)));
```

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+    let env = Env::default();
+
+    let addr = Address::generate(&env);
+    env.register_at(&addr, Contract, ());
+
+    let result = env.try_as_contract::<_, ContractError>(&addr, || {
+        panic_with_error!(&env, Error::from_contract_error(99));
+    });
+    assert!(result.is_err());
+    assert_eq!(result.unwrap_err(), Err(InvokeError::Contract(99)));
```
leighmcculloch: Does this work?
```suggestion
    assert_eq!(result, Err(Err(InvokeError::Contract(99))));
```

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+    let result = env.try_as_contract::<_, Error>(&addr, || {
+        panic!("please don't do this when writing contracts");
+    });
+    assert!(result.is_err());
+    assert_eq!(
+        result.unwrap_err(),
+        Ok(Error::from_type_and_code(
+            ScErrorType::WasmVm,
+            ScErrorCode::InvalidAction
+        ))
```
leighmcculloch: Does this work?
```suggestion
        result,
        Err(Ok(Error::from_type_and_code(
            ScErrorType::WasmVm,
            ScErrorCode::InvalidAction
        )))
```

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+        // should error as key doesn't exist in storage
+        env.storage().persistent().extend_ttl(&key, 1, 100);
+    });
+    assert!(result.is_err());
+    assert_eq!(
+        result.unwrap_err(),
+        Ok(Error::from_type_and_code(
+            ScErrorType::Storage,
+            ScErrorCode::MissingValue
+        ))
```
leighmcculloch: Does this work?
```suggestion
        result,
        Err(Ok(Error::from_type_and_code(
            ScErrorType::Storage,
            ScErrorCode::MissingValue
        )))
```

soroban-sdk/src/tests/env.rs
```diff
@@ -212,3 +220,115 @@ fn test_snapshot_file_disabled_after_creation() {
...
+    });
+
+    let result = env.try_as_contract::<Symbol, Error>(&addr, || {
+        env.storage()
+            .persistent()
+            .get::<Symbol, Symbol>(&key)
+            .unwrap()
+    });
+    assert!(result.is_ok());
+    assert_eq!(result.unwrap(), val);
```
leighmcculloch: Does this work?
```suggestion
    assert_eq!(result, Ok(val));
```

soroban-sdk/src/env.rs
```diff
@@ -1542,6 +1542,45 @@ impl Env {
...
+    ///     let client = HelloContractClient::new(&env, &contract_id);
+    ///
+    ///     let key = Symbol::new(&env, "foo");
+    ///     let val = Symbol::new(&env, "bar");
+    ///
+    ///     // Set storage using the contract
+    ///     client.set_storage(&key, &val);
+    ///
+    ///     // Successfully read the storage key
+    ///     let result = env.as_contract::<Symbol>(&contract_id, || {
```
leighmcculloch: Is the type annotation required here?
```suggestion
    ///     let result = env.as_contract(&contract_id, || {
```

soroban-sdk/src/env.rs
```diff
@@ -1555,6 +1594,88 @@ impl Env {
...
+    ///
+    ///     // Successfully read the storage key
+    ///     let result = env.try_as_contract::<Symbol, Error>(&contract_id, || {
+    ///         env.storage()
+    ///             .persistent()
+    ///             .get::<Symbol, Symbol>(&key)
+    ///             .unwrap()
+    ///     });
+    ///     assert!(result.is_ok());
+    ///     assert_eq!(result.unwrap(), val);
```
leighmcculloch: Does this work?
```suggestion
    ///     assert_eq!(result, Ok(val));
```

soroban-sdk/src/env.rs
```diff
@@ -1555,6 +1594,88 @@ impl Env {
...
+    ///
+    ///     // Attempting to extend TTL of a non-existent key throws an error
+    ///     let new_key = Symbol::new(&env, "baz");
+    ///     let result = env.try_as_contract::<_, Error>(&contract_id, || {
```
leighmcculloch: Is the error annotation required?
```suggestion
    ///     let result = env.try_as_contract(&contract_id, || {
```

soroban-sdk/src/env.rs
```diff
@@ -1542,6 +1542,45 @@ impl Env {
...
+    ///     // Successfully read the storage key
+    ///     let result = env.as_contract::<Symbol>(&contract_id, || {
+    ///         env.storage()
+    ///             .persistent()
+    ///             .get::<Symbol, Symbol>(&key)
+    ///             .unwrap()
+    ///     });
+    ///     assert_eq!(result, val);
+    /// }
+    /// ```
```
leighmcculloch, line 1583: 👏🏻 new docs ftw

soroban-sdk/src/env.rs
```diff
@@ -1555,6 +1594,88 @@ impl Env {
...
+    ///     let result = env.try_as_contract::<_, Error>(&contract_id, || {
+    ///         env.storage().persistent().extend_ttl(&new_key, 1, 100);
+    ///     });
+    ///     assert!(result.is_err());
+    ///     assert_eq!(
+    ///         result.unwrap_err(),
+    ///         Ok(Error::from_type_and_code(
+    ///             ScErrorType::Storage,
+    ///             ScErrorCode::MissingValue
+    ///         ))
```
leighmcculloch: Does this work?
```suggestion
    ///     assert_eq!(
    ///         result,
    ///         Err(Ok(Error::from_type_and_code(
    ///             ScErrorType::Storage,
    ///             ScErrorCode::MissingValue
    ///         )))
```

PR #1663, opened 2026-01-03
Add comprehensive Poseidon/Poseidon2 support for BN254 and BLS12-381

jayz22:
##  Summary

  This PR provides comprehensive Poseidon and Poseidon2 hash function implementations for both BN254 and BLS12-381 curves, with extensive parameter coverage and test validation against external reference implementations.

###  Convenience Hash Methods

  - Added poseidon_hash<N>() - matches https://github.com/iden3/circomlib/blob/35e54ea21da3e8762557234298dbb553c175ea8d/circuits/poseidon.circom
  - Added poseidon2_hash<N>() - matches https://github.com/noir-lang/noir/blob/abfee1f54b20984172ba23482f4af160395cfba5/noir_stdlib/src/hash/poseidon2.nr

###  Poseidon Parameters (poseidon_params.rs)

  - BN254: MDS matrix and round constants for t=2, t=3, t=4 (validated against circomlib)
  - BLS12-381: MDS matrix and round constants for t=2, t=3, t=4 (validated against reference Sage script and [poseidon-bls12381-circom](https://github.com/jmagan/poseidon-bls12381-circom))

###  Poseidon2 Parameters (poseidon2_params.rs)

  - BN254: Diagonal matrix (MAT_DIAG) and round constants for t=2, t=3, t=4
  - BLS12-381: Diagonal matrix and round constants for t=2, t=3, t=4
  - Parameters generated using reference Sage script and validated against reference test vectors (generated by the script)

###  Sponge Implementations

  - PoseidonSponge and Poseidon2Sponge with configurable parameters via PoseidonConfig and Poseidon2Config
  - Proper capacity/rate handling matching reference implementations

###  Test Coverage

- Poseidon (BN254) - hash_n validated against circomlib
-  Poseidon (BLS12-381) - hash_n validated against poseidon-bls12381-circom
-  Poseidon2 (BN254) - hash validated against barretenberg, permutation validated against reference test vectors
-  Poseidon2 (BLS12-381) - permutation validated against reference test vectors


soroban-sdk/src/crypto.rs
```diff
@@ -199,26 +199,17 @@ impl Crypto {
...
-    pub fn poseidon_hash(&self, inputs: &Vec<U256>, field: Symbol) -> U256 {
-        // The initial value for the capacity element initialized with 0 for standard Poseidon
-        let iv = U256::from_u32(&self.env, 0);
-        let mut sponge = PoseidonSponge::new(&self.env, iv, field);
-        for input in inputs.iter() {
-            sponge.absorb(input);
-        }
-        sponge.squeeze()
+    /// Performs a Poseidon hash using a sponge construction that matches circom's [implementation](https://github.com/iden3/circomlib/blob/35e54ea21da3e8762557234298dbb553c175ea8d/circuits/poseidon.circom)
+    pub fn poseidon_hash<const N: usize>(&self, field_type: Symbol, inputs: &[U256; N]) -> U256 {
```
leighmcculloch: The use of the `const N` and an array type look unnecessary. A const N is only useful if the value of N will always or must always be known at compile time. Due to monomorphisastion in Rust this function could be duplicated for every value of N. As far as I can tell, the length of the inputs value can be discovered at runtime. The following appears to work:

```diff
diff --git i/soroban-sdk/src/crypto.rs w/soroban-sdk/src/crypto.rs
index 8b9d956..a285f73 100644
--- i/soroban-sdk/src/crypto.rs
+++ w/soroban-sdk/src/crypto.rs
@@ -200,14 +200,15 @@ impl Crypto {
     }
 
     /// Performs a Poseidon hash using a sponge construction that matches circom's [implementation](https://github.com/iden3/circomlib/blob/35e54ea21da3e8762557234298dbb553c175ea8d/circuits/poseidon.circom)
-    pub fn poseidon_hash<const N: usize>(&self, field_type: Symbol, inputs: &[U256; N]) -> U256 {
-        let config = PoseidonConfig::new(&self.env, field_type, N as u32);
+    pub fn poseidon_hash(&self, field_type: Symbol, inputs: &[U256]) -> U256 {
+        let rate = inputs.len().try_into().unwrap(); // TODO: Handle this error case?
+        let config = PoseidonConfig::new(&self.env, field_type, rate);
         poseidon_sponge::hash(&self.env, inputs, config)
     }
 
     /// Performs a poseidon2 hash with a sponge construction that matches noir's implementation.
     /// Uses rate=3 (t=4) to match the noir/barretenberg [implementation](https://github.com/noir-lang/noir/blob/abfee1f54b20984172ba23482f4af160395cfba5/noir_stdlib/src/hash/poseidon2.nr).
-    pub fn poseidon2_hash<const N: usize>(&self, field_type: Symbol, inputs: &[U256; N]) -> U256 {
+    pub fn poseidon2_hash(&self, field_type: Symbol, inputs: &[U256]) -> U256 {
         let config = Poseidon2Config::new(&self.env, field_type, 3);
         poseidon2_sponge::hash(&self.env, inputs, config)
     }
diff --git i/soroban-sdk/src/crypto/poseidon2_sponge.rs w/soroban-sdk/src/crypto/poseidon2_sponge.rs
index c922c05..67123e6 100644
--- i/soroban-sdk/src/crypto/poseidon2_sponge.rs
+++ w/soroban-sdk/src/crypto/poseidon2_sponge.rs
@@ -135,11 +135,12 @@ impl Poseidon2Sponge {
     }
 }
 
-pub fn hash<const N: usize>(env: &Env, inputs: &[U256; N], config: Poseidon2Config) -> U256 {
+pub fn hash(env: &Env, inputs: &[U256], config: Poseidon2Config) -> U256 {
     // The initial value for the capacity element initialized with `input.len() * 2^64` for Poseidon2
-    let iv = U256::from_u128(env, (N as u128) << 64);
+    let inputs_len_u128: u128 = inputs.len().try_into().unwrap(); // TODO: Handle this error case?
+    let iv = U256::from_u128(env, inputs_len_u128 << 64);
     let mut sponge = Poseidon2Sponge::new(env, iv, config);
-    let input_vec = Vec::from_array(env, inputs.clone());
+    let input_vec = Vec::from_slice(env, inputs);
     sponge.absorb(&input_vec);
     sponge.squeeze()
 }
diff --git i/soroban-sdk/src/crypto/poseidon_sponge.rs w/soroban-sdk/src/crypto/poseidon_sponge.rs
index 62673ad..6f2570d 100644
--- i/soroban-sdk/src/crypto/poseidon_sponge.rs
+++ w/soroban-sdk/src/crypto/poseidon_sponge.rs
@@ -144,11 +144,11 @@ impl PoseidonSponge {
     }
 }
 
-pub fn hash<const N: usize>(env: &Env, inputs: &[U256; N], config: PoseidonConfig) -> U256 {
+pub fn hash(env: &Env, inputs: &[U256], config: PoseidonConfig) -> U256 {
     // The initial value for the capacity element initialized with 0 for standard Poseidon
     let iv = U256::from_u32(env, 0);
     let mut sponge = PoseidonSponge::new(env, iv, config);
-    let input_vec = Vec::from_array(env, inputs.clone());
+    let input_vec = Vec::from_slice(env, inputs);
     sponge.absorb(&input_vec);
     sponge.squeeze()
 }

```

leighmcculloch (commented):
Reviewing the top-level arrangement of the exported fns and that interface, it looks good, there's just one piece of feedback I mention inline to remove the generics that don't appear to be necessary.

Defer to @sisuresh for a review of the crypto internals.

soroban-sdk/src/crypto/poseidon_params.rs
```diff
@@ -1,360 +1,4325 @@
 use crate::{bytesn, vec, Env, Vec, U256};
 pub const T: usize = 3;
```
teddav: `T` is not needed anymore

soroban-sdk/src/crypto/poseidon_sponge.rs
```diff
@@ -1,99 +1,154 @@
...
-            rounds_f.into(),
-            rounds_p.into(),
-            self.mds.to_object(),
-            self.rc.to_object(),
+            self.config.rounds_f.into(),
+            self.config.rounds_p.into(),
+            self.config.mds.to_object(),
+            self.config.rc.to_object(),
         )
         .unwrap_infallible()
```
teddav, line 114: no unwrap please 🙏

soroban-sdk/src/crypto/poseidon_sponge.rs
```diff
@@ -1,99 +1,154 @@
...
+    // field_type - 'BN254' or 'BLS12_381'
+    // rate - number of elements the sponge can absorb/squeeze continously at once, after which a permutation is
+    // required.
+    // capacity - number of extra elements the state contains, capacity provides more safety. capacity = 1 in our simplified sponge (matching circom's).
+    // state.len() == rate + capacity
+    // returns a Poseidon object initialize with the prepulated parameters for use, you can call hash on it.
+    // limitation - single sqeeze, no duplex switching between absorb and sqeeze
+    pub fn new(env: &Env, iv: U256, config: PoseidonConfig) -> Self {
         // the 0-th element is reserved for capacity
         let mut state = vec![env, iv];
```
teddav, line 74: the should initialize the state with `CAPACITY` elements

soroban-sdk/src/crypto.rs
```diff
@@ -199,26 +199,17 @@ impl Crypto {
...
-        let mut sponge = Poseidon2Sponge::new(&self.env, iv, field);
-        for input in inputs.iter() {
-            sponge.absorb(input);
-        }
-        sponge.squeeze()
+    /// Performs a poseidon2 hash with a sponge construction that matches noir's implementation.
+    /// Uses rate=3 (t=4) to match the noir/barretenberg [implementation](https://github.com/noir-lang/noir/blob/abfee1f54b20984172ba23482f4af160395cfba5/noir_stdlib/src/hash/poseidon2.nr).
+    pub fn poseidon2_hash<const N: usize>(&self, field_type: Symbol, inputs: &[U256; N]) -> U256 {
+        let config = Poseidon2Config::new(&self.env, field_type, 3);
+        poseidon2_sponge::hash(&self.env, inputs, config)
```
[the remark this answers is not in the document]
teddav: good catch 😊

soroban-sdk/src/crypto/poseidon_sponge.rs
```diff
@@ -1,99 +1,154 @@
...
 
-    pub fn absorb(&mut self, input: U256) {
+    pub fn absorb(&mut self, inputs: &Vec<U256>) {
         assert!(!self.squeezed);
-        if self.cache.len() == RATE {
-            // If we're absorbing, and the cache is full, apply the sponge permutation to compress the cache
+        let cache_len = self.cache.len();
+        let inputs_len = inputs.len();
+
+        if cache_len + inputs_len > self.config.rate {
```
teddav, line 124: Since in `PoseidonConfig`, the `rate` we're using corresponds to the number of inputs, then this if...else is not needed. All the inputs will be absorbed at once. Or am i missing something?

soroban-sdk/src/crypto.rs
jayz22: Fixed.

soroban-sdk/src/crypto/poseidon_params.rs
jayz22: Removed.

soroban-sdk/src/crypto/poseidon_sponge.rs
jayz22: Because it's a host function call, which is falliable in native but not when called by wasm.

soroban-sdk/src/crypto/poseidon_sponge.rs
jayz22: yeah, in our sponge choice, the capacity is always 1 (we could expose it as a parameter but doesn't feel like necessary). Updated the code comment for better clarity.

soroban-sdk/src/crypto/poseidon_sponge.rs
jayz22: The sponge implementation actually can handle inputs larger than rate (similar to how `poseidon2.nr` does), in which case it will repeatedly absorb and permute until the input is consumed.
In circom's version it only handles 1-pass, which is what our `crypto::poseidon_hash` replicates.

soroban-sdk/src/crypto.rs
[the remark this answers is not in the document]
jayz22: No, the rate=3 is intentional. Updated the comments for clarity.

soroban-sdk/src/crypto/poseidon_sponge.rs
teddav: yes, but in this case, the program will panic before if you have more than the accepted number of inputs.
for example here: https://github.com/jayz22/rs-soroban-sdk/blob/6a7d4a629ee0e9c617745736335db09f76361af8/soroban-sdk/src/crypto/poseidon_params.rs#L29, due to the `unimplemented` macro.
So the user will not be able to hash more than 5 inputs (t=6).

jayz22: As discussed, there are two ways of using Poseidon:
- via `Crypto::poseidon_hash` , this is intended to match circom (`poseidon.circom`) up to t=6. In this case any array length exceeding 5 (maximum t -1) will return unimplemented error, which is correct, since we haven't setup the parameter set for it. If you do it in circom, and pass in an array of size 17, it will also crash (because it has max_t = 17).
- if you want more flexibility, you can initialize your `PoseidonConfig` manually (e.g. with t=3), then call `poseidon_sponge::hash`, with arbitrary number of inputs, this will repeatedly absorb-and-permute the whole array, which is a valid and secure operation, it's just not what circom does. But you have the option to.

Improved the documentations to make this more clear.

soroban-sdk/src/crypto.rs
jayz22: On further consideration (thanks to inputs from @teddav), I've decided to use `Vec` in the hash interface instead of array/slice. 
Initially I went with fixed-length array in the interface to make it more clear, since most use cases should know beforehand how many inputs to hash (like in a MT). However, this is more restrictive (there is no easy Vec->Slice conversion), and provide no cost benefit since we need to construct a Vec for the host function call anyway.

sisuresh (approved):

PR #1667, opened 2026-01-05
Add bn254 support and expose poseidon/poseidon2 host functions

mootz12:
### What

* Add bn254 support (#1615, #1628)
  - A new module `crypto/bn254.rs` with Bn254 point and field types, and expose related host operations.
  - Implement `Neg` trait for `G1Affine` (#1630)
  - Update bn254 typenames and add mapping (#1645), also updated bls12-381 type names for consistency
* Add raw Poseidon, Poseidon2 support in `CryptoHazmat`, gated by "hazmat-crypto" feature

### Why

Expose new crypto functions for protocol 25

### Known limitations

Only the raw host function exposure for Poseidon/Poseidon2 has been added in this PR. The more elaborate setup for sponge, parameters, hash function wrappers will be done in a separate repo [here](https://github.com/stellar/rs-soroban-poseidon)


leighmcculloch:
@jayz22 Should this wait on https://github.com/stellar/rs-soroban-sdk/pull/1663 merging first? Or can that go to main separately?

soroban-sdk/src/crypto/bn254.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// # Returns:
+    /// - `true` if the pairing check holds (i.e., the product of pairings equals 1),
+    ///   otherwise `false`.
+    ///
+    /// # Panics:
+    /// - If the lengths of `vp1` and `vp2` are not equal or if they are empty.
+    pub fn pairing_check(&self, vp1: Vec<Bn254G1Affine>, vp2: Vec<Bn254G2Affine>) -> bool {
+        let env = self.env();
+        internal::Env::bn254_multi_pairing_check(env, vp1.into(), vp2.into())
```
teddav, line 333: why unwrap? can we return a Result<bool> instead? and have a proper error message to know what went wrong?

soroban-sdk/src/crypto/poseidon2_params.rs
```diff
@@ -0,0 +1,10483 @@
...
+                        e,
+                        0x0000000000000000000000000000000000000000000000000000000000000000
+                    )
+                    .into(),
+                ),
+                U256::from_be_bytes(
+                    e,
+                    &bytesn!(
+                        e,
+                        0x0000000000000000000000
```
leighmcculloch: Wow, this makes a lot of host fn calls. @jayz22 This works within the limits?

soroban-sdk/src/crypto/poseidon2_sponge.rs
```diff
@@ -0,0 +1,155 @@
...
+    rounds_f: u32,
+    rounds_p: u32,
+    m_diag: Vec<U256>,
+    rc: Vec<Vec<U256>>,
+}
+
+impl Poseidon2Config {
+    pub fn new(env: &Env, field_type: Symbol, rate: u32) -> Self {
+        let t = rate + CAPACITY;
+        if field_type == soroban_sdk_macros::internal_symbol_short!("BN254") {
```
leighmcculloch: I'm wondering what the motivation is for using a symbol here, instead of a Rust enum?

soroban-sdk/src/crypto/poseidon2_params.rs
```diff
@@ -0,0 +1,10483 @@
...
+            U256::from_be_bytes(e, &bytesn!(e, 0x00544b8338791518b2c7645a50392798b21f75bb60e3596170067d00141cac15).into()),
+            U256::from_be_bytes(e, &bytesn!(e, 0x222c01175718386f2e2e82eb122789e352e105a3b8fa852613bc534433ee428b).into()),
+        ]
+    }
+    else {
+        unimplemented!("Poseidon2 preset parameters only support t <= 4. You can supply your own parameters and call CryptoHazmat::poseidon2_permutation")
+    }
+}
+
+pub(crate) fn get_rc_bn254(e: &Env, t: u32) -> Vec<Vec<U256>> {
```
leighmcculloch: The other big fns use `#[rustfmt::skip]` to format it easier to work with, but this fn seems to be formatted differently. Could we keep them consistent?

soroban-sdk/src/crypto/poseidon2_params.rs
```diff
@@ -0,0 +1,10483 @@
...
+    match t {
+        2 | 3 | 4 => 56,
+        _ => unimplemented!("Poseidon2 preset parameters only support t = 2, 3, 4. You can supply your own parameters and call CryptoHazmat::poseidon2_permutation"),
+    }
+}
+
+// These parameters matches reference parameters generated with the reference sage script (t = 4, field = BN254):
+// https://github.com/HorizenLabs/poseidon2/blob/055bde3f4782731ba5f5ce5888a440a94327eaf3/poseidon2_rust_params.sage
+#[rustfmt::skip]
+pub(crate) fn get_mat_diag_m_1_bn254(e: &Env, t: u32) -> Vec<U256> {
```
leighmcculloch: Could it make sense to use an enum for `t`? It looks like it's operating like an enum more than an integer. Or is that not convention?

soroban-sdk/src/crypto/utils.rs
```diff
@@ -0,0 +1,60 @@
+use crate::BytesN;
+
+// This routine was copied with slight modification from the arkworks library:
+// https://github.com/arkworks-rs/algebra/blob/bf1c9b22b30325ef4df4f701dedcb6dea904c587/ff/src/biginteger/arithmetic.rs#L66-L79
```
leighmcculloch, line 4: The citation.cff file from that repository requests that we credit use with the following, and we should mention the license 

```suggestion
// https://github.com/arkworks-rs/algebra/blob/bf1c9b22b30325ef4df4f701dedcb6dea904c587/ff/src/biginteger/arithmetic.rs#L66-L79
// Copyright 2022 arkworks contributors. arkworks zkSNARK ecosystem [Computer software]. https://github.com/arkworks-rs/
// Licensed under the Apache License, Version 2.0
```

soroban-sdk/src/crypto/poseidon_sponge.rs
```diff
@@ -0,0 +1,163 @@
...
+        // zero-pad the cache
+        for _ in self.cache.len()..self.config.rate {
+            self.cache.push_back(U256::from_u32(&self.env, 0));
+        }
+        // add the cache into sponge state, leaving the 0-th element
+        for i in 0..self.config.rate {
+            let elem = self
+                .state
+                .get_unchecked(i + CAPACITY)
+                .add(&self.cache.get_unchecked(i));
```
sisuresh: I had Claude review the poseidon code, and it noticed an issue here that I discussed with @jayz22 - this isn't performing the prime modulus on each add, so it can overflow U256 on inputs that could be valid.

jayz22: Good catch. This has code been refactored and removed out of this PR.

soroban-sdk/src/crypto/poseidon2_params.rs
jayz22: Refactored and moved out of this PR.

soroban-sdk/src/crypto/poseidon2_sponge.rs
jayz22: Refactored and moved out of this PR.

soroban-sdk/src/crypto/poseidon2_params.rs
jayz22: Refactored and moved out of this PR.

soroban-sdk/src/crypto/poseidon2_params.rs
jayz22: Refactored and moved out of this PR.

soroban-sdk/src/crypto/utils.rs
jayz22: Fixed.

soroban-sdk/src/crypto/bn254.rs
jayz22: @teddav by default every host error registers a `DiagnosticEvent` in the host (most of them has a more detailed message, such as the value that was wrong), which gives sufficient info for debugging. Every internal host error traps the contract, which manifests in a panic. This is both a restriction of Wasm 1.0 standard (returning a single value) and our design choice in Soroban. 
This particular function `bn254_multi_pairing_check` also returns a `false` if the pairing check fails, which is the one expected error for which the user can take action on. All other errors are related to malformed inputs (vector lengths mismatch) or host internal errors (e.g. budget limit exceed).

PR #1672, opened 2026-01-09
Implement spec shaking using dead code elimination and data section of wasm

### LINKED ISSUE #1330: Support referencing and re-exporting types from other libs for type sharing

### LINKED ISSUE #1569: Exports inconsistently leaking from libraries into contracts
_This issue is a lightweight placeholder and needs more details._

@brozorec showed me today a contract build where a contract using the open zeppelin stellar contracts is picking up some random exports from the libraries.

He shared an example here: https://github.com/OpenZeppelin/stellar-contracts/pull/373

This new behaviour has been noticed as part of upgrading both to a new sdk as well as new contract code, so there's a few moving parts and it is possible this is a result of one of those moving parts. But after a cursory look I think this might be a case of exports leaking from libraries to the contract, but in a way that's inconsistent.

We know that control over what exports leak from libraries to the main cdylib being built is very limited in the rust compiler. This is an issue we have explored in:
- https://github.com/stellar/rs-soroban-sdk/issues/1439

### LINKED ISSUE #2030: `stellar contract build` should strip contract specs, removing types that are not referenced by functions or events
We have an issue in the soroban-sdk that any exported `contracttypes` defined in the SDK, or any library for that matter, that don't get used in the importing contract still end up in the contract spec.

The problem results in some awkward behaviour where we avoid exporting contracttypes in the SDK, even though doing so would be convenient and enable using contracttypes more frequently there.

Because of how proc-macros run in isolation and cannot coordinate in wasm builds in Rust (the ctor crate cannot be used like we use it in non-wasm builds in tests to coordinate), there's nothing the soroban-sdk can do to identify whether a contracttype that sometimes needs to be exported hasn't been used and can be excluded.

The stellar-cli contract build command though does have an opportunity to help here. It can inspect the contract specs and identify unused types that are using up space for now reason. It could either do this widely for any type that is unused, or do it only for types exported from the soroban sdk, because types can specify what library they came from.

This will require some care because it will require rewriting the wasm file after build, which is something the stellar-cli doesn't currently do. Currently the stellar-cli does modify the wasm, but only in an append fashion.

leighmcculloch:
### What

Implement spec shaking using dead code elimination and data section of wasm. Embed markers in the WASM data section for each type/event that is used at external boundaries. These markers enable post-build tools to filter unused entries from the `contractspecv0` section.

Also, all contracttypes, contracterrors, and contractevents are exported in contract specs now if they are directly or indirectly referenced by a function parameter, return value, event.

### Why

Contract WASM files can contain unused type and event definitions in the `contractspecv0` custom section that inflate binary size unnecessarily.

No shaking of specs occurrs today.

Because of this contract imports never export types otherwise importing one contract would export all its types.

Because of how proc-macros run in isolation and cannot coordinate in WASM builds in Rust, there's nothing the soroban-sdk can do to identify whether a contracttype that sometimes needs to be exported hasn't been used and can be excluded.

The macros can however embed markers in the regular data section for each type/event when it's used. These markers survive DCE only if the corresponding code is reachable. A post-build tool (stellar-cli) can extract these markers and filter the spec accordingly.

Close #1569
Close https://github.com/stellar/stellar-cli/issues/2030

### Side-effects

A side-effect of this change, is that types/events brought into a contract from a `contractimport` can now be automatically exported from that contract if they use the types at their boundary. This is why the change also removes the `export = false` from types/events that come from a `contractimport`.

Close #1330

### How

The SDK embeds markers using volatile reads in conversion/publish functions:

**Marker format (12 bytes):**
- Bytes 0-5: `SpEcV1` magic prefix
- Bytes 6-13: 64-bit truncated SHA256 hash of the spec entry XDR bytes

**Embedding mechanism:**
```rust
impl SpecShakingMarker for MyType {
    #[inline(always)]
    fn spec_shaking_marker() {
        {
            static MARKER: [u8; 14] = *b"SpEcV1\x93\xd5\x05\x04\x36\x68\x40\xd2";
            let _ = unsafe { ::core::ptr::read_volatile(MARKER.as_ptr()) };
        }
    }
}
```

**When markers are included:**
- Types: `spec_shaking_marker()` is called from `TryFromVal` implementations via `IncludeSpec` trait
- Events: `spec_shaking_marker()` is called from the `publish()` method

**Why this works:** If a type/event is never used at an external boundary, its `__include_spec_marker()` function is never called, and DCE removes both the function and its marker. If the type is used, the marker survives in the data section.

**Type identities:** Note that a long time problem with doing any sort of work like this has been identifying the identity of one spec type vs another. That is improved and worked around in this change by using a hash of the spec entry rather than the name of the spec entry.

### CLI Support Required

This feature requires CLI support for filtering specs using these markers. See: https://github.com/stellar/stellar-cli/pull/2353

### Try It Out

Install the modified `stellar-cli` that strips unused specs:
```
cargo install --locked --git https://github.com/stellar/stellar-cli --branch spec-clean stellar-cli
```

Import the modified `soroban-sdk` that marks used specs:
```
[dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "spec-markers", features = ["experimental_spec_shaking_v2"] }

[dev-dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "spec-markers", features = ["experimental_spec_shaking_v2", "testutils"] }"
```

Build using the stellar-cli:
```
stellar contract build
```

Inspect the contract interface and it should only show types actually used at the boundary of the contract (inputs, outputs, and events):
```
stellar contract info interface --wasm target/wasm32v1-none/release/contract.wasm
```

### Todo
- [x] Figure out a roll out plan, because you can't really use this sdk until you use the new cli, otherwise your specs will increase in size rather than decrease because imported types are now exported then stripped.
- [x] Evaluate the cost of the volatile read on a large contract (expect impact is small)
- [x] Look for alternative ways other than a volatile read to force the marker to stick around
   - [x] Try black box
   - [x] Try `#[used]`
- [x] Markers need to be included for any nested user defined types and types inside containers
- [x] Come up with better name for feature
- [x] Come up with better name for env var (should it just be Stellar CLI version?)
- [x] Write out every possible test case, with dimensions:
	- [x] type: struct, struct tuple, enum, enum int, error
	- [x] containers: vec, map, option, result, tuple (0-12), struct, struct tuple, enum
	- [x] mod: lib, sub
	- [x] visibility: pub, pub(crate), private
	- [x] imported: and not exported, and exported
	- [x] context: contract fn param, non-contract fn param, contract fn return, non-contract fn return, event topic field, event data field


soroban-sdk-macros/src/spec_marker.rs
```diff
@@ -0,0 +1,58 @@
...
+/// Markers are embedded in conversion/usage functions with a volatile read.
+/// When the type is used, the function is called and the marker is included.
+/// When the type is unused, the function is DCE'd along with its marker.
+///
+/// Post-processing tools can:
+/// 1. Scan the WASM data section for "SpEc" patterns
+/// 2. Extract the hash from each marker
+/// 3. Match against specs in contractspecv0 section (by hashing each spec)
+/// 4. Strip unused specs from contractspecv0
+// TODO: Move the spec marker logic into a crate that can be shared with the CLI.
```
mootz12: Is this needed yet?

soroban-spec/src/shaking.rs
```diff
@@ -0,0 +1,58 @@
+/// Spec marker generation for identifying used spec entries.
+///
+/// The marker is a byte array in the data section with a distinctive pattern:
+/// - 4 bytes: "SpEc" prefix
+/// - 8 bytes: first 64 bits of SHA256 hash of the spec entry XDR
```
mootz12, line 19: what was the reason to just use the first 8 bytes? It's probably specific enough, but this is just incase some1 doesn't use a post processing tool like stellar-cli it isn't as impactful?

soroban-sdk-macros/src/spec_marker.rs
```diff
@@ -0,0 +1,58 @@
...
+        hash[7],
+    ]
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_spec_marker() {
```
mootz12: (picky) can we add a test here to assert against a fixed spec_marker? It should fail if the structure of spec_market changes, whereas this one won't as long as the first 4 bytes and len don't change.

soroban-spec/src/marker.rs
```diff
@@ -0,0 +1,58 @@
+/// Spec marker generation for identifying used spec entries.
+///
+/// The marker is a byte array in the data section with a distinctive pattern:
+/// - 4 bytes: "SpEc" prefix
```
mootz12: is it worthwhile to version tag this at all?

`SpEcV1`

In the event this ever changes, would we need to maintain backwards compatibility? Or is it more of a "use the correct CLI version for your given SDK version"

soroban-sdk-macros/src/derive_event.rs
```diff
@@ -193,6 +197,20 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
         "__SPEC_XDR_EVENT_{}",
         input.ident.to_string().to_uppercase()
     );
+    // Create a marker that identifies this spec entry. The marker is a byte array
+    // in the data section with a distinctive pattern: "SpEc" + truncated SHA256.
+    // Post-build toOls can scan the data section for "SpEc" markers and match
```
mootz12:
```suggestion
    // Post-build tools can scan the data section for "SpEc" markers and match
```

soroban-sdk-macros/src/derive_struct_tuple.rs
```diff
@@ -89,10 +96,44 @@ pub fn derive_type_struct_tuple(
...
+                    {
+                        // Marker in data section. Post-build tools can scan for "SpEc"
+                        // patterns and match against specs in contractspecv0.
+                        static MARKER: [u8; #marker_len] = *#marker_lit;
+                        // Volatile read prevents DCE within live function.
+                        let _ = unsafe { ::core::ptr::read_volatile(MARKER.as_ptr()) };
+                    }
+                }
+            }
+        })
```
mootz12: Could we extract this logic out of the `derive_*` files to reduce some code duplication?

soroban-sdk-macros/src/derive_struct_tuple.rs
```diff
@@ -89,10 +96,44 @@ pub fn derive_type_struct_tuple(
...
+                fn include_spec_marker() {
+                    // Include markers for nested field types.
+                    #(<#field_types as #path::IncludeSpecMarker>::include_spec_marker();)*
+                    #[cfg(target_family = "wasm")]
+                    {
+                        // Marker in data section. Post-build tools can scan for "SpEc"
+                        // patterns and match against specs in contractspecv0.
+                        static MARKER: [u8; #marker_len] = *#marker_lit;
+                        // Volatile read prevents DCE within live function.
+                        let _ = unsafe { ::core::ptr::read_volatile(MARKER.as_ptr()) };
```
mootz12: did `::core::hint::black_box(&MARKER);` not work?

soroban-sdk/src/into_val_for_contract_fn.rs
```diff
@@ -0,0 +1,32 @@
...
+//! IntoVal<Env, Val>.
+//!
+//! This trait exists to call `IncludeSpecMarker::include_spec_marker()` to ensure that type
+//! specs are included in the WASM when types are used at external boundaries
+//! (function return values).
+
+use crate::{Env, IncludeSpecMarker, IntoVal, Val};
+
+#[doc(hidden)]
+#[deprecated(
```
mootz12, line 16: is `doc(hidden)` not enough?

mootz12 (commented):
Looks good overall. Leaving review as comment for now as I haven't looked at the CLI side or done much testing.

One thing to note is this might have some impacts if the `spec-marker` ever changes to release complexity between the CLI and SDK.

Might not be anything we can do about it :/

soroban-spec/src/shaking.rs
leighmcculloch: Yeah exactly. And because uniqueness doesn't need to be universal, just across the fns present in a single build, so the collision likelihood is so much less.

soroban-spec/src/marker.rs
leighmcculloch: We could add this. The SDK will also have the SDK version embedded inside the meta custom section and I was planning to update the stellar-cli, so I think it can work out from that version what it should do.

soroban-sdk-macros/src/derive_struct_tuple.rs
leighmcculloch: I tried `#[used]`, `black_box` and the volatile read.

- `#[used]` always included the marker even when the code wasn't called.

- `black_box` never included the marker.

- `read_volatile` includes the marker one time when the code path is called and not when it isn't.

soroban-sdk/src/into_val_for_contract_fn.rs
leighmcculloch: This is a pattern we've used in the sdk to discourage folks from accidentally calling internal apis. It gives us an opportunity to provide a message about why the api shouldn't be used. While doc hidden is good for keeping the SDK docs focused on the public api, the downside is that when someone uses a fn they found, if they go look for docs about it, they won't find anything and so there's nothing to read and tell them "don't use this."

soroban-spec/src/marker.rs
leighmcculloch: 7708b537b66250d3873d638a4b1221a4968852b3

soroban-spec/src/shaking.rs
```diff
@@ -0,0 +1,448 @@
...
+            outputs: VecM::default(),
+        });
+
+        let marker = generate_marker_for_entry(&entry);
+
+        // Marker should be 12 bytes
+        assert_eq!(marker.len(), LEN);
+
+        // First 4 bytes should be magic
+        assert_eq!(&marker[..6], MAGIC.as_slice());
```
mootz12: [picky] test comments are stale from prev impl

soroban-spec/src/shaking.rs
```diff
@@ -0,0 +1,448 @@
...
+
+    #[test]
+    fn test_generate_marker_for_entry_struct() {
+        let entry = make_struct("MyStruct", vec![("field", ScSpecTypeDef::U32)]);
+        let marker = generate_marker_for_entry(&entry);
+
+        // Marker should be 12 bytes
+        assert_eq!(marker.len(), LEN);
+
+        // First 4 bytes should be magic
```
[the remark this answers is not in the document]
mootz12: :ditto:

soroban-sdk/Cargo.toml
```diff
@@ -63,6 +63,7 @@ sha2 = "0.10.7"
 [features]
 alloc = []
 testutils = ["soroban-sdk-macros/testutils", "soroban-env-host/testutils", "soroban-ledger-snapshot/testutils", "dep:ed25519-dalek", "dep:arbitrary", "dep:derive_arbitrary", "dep:ctor", "dep:soroban-ledger-snapshot"]
+experimental_spec_shaking_v2 = ["soroban-sdk-macros/experimental_spec_shaking_v2"]
```
mootz12, line 66: [picky] Is there a reason we are calling this `*_v2`? I find it slightly confusing given there was never a `v1` outside of a draft PR.

soroban-spec/src/shaking.rs
```diff
@@ -0,0 +1,448 @@
...
+/// # Arguments
+///
+/// * `entries` - The spec entries to filter
+/// * `markers` - Markers extracted from the WASM data section
+///
+/// # Returns
+///
+/// Filtered entries with only used types/events remaining.
+#[allow(clippy::implicit_hasher)]
+pub fn filter(entries: Vec<ScSpecEntry>, markers: &HashSet<Marker>) -> Vec<ScSpecEntry> {
```
mootz12: [picky] `&[ScSpecEntry]`?

mootz12 (commented):
Looks great! Left a couple pedantic comments  inline.

Leaving review as comment for now - will revise after doing some manual testing.

soroban-sdk/Cargo.toml
leighmcculloch: I named it v2 because the sdk sort of does a v1 behaviour for excluding specs today, but it is super primitive by never exporting types imported from wasm. But I think the name is confusing. I'm going to rename the feature to dead spec elimination (dse) which is more accurate, and conveniently aligns with the Rust feature, dead code elimination, that it actually uses to achieve the goal.

leighmcculloch: Actually I just walked that change back. Why: when I renamed the feature to dead code elimination the feature is misleading because it isn't only about dead code elimination and turns on exporting within contractimport. I think I prefer the v1 vs v2 naming because it implies there's behaviour today that's captured in v1 and an entirely new approach v2.

Does that make sense?

soroban-spec/src/shaking.rs
leighmcculloch: Changed to use iterators: 5af448c6444fb799f39c24fadd6483e797713cb5

soroban-spec/src/shaking.rs
leighmcculloch: 6121af0ffd6f58cca4fe8079b10dd6d69c7b7c90

soroban-spec/src/shaking.rs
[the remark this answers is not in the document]
leighmcculloch: 6121af0ffd6f58cca4fe8079b10dd6d69c7b7c90

soroban-sdk/Cargo.toml
leighmcculloch: 5cd7f3b37429cc07775f28d2ab2f1180016fa1b4

graydon (approved):
I was asked to confirm the "volatile loads don't get DCE'd" part of this project and as far as I can tell, this is correct. Investigation is here: https://gist.github.com/graydon/d0377a5779ec278555cda4fe54d1beaf

soroban-sdk/src/lib.rs
```diff
@@ -1057,7 +1104,7 @@ pub use env::FromVal;
 pub use env::IntoVal;
 /// Used to do conversions between values in the Soroban environment.
 pub use env::TryFromVal;
-/// Used to do conversions between values in the Soroban environment.
+/// Used to do convers_ons between values in the Soroban environment.
```
mootz12: typo

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         working-directory: ${{ fromJSON(needs.collect-crates.outputs.dirs) }}
+        experimental_spec_shaking_v2: [true, false]
```
mootz12, line 39: [question] Are we concerned about the number of CI jobs this is going to spawn?

OZ can _probably_ be done in one job

mootz12 (approved):
LG2M. One minor comment and a general question that doesn't impact merging this.

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
leighmcculloch: That's fine, jobs are cheap, we use this pattern elsewhere and we are way under the limit where this becomes problematic.

soroban-sdk/src/lib.rs
leighmcculloch:
```suggestion
/// Used to do conversions between values in the Soroban environment.
```

leighmcculloch: 73c3ba599c5b60f2d3668f08588989b769be624e

PR #1674, opened 2026-01-09
Remove Poseidon, Poseidon2

jayz22:
### What

Remove Poseidon, Poseidon2 code from the sdk.

### Why

To prepare for migrating this into a separate repo, for independent review/audit/release cadence from the regular sdk.

### DIFF
--- a/soroban-sdk/src/crypto.rs
+++ b/soroban-sdk/src/crypto.rs
@@ -3,7 +3,7 @@
 use crate::{
     env::internal::{self, BytesObject},
     unwrap::UnwrapInfallible,
-    Bytes, BytesN, ConversionError, Env, IntoVal, TryFromVal, Val,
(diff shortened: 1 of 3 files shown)


soroban-sdk/src/crypto.rs
```diff
@@ -311,68 +264,4 @@ impl CryptoHazmat {
...
-            d.into(),
-            rounds_f.into(),
-            rounds_p.into(),
-            mat_internal_diag_m_1.to_object(),
-            round_constants.to_object(),
-        )
-        .unwrap_infallible();
-
-        result.try_into_val(env).unwrap_infallible()
-    }
```
mootz12, line 377: Should we keep these? These look like they are directly invoking the host functions, which are normally exposed through the SDK to contracts should they want them.

jayz22: Yeah good point. I added the host function wrappers back to the crypto_hazmat, with some simple sanity test against the host. We can drop them later if they are not useful.

mootz12 (approved):

PR #1677, opened 2026-01-14
Add test resource limit enforcement functionality.

dmkozh:
### What

Add test resource limit enforcement functionality. Also enable mainnet limit enforcement by default. These shouldn't be disruptive to most of the existing users (as they're likely able to run their contracts in testnet or mainnet), and for those in the experimentation stage this might be useful for early discovery of the issues.

The limits enforced can be customized or enforcement can be disabled altogether.

### Why

Resource limit enforcement provides early feedback for the contracts that might be too heavy to be usable after deployment. The resource estimation is not perfect, but usually the test environment will have under-estimated resources (e.g. due to not using the compiled Wasms), i.e. there shouldn't be annoying 'false positives' for this. 

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/src/_migrating.rs
+++ b/soroban-sdk/src/_migrating.rs
@@ -15,7 +15,14 @@
 //!    Define traits with default implementations using `#[contracttrait]`, then implement them
 //!    in contracts using `#[contractimpl(contracttrait)]`.
 //!
+//! 5. [Resource limit enforcement enabled by default in tests][v25_resource_limits].
+//!    `Env::default()` now enforces Mainnet resource limits for contract invocations.
+//!    Tests will fail if limits are exceeded. This provides early warning of contracts that
+//!    may be too resource-heavy for Mainnet. If you see test failures after upgrading,
+//!    use `env.cost_estimate().disable_resource_limits()` to opt-out while optimizing.
+//!
 //! [v25_contracttrait]: v25_contracttrait
+//! [v25_resource_limits]: v25_resource_limits
 //!
 //! # Migrating from v22 to v23
 //!
@@ -272,3 +279,4 @@ pub mod v25_bn254;
 pub mod v25_contracttrait;
 pub mod v25_event_testing;
 pub mod v25_poseidon;
+pub mod v25_resource_limits;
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/testutils/cost_estimate.rs
```diff
@@ -89,4 +91,65 @@ impl CostEstimate {
...
+    }
+}
+
+
+/// Predefined network invocation resource limits.
+pub trait NetworkInvocationResourcesLimits {
+    fn mainnet() -> Self;
+}
+
+impl NetworkInvocationResourcesLimits for InvocationResourceLimits {
```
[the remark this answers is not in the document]
dmkozh: Fixed.

soroban-sdk/src/tests/cost_estimate.rs
```diff
@@ -161,17 +161,17 @@ fn test_cost_estimate_budget() {
     // Budget breakdown corresponds to the last invocation only.
     expect![[r#"
         ===============================================================================================================================================================================
-        Cpu limit: 100000000; used: 242479
-        Mem limit: 41943040; used: 1126705
+        Cpu limit: 100000000; used: 257503
```
dmkozh, line 164: This change is semi-intentional: the env PR happened to use metered XDR conversions in the limit enforcement, but that's actually a good thing for the purpose of limit enforcement, because it reflects some XDR roundtrips that exist in e2e execution, but SDK doesn't do. So this improves the estimate accuracy, albeit in a pretty insignificant fashion.

soroban-sdk/src/testutils/cost_estimate.rs
```diff
@@ -89,4 +91,63 @@ impl CostEstimate {
...
+            .host()
+            .set_invocation_resource_limits(None)
+            .unwrap();
+    }
+}
+
+/// Predefined network invocation resource limits.
+pub trait NetworkInvocationResourceLimits {
+    fn mainnet() -> Self;
+}
```
jayz22, line 131: Nit: why does this need to be a trait? (Instead directly adding `mainnet()` to `InvocationResourceLimits`). Do you plan to extend this to a different type (like a different mainnet setting for testing or something)?

dmkozh: >  (Instead directly adding mainnet() to InvocationResourceLimits).

That would need to happen in env (as it's not possible to add impl for struct in a different crate). I'd like to avoid hardcoding configs in env as it's much easier to make a quick SDK release with the updated numbers.

jayz22 (approved):

PR #1678, opened 2026-01-15
Add MuxedAddress::from_str and from_string methods

leighmcculloch:
### What
Add `from_str` and `from_string` methods to `MuxedAddress` for parsing Stellar strkey addresses (G.../M.../C...) directly in contracts. Uses a no-alloc branch of stellar-strkey to enable strkey parsing in the Wasm target without heap allocation.

### Why
Contracts need to parse strkey strings into addresses, particularly for cross-chain scenarios where addresses are shared between different environments. The standard stellar-strkey crate requires allocation, so this uses a no-alloc fork compatible with `#![no_std]` Wasm builds.

Before this is merged the stellar-strkey crate will be released.

This functionality is a temporary holdover until:
- https://github.com/stellar/stellar-protocol/discussions/1840

### Dependencies

- https://github.com/stellar/rs-stellar-strkey/pull/99

### Try it out

```toml
[dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "muxedaddress-fromstr" }

[dev-dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "muxedaddress-fromstr", features = ["testutils"] }
```

### Status
Targeting `main` after the stellar-strkey crate is released and version is updated here.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -99,11 +99,29 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - run: rustup update
-    - uses: stellar/binaries@v45
+    - run: rustup update stable
+    - name: Determine Rust n-2 version since cargo-semver-checks is rarely available for the latest version
+      id: rust-version
+      run: |
+        current_version=$(rustc +stable --version | grep -oE '[0-9]+\.[0-9]+' | head -1)
+        major=$(echo $current_version | cut -d. -f1)
+        minor=$(echo $current_version | cut -d. -f2)
+        prev_minor=$((minor - 2))
+        prev_version="${major}.${prev_minor}"
+        echo "Latest stable: $current_version, using n-2: $prev_version"
+        echo "version=$prev_version" >> $GITHUB_OUTPUT
+    - name: Install Rust ${{ steps.rust-version.outputs.version }}
+      run: |
+        rustup install ${{ steps.rust-version.outputs.version }}
+        rustup override set ${{ steps.rust-version.outputs.version }}
+    - name: Verify Rust version
+      run: |
+        rustc --version
+        cargo --version
+    - uses: stellar/binaries@v51
       with:
         name: cargo-semver-checks
-        version: 0.44.0
+        version: 0.45.0
     - run: >
         cargo semver-checks
         --exclude soroban-meta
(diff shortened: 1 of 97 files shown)


deny.toml
```diff
@@ -244,6 +244,7 @@ deny = [
 # Certain crates/versions that will be skipped when doing duplicate detection.
 skip = [
     { name = "hashbrown", version = "=0.13.2" },
+    { name = "stellar-strkey", version = "=0.0.13" },
```
mootz12, line 247: is there not any concern that the env uses an older version?

soroban-sdk/src/muxed_address.rs
```diff
@@ -181,6 +181,103 @@ impl From<&Address> for MuxedAddress {
...
+        // MuxedEd25519Account: { id: uint64, ed25519: uint256 }
+        // Total: 48 bytes
+        const SCVAL_ADDRESS: i32 = ScValType::Address as i32;
+        const SCADDRESS_MUXED_ACCOUNT: i32 = ScAddressType::MuxedAccount as i32;
+        let mut buf = [0u8; 48];
+        buf[0..4].copy_from_slice(&SCVAL_ADDRESS.to_be_bytes());
+        buf[4..8].copy_from_slice(&SCADDRESS_MUXED_ACCOUNT.to_be_bytes());
+        buf[8..16].copy_from_slice(&muxed.id.to_be_bytes()); // 8-byte mux id (big-endian)
+        buf[16..48].copy_from_slice(&muxed.ed25519); // 32-byte ed25519 public key
+        let xdr_bytes = Bytes::from_slice(env, &buf);
```
mootz12, line 276: This feels like it would be better located within the strkey crate. But, don't think it's necessary to hold things up to fix now.

leighmcculloch: The Rust XDR crate depends on the strkey crate. It uses the strkey crate to present friendly string versions of some XDR types. For that reason the strkey crate can't be dependent on the xdr crate. In any case the strkey crate doesn't really have a responsibility to know how to produce XDR. Ideally we'd be using the stellar-xdr crate here to encode it, but at the moment it doesn't support alloc, and even if it did, a narrowly scoped encoding here keeps the overhead low.

deny.toml
leighmcculloch: We can import both, they will be used independently. In terms of preventing divergence we have a fuzz test in the strkey crate that tests itself with v13 which is the env/core version in use at the moment. So far no differences have showed up.

Issue #1679, opened 2026-01-15
`contractevent` macro orders data alphabetically for vec data format


mootz12:
The `contractevent` macro orders the entries in the vec alphabetically by the struct field name.

This is not an obvious outcome, and could result in users emitting events in the wrong order. With the additional test helpers for creating events, this might be missed entirely.

Example - if you re-order the vec example in the docs:
```rs
// Event topics can also be noted in the struct, so they're dynamic.
// This event will be published with the following structure:
// `["deposit", addr: Address, token: Address], data = [time: u64, amount: i128]`
#[contractevent(data_format = "vec")]
pub struct Deposit {
    #[topic]
    addr: Address,
    #[topic]
    token: Address,
    time: u64,
    amount: i128,
} 
```

The output will still be `data = [amount: i128, time: u64]`.

The expected output would be `data = [time: u64, amount: i128]`

leighmcculloch:
I reviewed the SEP-41 token events and the only event with a vec data format is the approve event, which has its fields ordered to match in any case: https://github.com/stellar/rs-soroban-sdk/blob/7041ac2e5bb6e99da9b21f4cb7963d307bcda886/soroban-token-sdk/src/events.rs#L3-L11

And same with the SAC events: https://github.com/stellar/stellar-asset-contract-spec/blob/main/stellar-asset-spec.json#L299-L323

leighmcculloch:
The bug is only present in the publishing of the event, not in the order of the fields in the spec encoding into the contract:
 - Spec generation: Uses params in declaration order - spec is correct
 - Data writing: Sorted alphabetically - data was wrong

 For the example in the issue, this means:
 - Spec says: `data = [time: u64, amount: i128]`
 - Actual data published: `data = [amount: i128, time: u64]`

This discrepancy makes it possible to identify affected contracts by inspecting their specs and looking for specs with events where the fields would change order when sorted by their name.

This script finds impacted contracts:
https://github.com/leighmcculloch/contract-wasms/blob/main/.scripts/find_event_1679.sh

This csv lists impacted contracts (data is best effort based on visibility in last checkpoint archive):
https://github.com/leighmcculloch/contract-wasms/blob/main/analysis/find_event_1679_results.csv

leighmcculloch:
Taking discussion on impact to:
- https://github.com/stellar/stellar-protocol/discussions/1848

PR #1680, opened 2026-01-16
Fix contractevent vec data format to preserve field declaration order

mootz12:
### What

Only sort contractevent fields when building events where the data format is `map`.

### Why

For contractevents with a data_format of vec, sorting the fields caused the data to be published in a different order than what the event defined, which is not expected.

Closes https://github.com/stellar/rs-soroban-sdk/issues/1679

### Known limitations

N/A

### DIFF
--- a/soroban-sdk-macros/src/derive_event.rs
+++ b/soroban-sdk-macros/src/derive_event.rs
@@ -229,16 +229,11 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
     let data_params = params
         .iter()
         .filter(|p| p.location == ScSpecEventParamLocationV0::Data)
-        .sorted_by_key(|p| p.name.to_string()) // must be sorted for map_new_from_slices
         .collect::<Vec<_>>();
     let data_params_count = data_params.len();
     let data_idents = data_params
         .iter()
         .map(|p| format_ident!("{}", p.name.to_string()))
-        .collect::<Vec<_>>();
-    let data_strs = data_idents
-        .iter()
-        .map(|i| i.to_string())
         .collect::<Vec<_>>();
     let data_to_val = match args.data_format {
         DataFormat::SingleValue if data_params_count == 0 => quote! {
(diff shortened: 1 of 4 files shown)


soroban-sdk-macros/src/derive_event.rs
```diff
@@ -258,14 +253,26 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
...
+            // Must be sorted for map_new_from_slices
+            let data_idents_sorted = data_params
+                .iter()
+                .sorted_by_key(|p| p.name.to_string())
+                .map(|p| format_ident!("{}", p.name.to_string()))
+                .collect::<Vec<_>>();
+            let data_strs_sorted = data_idents_sorted
+                .iter()
+                .map(|i| i.to_string())
+                .collect::<Vec<_>>();
```
mootz12, line 266: rebuilding these for clarity. We could just do:

```
            // Must be sorted for map_new_from_slices
            let data_idents_sorted: Vec<_> = data_idents
                .iter()
                .sorted_by_key(|i| i.to_string())
                .collect();
            let data_strs_sorted: Vec<_> =
                data_idents_sorted.iter().map(|i| i.to_string()).collect();
```

but this sorts based on the output of `format_ident` instead of `p.name`. Currently this is the same.

soroban-sdk/src/tests/contract_event.rs
```diff
@@ -399,6 +399,46 @@ fn test_data_vec_no_data() {
...
+        addr: symbol_short!("user"),
+        time: 1000u64,
+        amount: 500i128,
+    };
+    env.as_contract(&id, || {
+        event.publish(&env);
+    });
+
+    let data: Val = (1000u64, 500i128).into_val(&env);
+    let expected_event = xdr::ContractEvent {
```
leighmcculloch, line 429: If this test asserts against the tuple form, it will be easier to cherry-pick this commit unchanged to the v23 branch. This is fine too, but the commit will need modifying.

leighmcculloch (approved):

soroban-sdk/src/tests/contract_event.rs
mootz12: To keep it consistent with the other tests in this file I think I'll just modify it on the v23 branch.

It will have additional changes anyway due to test snapshots changing format.

PR #1687, opened 2026-01-20
Allow unreachable_code in wasm error helper

leighmcculloch:
### What

Add `#[allow(unreachable_code)]` attribute to the `error_from_error_val` function for wasm targets to suppress a compiler warning introduced in Rust 1.92.0.

### Why

I started seeing the following errors when building contracts using the soroban-sdk:

```
warning: unreachable expression
    --> soroban-sdk/src/env.rs:1770:6
     |
1769 |         self.env_impl.error_from_error_val(e)
     |         ------------------------------------- any code following this expression is unreachable
1770 |     }
     |      ^ unreachable expression
     |
note: this expression has type `Infallible`, which is uninhabited
    --> soroban-sdk/src/env.rs:1769:9
     |
1769 |         self.env_impl.error_from_error_val(e)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = note: `#[warn(unreachable_code)]` (part of `#[warn(unused)]`) on by default
```

It appears Rust 1.92.0 changed how `unreachable_code` warnings are detected.

The `error_from_error_val` function returns `Infallible`, an uninhabited type. This is intentional behaviour. The function is designed to never return so the warning is suppressed.

I think this is probably the result of https://github.com/rust-lang/rust/pull/142390 which landed in 1.92.0, but that change is quite deep and so I'm not certain of that.

In any case, I don't think what the code is doing here is incorrect, just a new warning is highlighting that there's an unreachable condition here, which is expected, so we should allow the condition.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -7,9 +7,31 @@ on:
         description: 'Version to bump to'
         required: true
 
+permissions:
+  contents: write
+
 jobs:
 
   bump-version:
     uses: stellar/actions/.github/workflows/rust-bump-version.yml@main
     with:
       version: ${{ inputs.version }}
+
+  bump-fuzz-lock:
+    needs: bump-version
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        ref: 'release/v${{ inputs.version }}'
+    - run: rustup update
+    - run: rustup install nightly
+    - name: Update fuzz Cargo.lock
+      run: cargo +nightly update --manifest-path tests/fuzz/fuzz/Cargo.toml soroban-sdk
+    - name: Commit fuzz Cargo.lock
+      run: |
+        git config user.name 'github-actions[bot]'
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/env.rs
```diff
@@ -1765,6 +1765,7 @@ impl internal::EnvBase for Env {
     // When targeting wasm we don't even need to do that, just delegate to
     // the Guest's impl, which calls core::arch::wasm32::unreachable.
     #[cfg(target_family = "wasm")]
+    #[allow(unreachable_code)]
     fn error_from_error_val(&self, e: crate::Error) -> Self::Error {
         self.env_impl.error_from_error_val(e)
     }
```
leighmcculloch, line 1771: @graydon Is it correct of me to think there nothing for us to change about what we're doing in this function?

graydon: as far as I can tell!

graydon (approved):
I concur with your assessment (though I also concur with "this is a little murky and I could be wrong")

PR #1688, opened 2026-01-20
Work around rustdoc ICE caused by intra-doc links in deprecated notes

leighmcculloch:
### What

Remove or escape square brackets in `#[deprecated(note = "...")]` attributes that were being parsed as intra-doc links by rustdoc.

### Why

Running `cargo +nightly doc` causes an internal compiler error (ICE) in rustdoc when it attempts to resolve intra-doc links like `[Env::try_invoke]` within `#[deprecated]` attribute notes. When the deprecated item is re-exported via glob import (`pub use module::*`), rustdoc fails to resolve the link in the correct scope and panics instead of gracefully handling the unresolved link.

This is a workaround for a bug in rustdoc that has been reported upstream:
- https://github.com/rust-lang/rust/issues/151411

This workaround is to unblock PRs on this repo, such as:
- https://github.com/stellar/rs-soroban-sdk/pull/1687

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -450,7 +450,7 @@ impl Env {
 
     /// Get the [Logs] for logging debug events.
     #[inline(always)]
-    #[deprecated(note = "use [Env::logs]")]
+    #[deprecated(note = "use Env::logs")]
     #[doc(hidden)]
     pub fn logger(&self) -> Logs {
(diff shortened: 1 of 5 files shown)


dmkozh (approved):

leighmcculloch:
The ICE has been fixed in rust nightly so this is no longer needed.

PR #1689, opened 2026-01-20
fix(rust_spec): Add doc strings to generated error enum and variants

willemneal:
### What

Insert in doc strings into generated error types and each of their variants

### Why

This way IDEs and generated rust code in generally will add the provided documentation to make it easier for devs to understand what they mean.

### Known limitations

N/A


leighmcculloch (changes requested):
See:
https://github.com/stellar/rs-soroban-sdk/blob/1d0fb5954fed6aa2644dd36376b44308d2793b0b/soroban-spec-rust/src/types.rs#L10-L13

Making it as safe as possible to `contractimport!` an untrusted contract is the tradeoff. It's a tradeoff with a cost for sure, but I think the tradeoff is more valuable.

willemneal:
> > Do you mean as doc tests? Could we not ban code blocks?
> 
> I don't think there's a way to ban code blocks within rust tests. The code generator would have to sanitize them away, or mark code blocks as non-executable, and guard against injection attacks.

I meant that we just don't even generate the doc string if there is a code block. Since they aren't necessary for errors anyway.

PR #1702, opened 2026-01-27
Disallow building with overflow-checks off

leighmcculloch:
### What
- Add a build-script probe in `build.rs` that detects whether `overflow-checks` are enabled by attempting an overflowing `u8` addition and catching whether it panics
- Set a custom `cfg` flag (`soroban_sdk_internal_overflow_checks_enabled`) when checks are on
- Add a `compile_error!` in `lib.rs` that fires when the flag is absent, with a link to the docs

### Why
Cargo's default `release` profile disables `overflow-checks`. Without this guard developers can silently build contracts with wrapping arithmetic, leading to unexpected behavior. The tooling for Soroban contracts that generates boilerplate for new contracts sets `overflow-checks = true` for the release profile, and docs encourage the same pattern, but this makes it a compile-time error so that anyone doesn't accidentally turn that off.

This is a workaround until `cfg(overflow_checks)` is stabilized (rust-lang/rust#111466).

### Known limitations

It may not detect overflow-checks being disabled in cross-compilation setups where the host and target profiles diverge or where the build-override profile is explicitly set to overflow-checks = true when the main profile is set to false. These are advanced configurations, and the check is primarily present to prevent developers using the default release profile that has checks off.


dmkozh (approved):

soroban-sdk/build.rs
```diff
@@ -5,6 +5,34 @@ pub fn main() {
...
+    // release profile that has checks off.
+    //
+    // This build.rs check can be removed once cfg(overflow_checks) is stabilized.
+    // Ref: https://github.com/rust-lang/rust/issues/111466
+    println!("cargo::rustc-check-cfg=cfg(soroban_sdk_internal_overflow_checks_enabled)");
+    let overflow_checks_enabled = std::panic::catch_unwind(|| {
+        let x = std::hint::black_box(255u8);
+        let _ = x + 1;
+    })
+    .is_err();
```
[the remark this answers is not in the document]
leighmcculloch: It should be unnecessary to set a rerun-if-env-changed for RUSTFLAGS.

Cargo fingerprints the full set of compiler flags (including RUSTFLAGS) as part of its build cache key. When RUSTFLAGS changes, Cargo considers the entire build stale — including build scripts — and reruns everything.

It's also important to note that if Cargo didn't consider the whole build stale in that case it would probably be insufficient to rerun just on the env var, because rust flags can be set via at least four different mechanisms, with the env being just one.

jayz22 (approved):

soroban-sdk/src/lib.rs
```diff
@@ -58,6 +58,18 @@ pub mod _migrating;
 #[cfg(all(target_family = "wasm", feature = "testutils"))]
 compile_error!("'testutils' feature is not supported on 'wasm' target");
 
+#[cfg(not(soroban_sdk_internal_overflow_checks_enabled))]
+const _: () = {
+    macro_rules! overflow_checks_not_enabled_error {
+        () => {
+            concat!("'overflow-checks' are disabled in the current profile. ",
+            "To prevent silent integer overflow, add `overflow-checks = true` to `[profile.release]` in the `Cargo.toml`. ",
+            "See https://developers.stellar.org/docs/build/smart-contracts/getting-started/hello-world#release-profile.")
```
leighmcculloch, line 67: I changed the error message to more closely resemble the same error that the stellar-cli produces that I think is easier to understand. From: 
https://github.com/stellar/stellar-cli/blob/a048a57a75762458b487052e0021ea704a926bee/cmd/soroban-cli/src/commands/contract/build.rs#L676-L678

soroban-sdk/src/lib.rs
```diff
@@ -58,6 +58,18 @@ pub mod _migrating;
...
+        };
+    }
+    compile_error!(overflow_checks_not_enabled_error!());
+};
```
mootz12, line 71: Does this need to be a macro? Not sure if `concat!` improves the error message, but if so, can't we just:

```
#[cfg(not(soroban_sdk_internal_overflow_checks_enabled))]
compile_error!(concat!(
    "'overflow-checks' are disabled in the current profile. ",
    "To prevent silent integer overflow, add `overflow-checks = true` to `[profile.release]` in the `Cargo.toml`. ",
    "See https://developers.stellar.org/docs/build/smart-contracts/getting-started/hello-world#release-profile."
));
```

mootz12 (approved):

soroban-sdk/src/lib.rs
leighmcculloch: It doesn't need to be a macro, but when the the text is directly in the compiler error the error that prints out is this duplicated large block of text. So I was just trying to get it so that the compiler error prints once.

leighmcculloch: Concat just keeps it a single line in the final which renders better in the terminal, but keeping it more horizontally compressed in the code makes it easier to edit and read there without wrapping.

leighmcculloch:
## CI Failure Analysis & Fix

### Root Causes

**1. test-example (privacy-pools) & test-example (privacy-pools/contract)** — `duplicate lang item panic_impl`

The `soroban-examples/privacy-pools` example depends on `soroban-poseidon` (from crates.io), which transitively depends on `soroban-sdk` from crates.io. The CI workflow's sed-based patching only replaces direct `soroban-sdk` dependencies in local `Cargo.toml` files — it cannot patch transitive dependencies pulled from the crates.io registry. This results in two different compilations of `soroban_sdk` in the dependency graph, each defining `#[panic_handler]`, causing `error[E0152]: duplicate lang item`.

**2. cargo-deny (advisories)** — `unsupported CVSS version: 4.0`

The advisory database now contains CVSS v4.0 entries that the old version of `cargo-deny` cannot parse. This was already fixed on `main` in commit 78ede6a (PR #1742), but this branch hasn't been merged with `main` since.

### Fix

Two changes needed:

**a) Merge `main`** — picks up the cargo-deny bump and version 25.1.1 changes.

**b) Add `[patch.crates-io]` to the workflow patching step** in `.github/workflows/test-with-soroban-examples.yml` — after the existing sed-based patching loop (line 93), add:

```yaml
        # Add [patch.crates-io] to workspace root so transitive dependencies
        # from crates.io also resolve to the local path versions.
        workspace_cargo="$workspace_root/Cargo.toml"
        if ! grep -q '\[patch\.crates-io\]' "$workspace_cargo"; then
          {
            echo ""
            echo "[patch.crates-io]"
            for crate in $crates; do
              rel_path=$(realpath --relative-to="$workspace_root" ${{ github.workspace }}/rs-soroban-sdk/$crate)
              echo "$crate = { path = \"$rel_path\" }"
            done
          } >> "$workspace_cargo"
        fi
```

This ensures that transitive crates.io dependencies (e.g. `soroban-poseidon → soroban-sdk`) also resolve to the local path versions, preventing the duplicate `panic_impl` error.

### To apply

```bash
git fetch origin main
git merge origin/main --no-edit
# Then apply the workflow change above to .github/workflows/test-with-soroban-examples.yml
```

---
*Generated by Copilot CLI — could not push directly due to token permission limitations.*

PR #1703, opened 2026-01-27
Fix range bound overflow in Vec/Bytes slice and GenRange gen_range for u64

leighmcculloch:
### What

- Replace unchecked `+1` / `-1` arithmetic in `Bytes::slice`, `Vec::slice`, and `Prng::gen_range` (u64) range bound conversions with `checked_add` / `checked_sub` using `expect_optimized`
- Add `expect_optimized` to the `UnwrapOptimized` trait for `Option` and `Result`, which traps via `wasm32::unreachable()` on Wasm and panics with the message in tests
- Add test coverage for `Bytes::slice`, `Vec::slice`, and `Prng::gen_range` boundary conditions including overflow edge cases

### Why

When compiled with `overflow-checks = false` (the default for release builds), the bare arithmetic in those functions silently wraps on boundary values like `u32::MAX` or `u64::MAX`. This causes the range passed to the host to differ from the caller's intent — for example, `Bytes::slice(0..=u32::MAX)` wraps to `slice(0..0)` returning empty, and `prng.gen_range((Bound::Excluded(u64::MAX), Bound::Unbounded))` wraps to the full `0..=u64::MAX` range. Contracts using user-controlled or computed range bounds could silently operate on wrong data or generate random numbers from an unintended range.

The tooling for Soroban contracts that generates boilerplate for new contracts sets `overflow-checks = true` for the release profile, and docs encourage the same pattern, but this ensures that these overflows are caught even if a developer accidentally turns that off.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
 
   complete:
     if: always()
-    needs: [fmt, check-git-rev-deps, semver-checks, build, test, build-fuzz, docs, readme, migration-docs]
+    needs: [fmt, check-git-rev-deps, semver-checks, build, test, expand-test-wasms, build-fuzz, docs, readme, migration-docs]
     runs-on: ubuntu-slim
     steps:
     - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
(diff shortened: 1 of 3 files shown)


soroban-sdk/src/tests/prng_range.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[test]
+fn test_gen_range_u64_exclusive_end() {
+    let env = Env::new_with_config(EnvTestConfig {
+        capture_snapshot_at_drop: false,
+    });
+    let id = env.register(TestPrngRangeContract, ());
+    env.as_contract(&id, || {
+        let val: u64 = env.prng().gen_range(5..10);
```
jayz22: Nit: missing `for _ in 0..100` here

PR #1709, opened 2026-01-29
Fix Result type generation for user-defined error enums named "Error"

### LINKED ISSUE #1710: Generated client code uses wrong Error type when user-defined error enum is named "Error"

All versions, I've noticed this behaviour since inception.

### What did you do?

Defined a contract with a user-defined error enum named `Error` and used it as the error type in a `Result` return type:

```rust
#[contracterror]
#[derive(Debug, PartialEq)]
pub enum Error {
    Overflow = 1,
}

#[contractimpl]
impl Contract {
    pub fn safe_add(a: u64, b: u64) -> Result<u64, Error> {
        a.checked_add(b).ok_or(Error::Overflow)
    }
}
```

Then imported this contract in another contract using `contractimport!` and attempted to use the generated client.

The generated client code should reference the user-defined `Error` enum as a UDT:

```rust
fn try_safe_add(...) -> Result<..., Result<u64, Error>>
```

Where `Error` is the user-defined error enum with the `Overflow` variant.

### What did you see instead?

The generated client code incorrectly uses the built-in `soroban_sdk::Error` type:

```rust
fn try_safe_add(...) -> Result<..., Result<u64, soroban_sdk::Error>>
```

This causes type mismatches when trying to match on the error variants, since `soroban_sdk::Error` doesn't have the user-defined `Overflow` variant.

**Note:** This only happens when the error enum is named exactly "Error". Using other names like "MyError" works correctly:

```rust
#[contracterror]
pub enum MyError {
    Overflow = 1,
}

pub fn safe_add_two(a: u64, b: u64) -> Result<u64, MyError> { ... }
// Generated client correctly produces: Result<..., Result<u64, MyError>>
```

The issue is in `soroban-sdk-macros/src/map_type.rs` where there's a special case that maps any type named "Error" to `ScSpecTypeDef::Error` (the built-in error type), even when it's a user-defined error enum.

willemneal:
### What

When a contract defines its own error enum named `Error` and uses it in a `Result<T, Error>` return type, the generated client code now emits `Result<T, Error>` (referencing the user-defined enum) instead of `Result<T, soroban_sdk::Error>`.

The fix lives entirely in client generation. The on-the-wire spec format is unchanged: a contract function returning `Result<_, Error>` still serializes the error type as `ScSpecTypeDef::Error`, exactly as every prior SDK release has done. During Rust client generation, `soroban-spec-rust` now checks whether the spec also contains a `UdtErrorEnumV0` named `Error`, and if so rewrites every `ScSpecTypeDef::Error` reference to `Udt { name: "Error" }` so that the generated trait references the user-defined type.

If the spec contains no error enum named `Error`, behavior is unchanged: `ScSpecTypeDef::Error` continues to map to `soroban_sdk::Error`, which covers contracts that intentionally use the SDK error type at their boundary.

### Why

Previously, contracts whose error enum was named `Error` (a natural choice — it's what every other Rust crate would name it) silently produced a generated client that returned `Result<_, soroban_sdk::Error>` instead of `Result<_, Error>`, hiding the contract's error variants behind the opaque SDK error type and breaking client-side error matching.

This iteration of the fix moves the resolution from spec emission to client interpretation, per review feedback (https://github.com/stellar/rs-soroban-sdk/pull/1709#discussion_r3145852674). Two benefits over the original approach of changing what `#[contractimpl]` emits:

1. **No spec wire-format change.** The bytes that go into a deployed wasm are identical to what `main` produces today, so no contract's spec hash drifts.
2. **Already-deployed contracts benefit immediately.** Any mainnet contract that defines an `Error` error enum will, the next time someone regenerates a Rust client against its spec with the new `soroban-spec-rust`, produce a correct `Result<_, Error>` signature — no redeploy required.

Example — before the fix:
```rust
// Contract defines:
pub enum Error { Overflow = 1 }
pub fn safe_add(a: u64, b: u64) -> Result<u64, Error>

// Generated client had:
fn try_safe_add(...) -> Result<..., Result<soroban_sdk::Error, ...>>
```

Example — after the fix:
```rust
// Generated client now correctly has:
fn try_safe_add(...) -> Result<..., Result<Error, ...>>
```

Closes https://github.com/stellar/rs-soroban-sdk/issues/1710

### Known limitations

If a contract both defines an error enum named `Error` *and* uses `soroban_sdk::Error` directly at a function boundary, every `Error` reference in the spec resolves to the user-defined UDT — there is no way to mix the two in one spec. This matches user intent: a contract that names its own error enum `Error` will not also be using the SDK error type by that name.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/soroban-spec-rust/src/lib.rs
+++ b/soroban-spec-rust/src/lib.rs
@@ -2,13 +2,14 @@ mod syn_ext;
 pub mod r#trait;
 pub mod types;
 
+use std::borrow::Cow;
 use std::{fs, io};
 
 use proc_macro2::TokenStream;
 use quote::quote;
 use sha2::{Digest, Sha256};
 use stellar_xdr::curr as stellar_xdr;
-use stellar_xdr::ScSpecEntry;
+use stellar_xdr::{ScSpecEntry, ScSpecTypeDef, ScSpecTypeUdt, ScSpecUdtUnionCaseV0};
 use syn::Error;
 
 use soroban_spec::read::{from_wasm, FromWasmError};
@@ -107,6 +108,9 @@ pub fn generate_without_file_with_options(
     specs: &[ScSpecEntry],
     opts: &GenerateOptions,
 ) -> Result<TokenStream, GenerateError> {
+    let specs = apply_error_udt_override(specs);
+    let specs: &[ScSpecEntry] = &specs;
+
     let mut spec_fns = Vec::new();
     let mut spec_structs = Vec::new();
     let mut spec_unions = Vec::new();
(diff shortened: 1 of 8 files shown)


leighmcculloch:
> My only question is whether there would ever be the case that the `Error` type in ScTypeDef would be found in a contract spec. Is it supposed to map to `soroban_sdk::Error`?

Yes I think so. Error is an SDK type, it's a valid type to appear at a contract boundary. Whether it is frequently used is a good question though, it may not be.

soroban-spec-rust/src/lib.rs
```diff
@@ -148,6 +152,93 @@ pub fn generate_without_file_with_options(
     }
 }
 
+/// The `#[contractimpl]` macro emits any type named `Error` as a UDT reference
+/// in the spec, so a contract that uses `soroban_sdk::Error` directly (rather
+/// than defining its own error enum) produces a spec with
+/// `Udt { name: "Error" }` but no matching error-enum entry. When that
```
leighmcculloch: I think we should do the reverse, emit a non-UDT Error.

Why: to reduce the observable behaviour change in contracts themselves, and shift the change to be in the interpretation of the specs.

So any type named Error always gets stored in the spec as the SDK Error type, which is what happens today.

Then, when a client sees the Error type, they first check for a UDT also named Error and use that instead. This I think shifts all the behaviour change to the CLI, which also means all existing contracts can benefit from this too.

leighmcculloch (commented):
See my comment inline. I'd like to flip the solution, same outcome, but more compatible with all existing contracts. Is there a reason you're leaning towards changing how contract specs are built, rather than changing how clients interpret them?

willemneal:
@leighmcculloch Great point! I updated with an inversion. And updated the generated artifacts since there are now the error types.

leighmcculloch (approved):

leighmcculloch:
@willemneal could you resolve conflicts? Thanks!

mootz12:
Closed in favor of #1862

PR #1712, opened 2026-02-02
Omit None-valued Option fields from maps

### LINKED ISSUE #1711: Support optional fields in contracttype and contractevent by omitting None values from Maps

This addresses the discussion at https://github.com/stellar/stellar-protocol/discussions/1750#discussioncomment-15654220 about optimizing how `Option<T>` fields are handled in `contracttype` and `contractevent` macros.

Currently, when a struct with `Option<T>` fields is converted to a Soroban `Map`:
- `None` values are stored as void values in the map, consuming unnecessary space
- When deserializing, all keys must be present even if they represent optional data

This creates two problems:
1. **Inefficiency**: Storing key-value pairs for `None` values wastes resources
2. **Migration difficulty**: Adding new optional fields to existing `contracttype` structs breaks deserialization of previously stored data, since the old data won't have the new keys

Modify the `contracttype` and `contractevent` derive macros to handle `Option<T>` fields specially:

**When converting a struct to a Map (serialization):**
- If a field is `Option::None`, omit the key-value pair entirely from the map
- If a field is `Option::Some(v)`, include the key-value pair as normal

**When converting a Map to a struct (deserialization):**
- If a key is missing from the map and the corresponding field is `Option<T>`, treat it as `Option::None`
- If a key is present, deserialize as normal

This approach:
- Is implemented entirely guest-side in the SDK macros
- Requires no protocol changes or new host functions
- Maintains backwards compatibility for reading existing data
- Enables contract data migration by allowing new optional fields

```mermaid
flowchart TD
    subgraph Serialization
        A[Struct with Option fields] --> B{Field value?}
        B -->|Some| C[Include key-value in Map]
        B -->|None - Current| D[Include key with void value]
        B -->|None - Proposed| E[Omit key from Map]
    end
    
    subgraph Deserialization
        F[Map] --> G{Key exists?}
        G -->|Yes| H[Deserialize value]
        G -->|No - Current| I[Error: missing key]
        G -->|No - Proposed| J[Return None for Option fields]
    end
    
    style D fill:#ffcccc,stroke:#ff0000
    style I fill:#ffcccc,stroke:#ff0000
    style E fill:#ccffcc,stroke:#00ff00
    style J fill:#ccffcc,stroke:#00ff00
```

**Legend:**
- Red = Current behavior (to be changed)
- Green = Proposed behavior

1. **Protocol-level change**: Modify `map_new_from_slices` host function to skip void values (discussed in the linked discussion, but requires protocol changes)

2. **New host function**: Add a `map_new_from_slices_v2` that has this behavior (also requires protocol changes)

3. **Status quo**: Require contracts to manually handle optional field serialization, which is error-prone and verbose

The guest-side macro approach proposed here avoids protocol changes while providing the same developer experience benefits.

leighmcculloch:
### What
Modify `contracttype` and `contractevent` derive macros to omit `Option::None` fields from serialized maps instead of storing void values. When deserializing, missing keys for `Option<T>` fields now return `None`. Backwards compatibility is maintained by treating existing void values as `None`.

### Why
Reduces storage costs by not persisting absent data. Enables contract data migration by allowing new optional fields to be added to existing structs without breaking deserialization of previously stored data.

Close #1711


leighmcculloch:
Some discussion about this issue on this discussion:
- https://github.com/orgs/stellar/discussions/1750#discussioncomment-15668079

PR #1713, opened 2026-02-03
fix: use include_bytes for contractfile macro to improve file dependency tracking

mootz12:
### What

Use `include_bytes!` during the `contractfile!` macro instead of embedding the WASM bytes directly to allow the file to be registered as a dependency, causing Cargo to rebuild when the file changes.

### Why

The `contractimport!` macro contained path where stale WASMs could be used. Generally, this was possible when multiple WASMs were imported into a single crate, often for crates that focused on testing multiple contracts at once. If some of the WASMs were changed but the source code of the testing crate untouched, the new WASMs would not be used in the test run.

To reproduce, the following can be done:
1. Init a project with `stellar contract init ./`
2. Create a new folder at the root, called `test_crate`, and add it to the workspace
3. Write a test in `test_crate` that uses `contractimport!` to verify that invoking `hello(arg: "Dev")` returns `["Hello", "Dev"]`, and modify the provided test at `contract/hello-world/src/test` to also use `contractimport!`.
4. Run `stellar contract build` and `cargo test --all --no-fail-fast`
5. Modify the contract to return "Bug" instead of "Hello" 
6. Run `stellar contract build` and `cargo test --all --no-fail-fast`
7. Note that the test within `contract/hello-world` crate will fail, while the one in `test_crate` will pass, as it uses the stale WASM.

### Known limitations

`include_bytes!` encodes the byte string slightly differently than `Literal::byte_string`, and that can be seen in the `test_expanded` changes.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -546,7 +546,14 @@ pub fn contractfile(metadata: TokenStream) -> TokenStream {
 
     // Determine absolute path to file.
     let file_abs = path::abs_from_rel_to_manifest(&args.file);
-    let file_abs_str = file_abs.to_string_lossy().into_owned();
+    let file_abs_str = match file_abs.to_str() {
+        Some(s) => s,
+        None => {
+            return Error::new(args.file.span(), "file path is not valid UTF-8")
+                .into_compile_error()
+                .into()
+        }
+    };
 
     // Read WASM from file to verify SHA256 hash at compile time.
     let wasm = match fs::read(&file_abs) {


soroban-sdk-macros/src/lib.rs
```diff
@@ -544,9 +544,12 @@ pub fn contractfile(metadata: TokenStream) -> TokenStream {
         Err(e) => return e.write_errors().into(),
     };
 
-    // Read WASM from file.
+    // Determine absolute path to file.
     let file_abs = path::abs_from_rel_to_manifest(&args.file);
-    let wasm = match fs::read(file_abs) {
+    let file_abs_str = file_abs.to_string_lossy().into_owned();
```
leighmcculloch: We should error rather than silently change the file path. This is very much an edge case but could result in some oddities where you verify the SHA256 hash of a different file to that which is included.

mootz12: Good call, added a compiler error if non UTF-8 characters are included in the file path

PR #1714, opened 2026-02-04
Deprecate BLS/BN short type aliases and disambiguate `Fr`

jayz22:
### What

  - Rename `bls12_381::Fr` → `Bls12381Fr` and `bn254::Fr` → `Bn254Fr`, keeping the old names as deprecated type aliases.
  - Deprecate the other BLS12-381 short aliases (`G1Affine`, `G2Affine`, `Fp`, `Fp2`) and the `crypto::BnScalar` re-export.
  - Add `Bls12381Fr` and `Bn254Fr` to the spec type mappings in `map_type.rs` (old mappings preserved).
  - Add missing `ArbitraryBn254Fp` for fuzz coverage.
  - Update all internal code, tests, and migration docs to use the canonical prefixed names.

### Why

Both `bls12_381` and `bn254` modules exported a struct named `Fr`. A contract that writes `use soroban_sdk::crypto::bn254::Fr` and uses it as  a contract type will have `Fr` silently resolve to the BLS12-381 scalar in the spec mapping, causing a confusing type mismatch at runtime (see [example](https://github.com/NethermindEth/stellar-private-transactions/pull/83#discussion_r2758135606)).

Disambiguating the structs and deprecating the short aliases gives existing code a clear compiler warning pointing to the new name, instead of a surprise runtime failure.

### Known limitations

More verbose names, which is better than surprise behavior. Old code continues to compile — users just see deprecation warnings. We can remove the deprecated aliases in a future release.


leighmcculloch:
This seems like a breaking change which I'd be hesitant to roll out now without spanning it over multi-versions. The types are in a major version so for better or worse they're out there.

> Without a good way to deprecate

Can we not put a deprecation on the aliases?

If we really need to we can inject a deprecation with the contractimpl and contracttype macros at the point of use, but I think we should be able to do this on the alias?

sisuresh:
> > Can we not put a deprecation on the aliases?
> 
> I believe that's what @sisuresh tried first, but the deprecation would affect both the alias and its original type. @sisuresh correct me if I'm wrong.

I don't think I tried this. The structure of this changed a couple times during development, but I believe you should be able to deprecate the aliases.

jayz22:
@sisuresh @leighmcculloch @mootz12 I've refreshed this PR to perform (backward compatible) deprecation, instead of removal of the short aliases. Please take another look.

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -723,61 +723,61 @@ mod objects {
...
             bytes[0] %= 0x1a;
-            Ok(Fp::from_array(env, &bytes))
+            Ok(Bls12381Fp::from_array(env, &bytes))
         }
     }
 
-    // For Fp2 (96 bytes)
+    // For Bls12381Fp2 (96 bytes)
     #[derive(Arbitrary, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
     pub struct ArbitraryFp2 {
```
mootz12: [picky] can we fix these arbitrary names to align with the type it is for?

soroban-sdk/src/crypto.rs
```diff
@@ -9,7 +9,10 @@ use crate::{
 pub mod bls12_381;
 pub mod bn254;
 pub(crate) mod utils;
-pub use bn254::Fr as BnScalar;
+/// Deprecated alias for `bn254::Bn254Fr`.
+/// Use `bn254::Bn254Fr` directly instead.
+#[deprecated(note = "use `bn254::Bn254Fr` instead")]
+pub use bn254::Bn254Fr as BnScalar;
```
mootz12, line 15: `BnScalar` should also exist in `map_type` right? If folks use the deprecated import I think it will try and match against `BnScalar`.

mootz12 (changes requested):

soroban-sdk/src/crypto.rs
jayz22: I think we can leave it, since it's already been left off, instead of adding the mapping type marking it as deprecated at the same time.
Honestly when this alias was added I was not thinking about using it as a contract type, instead just a convenience alias for contract internals, probably wasn't the best choice (I just saw contracts doing it so decided to add it).

jayz22: I will add it in for completeness, it is plausible for existing contracts to want to use it as contract type.

soroban-sdk/src/testutils/arbitrary.rs
jayz22: fixed

soroban-sdk/src/crypto.rs
jayz22: added `BnScalar` to map_type

mootz12 (approved):

PR #1716, opened 2026-02-05
Add u8, i8, u16, and i16 support for SorobanArbitrary

### LINKED ISSUE #1717: SorobanArbitrary does not support `u8, i8, u16, i16`
### What did you do?

For this code snippet:

```rust
#[contracttype]
pub struct VoteReceipt {
    /// Whether the voter has voted.
    pub has_voted: bool,
    /// The type of vote cast.
    pub vote_type: u8,
    /// The voting power used.
    pub votes: u128,
}
```

I got this error:

```
error[E0277]: the trait bound `u8: SorobanArbitrary` is not satisfied
  --> packages/governance/src/governor/storage.rs:69:1
   |
69 | #[contracttype]
   | ^^^^^^^^^^^^^^^ the trait `SorobanArbitrary` is not implemented for `u8`
   |
   = help: the following other types implement trait `SorobanArbitrary`:
             i128
             i32
             i64
             u128
             u32
             u64
```

Dove into the code, and saw that the implementations for these native types are missing

The above code snippet to compile fine

### What did you see instead?

The error I've pasted above

ozgunozerk:
### What

Add u8, i8, u16, and i16 support for SorobanArbitrary

### Why

`#[contracttype]` macro is failing due to `u8` is not supported in this code snippet:

```
#[contracttype]
pub struct VoteReceipt {
    /// Whether the voter has voted.
    pub has_voted: bool,
    /// The type of vote cast.
    pub vote_type: u8,
    /// The voting power used.
    pub votes: u128,
}
```

Close https://github.com/stellar/rs-soroban-sdk/issues/1717


leighmcculloch:
Thanks for the PR! Before going further on this, it's worth flagging that `u8` and `u16` aren't currently valid types in Soroban at all — not just for `SorobanArbitrary`. There are no `ScVal::U8` or `ScVal::U16` variants in the XDR, and the Soroban host runtime has no support for them either. The smallest integer types Soroban supports are `u32` / `i32`.

So even with `SorobanArbitrary` implemented for these types, they still wouldn't be usable in `#[contracttype]` structs or as contract function parameters — the `#[contracttype]` macro failure you're seeing is the symptom, not something that can be resolved in this SDK alone.

Adding real support would require:

1. Adding `U8` / `I8` / `U16` / `I16` variants to the `ScVal` XDR type (in stellar-xdr).
2. Adding handling for them in the Soroban host environment (soroban-env-host).
3. Then wiring them through the SDK (including `SorobanArbitrary`).

Implementing `SorobanArbitrary` alone wouldn't add the support — it would just make arbitrary-value generation compile for types the rest of the stack still rejects.

For the example in the PR description, the usual workaround today is to use `u32` for `vote_type`.

PR #1722, opened 2026-02-08
Add commentary to alloc module about use of static_mut_refs and light code reordering and vis changes

leighmcculloch:
### What
Add commentary to the use of `static_mut_refs` about why the use is safe. Reorder the definitions of types in the `alloc` module to follow a linear outside-in narrative. Remove the pub visibility of `BumpPointer`. 

### Why
The current layout placed the public API (`BumpPointer`, `#[global_allocator]`) at the bottom of the file, away from the `static mut` it depends on. Grouping them together makes it easier to provide commentary on them both, which requires understanding both the definition of the static mut as well as the line where the implicit ref is created. A linear top-to-bottom or outside-in organisation of the code tells a story where we are introduced to the code affecting outside code first, the global allocator, and then as we read down unwrap the implementation details. The `BumpPointer` type is not used outside of the module and so does not need to be pub.


soroban-sdk/src/alloc.rs
```diff
@@ -18,8 +18,39 @@
 use crate::unwrap::UnwrapOptimized;
 use core::alloc::{GlobalAlloc, Layout};
 
+#[global_allocator]
+static GLOBAL: BumpPointer = BumpPointer;
+
+struct BumpPointer;
```
leighmcculloch, line 24: Other than the addition of comments below, the only code change is the removal of `pub` from this item.

soroban-sdk/src/alloc.rs
```diff
@@ -88,21 +119,3 @@ impl BumpPointerLocal {
         self.alloc_slow_inline(bytes, align)
     }
 }
-
-pub struct BumpPointer;
```
leighmcculloch, line 92: Other than the addition of comments above, the only code change is the removal of `pub` from this item.

graydon (approved):

PR #1726, opened 2026-02-09
Document alloc feature

leighmcculloch:
### What
Convert the private `alloc` module into a public docs-only module that exports no types or values. Add module-level documentation covering: why `alloc` is disabled by default, how to enable it, how the bump-pointer allocator works, that `dealloc` is a no-op, how to use a custom allocator instead, and a link to the Rust `GlobalAlloc` docs.

Moved the bump allocator to another private file, that is gated on `target_family = "wasm"` so the documentation renders on all targets but the bump allocator is only included on wasm builds.

Add `doc(cfg(feature = "alloc"))` attribute so the feature badge appears in rustdoc.

### Why

The `alloc` module is private, so none of its code comments are visible in rustdoc. Making the module public with documentation gives contract authors visibility into the allocator's behaviour, guidance on when to enable the feature, and awareness that custom allocators are an option. The bump allocator's no-op `dealloc` and its implications (all heap data persists in Wasm linear memory for the VM's lifetime) were undocumented and are worth documenting.

### DIFF
--- a/soroban-sdk/src/alloc/mod.rs
+++ b/soroban-sdk/src/alloc/mod.rs
@@ -24,8 +24,8 @@
 //!
 //! With the feature enabled the SDK registers a global bump-pointer allocator
 //! that services all allocations made through Rust's [`alloc`](alloc_crate)
-//! APIs. This makes heap-allocated types such as `alloc::vec::Vec` and
-//! `alloc::string::String` available inside contracts, and enables SDK helpers
+//! APIs. This makes heap-allocated types such as `::alloc::vec::Vec` and
+//! `::alloc::string::String` available inside contracts, and enables SDK helpers
 //! that require allocation (e.g. [`Bytes::to_alloc_vec`]).
 //!
 //! [alloc_crate]: https://doc.rust-lang.org/alloc/


soroban-sdk/src/alloc/mod.rs
```diff
@@ -0,0 +1,53 @@
+//! Allocator used by contracts built with the `alloc` feature.
+//!
+//! The `alloc` feature is **disabled by default**. Contracts will generally be
+//! implementable using stack-allocated or host-managed types (e.g.
+//! [`Vec`](crate::Vec), [`Map`](crate::Map), [`Bytes`](crate::Bytes)). The
+//! feature is only necessary when when heap allocation is needed.
```
graydon: I think this is a little unclear -- or users are likely to think "heap allocation" is anything dynamic. I think you should spell out that it's only needed when you need to allocate arbitrary-sized values _in the wasm guest linear-memory heap_ and emphasize that the "host-managed objects" live in the host heap, and they are usually sufficient for anything dynamic.

I might write something like:

> The `alloc` feature is **disabled by default**. It exists to support a use-case that is both expensive (in terms of CPU time and code size) and typically unnecessary: allocating unbounded memory, dynamically, in the wasm guest's linear-memory heap. Soroban was designed to avoid this need most of the time: host objects like [`Vec`](crate::Vec), [`Map`](crate::Map), or [`Bytes`](crate::Bytes) already support dynamic growth, but live in the host heap and are efficient to use from the guest without any guest memory allocator. Moreover even if a contract _does_ want guest-memory storage of dynamic data, it can accomplish it in bounded static memory using a crate like [`heapless`](https://crates.io/crates/heapless). Turning on the `alloc` feature should usually be the last choice for contracts with no better option.

graydon (changes requested):
Mostly looks ok but minor change suggested (I think some of the copilot suggestions are also right?)

soroban-sdk/src/alloc/mod.rs
leighmcculloch: That is much clearer. Thank you! Add in 1e298ec

soroban-sdk/src/lib.rs
```diff
@@ -65,10 +65,9 @@ fn handle_panic(_: &core::panic::PanicInfo) -> ! {
     core::arch::wasm32::unreachable()
 }
 
-// Here we provide a `#[global_allocator]` that is a minimal non-freeing bump
-// allocator, appropriate for a WASM blob that runs a single contract call.
-#[cfg(all(feature = "alloc", target_family = "wasm"))]
-mod alloc;
+#[cfg(feature = "alloc")]
+#[cfg_attr(feature = "docs", doc(cfg(feature = "alloc")))]
+pub mod alloc;
```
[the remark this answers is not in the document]
leighmcculloch: This isn't really an issue because `::*` is a discouraged practice and most users of the sdk do not use the `::alloc` crate.

PR #1744, opened 2026-02-27
Fix #[cfg(test)] methods in #[contractimpl] blocks failing to compile with testutils

sisuresh:
## Summary

When a `pub fn` inside a `#[contractimpl]` block is annotated with `#[cfg(test)]`, the macro correctly propagates that attribute onto the generated `__Contract__<method>__invoke_raw_slice` function (via `#(#attrs)*` passthrough in `derive_fn`). However, `derive_contract_function_registration_ctor` only received method idents — not their attributes — so the generated registration ctor referenced `__Contract__<method>__invoke_raw_slice` unconditionally. In a non-test build with the `testutils` feature enabled, this produces:

```
error[E0425]: cannot find value `__Contract__persisted__invoke_raw_slice` in this scope
```

## Fix

Change `derive_contract_function_registration_ctor` to accept `(&Ident, &[Attribute])` pairs instead of just `&Ident`. For each method, extract any `cfg` attributes and emit them on the corresponding registration call in the ctor body, so `#[cfg(test)]` (and other cfg guards) are respected.

## Impact

Narrow compile-time only: any crate with a `#[cfg(test)]` public method inside `#[contractimpl]` would fail to compile with `testutils` enabled. No runtime behavior is affected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/soroban-sdk-macros/src/derive_contractimpl_trait_default_fns_not_overridden.rs
+++ b/soroban-sdk-macros/src/derive_contractimpl_trait_default_fns_not_overridden.rs
@@ -81,7 +81,7 @@ fn derive(args: &Args) -> Result<TokenStream2, Error> {
         &args.crate_path,
         &impl_ty,
         Some(trait_ident),
-        fns.iter().map(|f| &f.ident),
+        fns.iter().map(|f| (&f.ident, f.attrs.as_slice())),
     ));
 
     Ok(output)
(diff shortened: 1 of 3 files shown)


sisuresh:
Here's the issue - https://github.com/stellar/rs-soroban-sdk/issues/1747. It's legitimate in that the compiler errors, but my point is I'm not sure if we want to support this use case.

PR #1760, opened 2026-03-12
Copy trait docs to overridden functions without docs

### LINKED ISSUE #1653: Copy trait function docs to implemented functions that lack their own docs

When implementing a trait with `#[contractimpl(contracttrait)]`, doc strings from trait functions are only copied to the contract spec for default functions. If a contract overrides a trait function but doesn't provide its own documentation, the function has no docs in the generated contract spec, when it could at least have the docs copied over from the trait.

This means contracts that implement trait functions without adding their own documentation will have undocumented functions in their specs, even when the trait itself has comprehensive documentation available.

Related: https://github.com/stellar/rs-soroban-sdk/pull/1507#issuecomment-3667207719

When processing `#[contractimpl(contracttrait)]`, if an implemented function:
1. Overrides a trait function, AND
2. Does not have its own doc comments

Then the doc comments from the trait function should be copied to the generated contract spec for that function.

This would ensure that contract specs are fully documented when using well-documented traits, without requiring developers to duplicate documentation.

1. **Manual duplication**: Developers copy trait docs to each impl function (error-prone, maintenance burden)
2. **Status quo**: Accept that overridden functions without docs will have empty docs in the spec
3. **Shared/referenced docs**: A solution for sharing docs between traits and contracts in a way that doesn't embed them in every contract (would require spec format changes)

leighmcculloch:
### What
When `#[contractimpl(contracttrait)]` is used, overridden trait functions that lack their own doc comments now inherit docs from the trait function definition. Spec generation for overridden functions is moved from `contractspecfn` to the macro bridge, where trait docs are available for fallback. If an overridden function has its own doc comments, those are used (no change in behaviour).

### Example

```rust
#[contracttrait]
pub trait MyTrait {
    /// Transfers tokens between accounts.
    fn transfer(from: Address, to: Address, amount: i128) { /* default */ }
}

impl MyTrait for Contract {
    // No doc comment here — trait docs are inherited into the spec
    fn transfer(from: Address, to: Address, amount: i128) { /* custom impl */ }
}
```

Before: `transfer` in the contract spec has empty docs.
After: `transfer` in the contract spec has "Transfers tokens between accounts."

### Why
When implementing a trait with `#[contractimpl(contracttrait)]`, doc strings from trait functions were only copied to the contract spec for default (non-overridden) functions. If a contract overrode a trait function without adding its own documentation, the function had empty docs in the generated contract spec, even when the trait had comprehensive documentation available. This forced developers to manually duplicate trait docs on every override.

Close #1653

### Target
v26


soroban-sdk-macros/src/lib.rs
```diff
@@ -260,12 +260,24 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
-                #imp
-                #derived_ok
+            // When contracttrait is true, spec generation for overridden
+            // functions is handled by the macro bridge (so that trait docs can
+            // be used as fallback), so contractspecfn is not applied here.
+            let mut output = if args.contracttrait {
+                quote! {
+                    #[#crate_path::contractargs(name = #args_ident, impl_only = true)]
+                    #[#crate_path::contractclient(crate_path = #crate_path_str, name = #client_ident, impl_only = true)]
+                    #imp
```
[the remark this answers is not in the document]
mootz12: I think this is correct. We should emit a compiler error if the users does `#[contractimpl(contracttrait)]` without implementing a trait.

soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
```diff
@@ -99,11 +112,23 @@ pub fn generate_call_to_contractimpl_for_trait(
     spec_ident: &str,
 ) -> TokenStream2 {
     let impl_fn_idents = pub_methods.iter().map(|f| f.sig.ident.to_string());
+    // Serialize impl functions with their doc attrs so the macro bridge can
+    // generate spec entries, falling back to trait docs when the impl function
+    // lacks its own.
+    let impl_fn_strs: Vec<String> = pub_methods
+        .iter()
+        .map(|f| {
+            let doc_attrs: Vec<_> = f.attrs.iter().filter(|a| is_attr_doc(a)).collect();
```
mootz12, line 121: I believe this adds an edge case where a trait override functions can't be cfg gated:

Consider:

```rust
#[contract]
pub struct Contract;

#[contracttrait]
pub trait Trait {
    fn some_fn() -> u32 {
        7
    }
}

impl Trait for Contract {
    #[cfg(some_condition)]
    fn some_fn() -> u32 {
        9
    }
}
```

You would expect that this compiles, and either overrides the default trait function with the cfg condition is true, or uses the trait default if false.

---

I don't think this strictly is a blocker. Trait override functions have some existing edge cases with cfg flags that need to get worked out, so am OK pushing this to a separate issue.

soroban-sdk-macros/src/lib.rs
```diff
@@ -269,12 +269,24 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
         Ok(derived_ok) => {
-            let mut output = quote! {
-                #[#crate_path::contractargs(name = #args_ident, impl_only = true)]
-                #[#crate_path::contractclient(crate_path = #crate_path_str, name = #client_ident, impl_only = true)]
-                #[#crate_path::contractspecfn(name = #ty_str)]
-                #imp
-                #derived_ok
+            // When contracttrait is true, spec generation for overridden
+            // functions is handled by the macro bridge (so that trait docs can
+            // be used as fallback), so contractspecfn is not applied here.
```
mootz12, line 274: I think this breaks associated type resolution for overridden functions, as we skip the code path that does `flatten_associated_items_in_impl_fns`.

Consider:
```rust
#[contract]
pub struct Contract;

#[contracttrait]
pub trait Trait {
    type Value;

    fn echo_value(value: u64) -> u64 {
        value
    }
}

impl Trait for Contract {
    type Value = u64;

    fn echo_value(value: Self::Value) -> Self::Value {
        value + 1
    }
}
```

The resulting interface is:
```
#[soroban_sdk::contractargs(name = "Args")]
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn echo_value(env: soroban_sdk::Env, value: Value) -> Value;
}
```

PR #1761, opened 2026-03-13
Fix try_ client ignoring allow_non_root_auth flag

### LINKED ISSUE #1737: Generated `try_` client methods ignore `allow_non_root_auth` flag

Latest (main branch). The bug is in `soroban-sdk-macros/src/derive_client.rs` lines 306-308.

### What did you do?

Called a `try_` method on a generated client after enabling `mock_all_auths_allowing_non_root_auth()`:

```rust
let client = MyContractClient::new(&env, &contract_id);
client.mock_all_auths_allowing_non_root_auth().try_some_fn(&args);
```

The `try_` method should use `mock_all_auths_allowing_non_root_auth()` internally, matching the behavior of the non-`try_` method.

The non-`try_` method correctly checks the `allow_non_root_auth` flag:

```rust
// derive_client.rs:273-279 (non-try — correct)
if self.mock_all_auths {
    if self.allow_non_root_auth {
        self.env.mock_all_auths_allowing_non_root_auth();
    } else {
        self.env.mock_all_auths();
    }
}
```

### What did you see instead?

The `try_` method always calls `mock_all_auths()`, ignoring the `allow_non_root_auth` flag:

```rust
// derive_client.rs:306-308 (try_ — buggy)
if self.mock_all_auths {
    self.env.mock_all_auths(); // BUG: ignores allow_non_root_auth
}
```

This means `try_` variants silently use different authorization semantics than their non-`try_` counterparts. Auth that should be allowed via non-root auth may be unexpectedly rejected, or tests may pass with `try_` but for the wrong reasons.

**Suggested fix:**

```rust
// derive_client.rs:306-308 — should be:
if self.mock_all_auths {
    if self.allow_non_root_auth {
        self.env.mock_all_auths_allowing_non_root_auth();
    } else {
        self.env.mock_all_auths();
    }
}
```

leighmcculloch:
### What
Generated `try_` client methods now check `allow_non_root_auth` and call `mock_all_auths_allowing_non_root_auth()` when set, matching the behavior of their non-`try_` counterparts. A test is added to `token_client.rs` covering this path.

### Why
`try_` methods always called `mock_all_auths()`, ignoring the `allow_non_root_auth` flag set via `mock_all_auths_allowing_non_root_auth()`. This caused `try_` variants to silently use different authorization semantics, making non-root auth unexpectedly fail when invoked through `try_` methods. Fixes #1737.

Close #1737

### DIFF
--- a/tests-expanded/test_add_i128_tests.rs
+++ b/tests-expanded/test_add_i128_tests.rs
@@ -219,7 +219,11 @@ impl<'a> ContractClient<'a> {
                 self.env.mock_auths(mock_auths);
             }
             if self.mock_all_auths {
-                self.env.mock_all_auths();
+                if self.allow_non_root_auth {
+                    self.env.mock_all_auths_allowing_non_root_auth();
+                } else {
+                    self.env.mock_all_auths();
+                }
             }
         }
         use soroban_sdk::{FromVal, IntoVal};
--- a/tests-expanded/test_add_u128_tests.rs
+++ b/tests-expanded/test_add_u128_tests.rs
@@ -219,7 +219,11 @@ impl<'a> ContractClient<'a> {
                 self.env.mock_auths(mock_auths);
             }
             if self.mock_all_auths {
-                self.env.mock_all_auths();
+                if self.allow_non_root_auth {
+                    self.env.mock_all_auths_allowing_non_root_auth();
+                } else {
+                    self.env.mock_all_auths();
+                }
             }
         }
         use soroban_sdk::{FromVal, IntoVal};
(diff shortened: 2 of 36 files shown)


soroban-sdk-macros/src/derive_client.rs
```diff
@@ -304,7 +304,11 @@ pub fn derive_client_impl(crate_path: &Path, name: &str, fns: &[syn_ext::Fn]) ->
...
                             if self.mock_all_auths {
-                                self.env.mock_all_auths();
+                                if self.allow_non_root_auth {
+                                    self.env.mock_all_auths_allowing_non_root_auth();
+                                } else {
+                                    self.env.mock_all_auths();
+                                }
                             }
                         }
                         use #crate_path::{IntoVal,FromVal};
```
leighmcculloch, line 314: This is the same logic that already lives in the non-try fn:
https://github.com/stellar/rs-soroban-sdk/blob/c7ca0b1223ea1514f7dac8386e0be8d45b31ce4b/soroban-sdk-macros/src/derive_client.rs#L271-L281

dmkozh (approved):

PR #1769, opened 2026-03-17
Fix docs_from_attrs truncating mid-UTF-8 codepoint

### LINKED ISSUE #1768: `docs_from_attrs` can produce invalid UTF-8 when truncating multi-byte characters

25.3.0

### What did you do?

Define a type or function using any of the contract macros that call `docs_from_attrs`, with a doc comment that exceeds 1024 bytes where a multi-byte UTF-8 character (e.g. `é`) straddles the truncation boundary.

The affected macros are:
- `#[contracttype]` on structs (including tuple structs), enums, and integer enums
- `#[contractimpl]` (via spec function generation)
- `#[contracttrait]` (via spec function generation)
- `#[contracterror]` on error enums
- `#[contractevent]` on event structs

The doc string in the spec XDR is truncated to a valid UTF-8 string that fits within 1024 bytes.

### What did you see instead?

`docs_from_attrs` in `soroban-sdk-macros/src/doc.rs` converts the doc string to bytes and truncates at an arbitrary byte boundary using `Vec::truncate`. This can split a multi-byte UTF-8 codepoint (e.g. `é` is 2 bytes: `0xC3 0xA9`), leaving a dangling lead byte and producing invalid UTF-8 in the `StringM` stored in the contract's spec XDR.

This doesn't break anything in rs-soroban-sdk or stellar-cli today since nothing interprets the doc bytes as UTF-8 in a way that would fail. However, it is reasonable to expect that truncation of valid UTF-8 doc comments produces valid UTF-8.

leighmcculloch:
### What
Use `floor_char_boundary` to truncate doc comments at a valid UTF-8 character boundary in `docs_from_attrs`. Add a test that confirms a doc string where a multi-byte character straddles the truncation boundary produces valid UTF-8.

### Why
The current implementation truncates doc bytes with `Vec::truncate` at an arbitrary byte offset, which could split multi-byte UTF-8 codepoints and store invalid UTF-8 in the contract's spec XDR.

Close #1768

### DIFF
--- a/tests-expanded/test_associated_type_chained_tests.rs
+++ b/tests-expanded/test_associated_type_chained_tests.rs
@@ -1,10 +1,9 @@
 #![feature(prelude_import)]
 #![no_std]
-#[prelude_import]
-use core::prelude::rust_2021::*;
 #[macro_use]
 extern crate core;
-extern crate compiler_builtins as _;
+#[prelude_import]
+use core::prelude::rust_2021::*;
 use soroban_sdk::{contract, contractimpl, symbol_short, Env, Error, Vec};
 pub trait AssociatedType {
     type Val;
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -16,7 +16,7 @@ members = [
 
 [workspace.package]
 version = "25.3.0"
-rust-version = "1.84.0"
+rust-version = "1.91.0"
```
leighmcculloch, line 19: `floor_char_boundary` was stabilized in 1.91, so we need to bump up to it. We could include a polyfill but I think it's fine to bump up to a recent version of Rust.

leighmcculloch: We also need:
- https://github.com/stellar/soroban-examples/pull/404

leighmcculloch: And:
- https://github.com/stellar/soroban-examples/pull/405

PR #1773, opened 2026-03-17
Add support for chained associated types and returned associated types

mootz12:
### What

Resolve chained associated types and return-type associated types in `#[contractimpl]` macro expansion. The `flatten_associated_items_in_impl_fns` function now iteratively resolves associated type definitions that reference other associated types (e.g., `type End = Self::Foo` where `type Foo = Self::Base`) and substitutes `Self::*` in function return types, not just input parameters.

### Why

Using chained associated types and returning associated types from interfaces is valid Rust syntax but resulted in unexpected compiler errors. 

### Known limitations

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1923,6 +1923,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "test_associated_types_chained"
+version = "25.3.0"
+dependencies = [
+ "soroban-sdk",
+]
+
+[[package]]
 name = "test_associated_types_contracttrait"
 version = "25.3.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


soroban-sdk-macros/src/syn_ext.rs
```diff
@@ -309,43 +309,71 @@ fn unpack_result(typ: &Type) -> Option<(Type, Type)> {
...
+                    if let Some(resolved) = resolve_self_type(&t.ty, &associated_types) {
+                        *t.ty = resolved;
                     }
                 }
             }
+            if let ReturnType::Type(_, ty) = &mut f.sig.output {
+                if let Some(resolved) = resolve_self_type(ty, &associated_types) {
+                    **ty = resolved;
+                }
+            }
```
mootz12, line 354: This is a (sorta unrelated) change Claude recommended. Calling it out explicitly to see if we intentionally didn't flatten function return values before for a reason I am missing.

PR #1780, opened 2026-03-19
Improve parsing for associated type usage and add better errors for unsupported patterns

### LINKED ISSUE #1779: `#[contractimpl]` macro: associated type flattening is one-pass and non-recursive

Main

### What did you do?

Used a chained associated type alias in my Soroban contract:

```rust
use soroban_sdk::{contract, contractimpl, Env};

pub trait Chained {
    type Base;
    type Foo;
    type End;
    fn get(env: Env) -> Self::End;
}

#[contract] pub struct Contract;

#[contractimpl]
impl Chained for Contract {
    type Base = u64;
    type Foo  = Self::Base;   // alias chain depth 1
    type End  = Self::Foo;    // alias chain depth 2

    fn end(env: Env, a: Self::End) -> Self::End {   // <-- `a` and return type never resolved to u64
        a
    }

    fn foo(env: Env, a: Self::Foo) -> Self::Foo { // <-- return type never resolved to u64
        a
    }
}
```

* A clear compiler error noting I can't use chained associated types like "associated type `Self::End` could not be resolved; ensure it is not referenced by another associated type or use the concrete type directly"
* The contract to compile when a single associated type is used as the return value

### What did you see instead?

A build error `error[E0223]: ambiguous associated type`

mootz12:
### What

Flatten associated types for function return values as well as function inputs. Issue errors so the compiler can clearly warn users when the macro logic cannot resolve an associated type. Currently, associated types can only reference concrete types that are within the macros scope. Anything out of scope (like supertrait types not defined in the same impl block), will throw a compiler error.

### Why

Returning associated types, chaining associated types, or using supertrait associated types resulted in no compiler errors but had unclear build errors. 

For chained associated types and returning associated types, the macro has enough information to flatten these.

Closes #1779 

### Known limitations

None

### DIFF
--- a/soroban-sdk/src/muxed_address.rs
+++ b/soroban-sdk/src/muxed_address.rs
@@ -354,6 +354,25 @@ impl MuxedAddress {
 }
 
 #[cfg(not(target_family = "wasm"))]
+impl From<&MuxedAddress> for ScVal {
+    fn from(v: &MuxedAddress) -> Self {
+        // This conversion occurs only in test utilities, and theoretically all
+        // values should convert to an ScVal because the Env won't let the host
+        // type to exist otherwise, unwrapping. Even if there are edge cases
+        // that don't, this is a trade off for a better test developer
+        // experience.
+        ScVal::try_from_val(&v.env, &v.to_val()).unwrap()
+    }
+}
+
+#[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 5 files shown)


leighmcculloch (approved):
nice error handling using the spans 👍🏻

PR #1788, opened 2026-03-24
Add compiler error when reserved type names are used as contract types

### LINKED ISSUE #1787: `#[contractype]` structs matching built-in types get misclassified

25.3.0

### What did you do?

Wrote a contract with a `#[contracttype]` struct that matched the name of a built-in Soroban type.

PoC:
```rust
#![no_std]
use soroban_sdk::{U256, contract, contractimpl, contracttype, Vec};

#[contracttype]
pub struct Address {
    pub id: u256,
}

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn protected(a: Address) -> u128 {
        a.id
    }
}

#[cfg(test)]
mod test {
    use super::*;
    use soroban_sdk::Env;
    
    // This passes
    #[test]
    fn test_protected() {
        let e = Env::default();
        let contract_id = e.register(Contract, ());

        let address = contract::Address { id: u128::MAX };
        let result = client.protected(&address);
        assert_eq!(result, u128::MAX);
    }
}

#[cfg(test)]
mod test_with_wasm {
    use soroban_sdk::Env;
    mod contract {
        soroban_sdk::contractimport!(file = "../../target/wasm32v1-none/release/test_udt.wasm");
    }

    // this fails!
    #[test]
    fn test_protected_wasm() {
        let e = Env::default();
        let contract_id = e.register(contract::WASM, ());
        let client = contract::Client::new(&e, &contract_id);

        let address = contract::Address { id: u128::MAX };
        let result = client.protected(&address);
        assert_eq!(result, u128::MAX);
    }
}
```

The type not to get silently overwritten. Either it should be supported to override built-in types with udts, or, a compiler error should be emitted.

### What did you see instead?

The type got overwritten in the WASM.

```
error[E0308]: mismatched types
   --> tests/udt/src/lib.rs:183:39
    |
183 |         let result = client.protected(&address);
    |                             --------- ^^^^^^^^ expected `soroban_sdk::Address`, found `contract::Address`
    |                             |
    |                             arguments to this method are incorrect
    |
    = note: `contract::Address` and `soroban_sdk::Address` have similar names, but are actually distinct types
```

mootz12:
### What

Adds a compile-time guard that errors on `#[contracttype]` definitions whose names collide with any of the reserved built-in type names recognized by map_type.

### Why

The map_type macro logic in `soroban-sdk-macros` recognizes SDK built-in types by matching only the last path segment (e.g., Address, Bytes, Symbol, Val). A user-defined `#[contracttype]` with the same name as a built-in would be silently mapped to the built-in type in the contract spec.

Closes #1787 

#### contractevent

A `#[contractevent]` struct can safely be named any reserved type name. This is stored in a separate spec entry `ScSpecEventV0`, and users should be able to emit events titled `Address`.

For fields, custom types require a few implementations generated by the `#[contracttype]` macro.

#### contracterror

A `#[contracterror]` enum can be named the `Error` reserved type name. Other names will error if interacting with the WASM. Given the scenario a user names an `Error` a name like `Address` is not likely, this was omitted from this PR.

Fixing the `Error` collision itself being tracked in #1710 

#### contracttrait

The name of a `#[contracttrait]` trait does not end up in the spec. Types used require implementations generated by the `#[contracttype]` macro.

### Known limitations

Testing was done mostly manually, this could provide some incentive to improve testing for compiler errors.

### DIFF
--- a/soroban-sdk-macros/src/lib.rs
+++ b/soroban-sdk-macros/src/lib.rs
@@ -40,7 +40,7 @@ use derive_trait::derive_trait;
 
 use darling::{ast::NestedMeta, FromMeta};
 use macro_string::MacroString;
-use map_type::is_input_type_spec_safe;
+use map_type::is_mapped_type_udt;
 use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote, ToTokens};
@@ -422,7 +422,7 @@ pub fn contracttype(metadata: TokenStream, input: TokenStream) -> TokenStream {
     let vis = &input.vis;
     let ident = &input.ident;
     let attrs = &input.attrs;
-    match is_input_type_spec_safe(ident, &input.generics) {
+    match is_mapped_type_udt(ident, &input.generics) {
         Ok(()) => {}
         Err(e) => return e.to_compile_error().into(),
     }
--- a/soroban-sdk-macros/src/map_type.rs
+++ b/soroban-sdk-macros/src/map_type.rs
@@ -1,3 +1,4 @@
+use quote::ToTokens;
 use stellar_xdr::curr as stellar_xdr;
 use stellar_xdr::{
     ScSpecTypeBytesN, ScSpecTypeDef, ScSpecTypeMap, ScSpecTypeOption, ScSpecTypeResult,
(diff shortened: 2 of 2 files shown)


soroban-sdk-macros/src/lib.rs
```diff
@@ -421,6 +422,17 @@ pub fn contracttype(metadata: TokenStream, input: TokenStream) -> TokenStream {
     let vis = &input.vis;
     let ident = &input.ident;
     let attrs = &input.attrs;
+    if is_reserved_type_name(&ident.to_string()) {
+        return Error::new(
+            ident.span(),
+            format!(
+                "type name `{}` conflicts with a Soroban type and cannot be used as a contract type name",
```
leighmcculloch:
```suggestion
                "type name `{}` conflicts with a soroban_sdk type and cannot be used as a contract type name",
```

soroban-sdk-macros/src/map_type.rs
```diff
@@ -21,6 +21,51 @@ pub const BN254_FP_SERIALIZED_SIZE: u32 = 32;
 pub const BN254_G1_SERIALIZED_SIZE: u32 = BN254_FP_SERIALIZED_SIZE * 2; // 64
 pub const BN254_G2_SERIALIZED_SIZE: u32 = BN254_G1_SERIALIZED_SIZE * 2; // 128
 
+/// Returns true if `name` matches a built-in Soroban type that
+/// is recognized by `map_type` under the `PathArguments::None` arm. Types
+/// with these names cannot be used as `#[contracttype]` names because
+/// `map_type` would silently classify them as built-in types instead of
+/// user-defined types in the contract spec.
+pub fn is_reserved_type_name(name: &str) -> bool {
```
leighmcculloch: List is missing Vec and Map.

I wonder how we can write a test to make sure there's no gaps between this function and the map_type function?

leighmcculloch: Also, Option, Result.

mootz12: Agreed - attempted a change to use `map_type` under the hood to prevent divergence, and ensure any reference to the contract type would be mapped correctly.

mootz12: Just to clarify - `Result`, `Option`, `Vec`, `BytesN`, etc all work without collision if used directly as a udt, but to prevent confusion they will result in compiler errors now.

soroban-sdk-macros/src/map_type.rs
```diff
@@ -21,6 +21,39 @@ pub const BN254_FP_SERIALIZED_SIZE: u32 = 32;
...
+///
+/// This function is used to check if a user-defined type input maps to a UDT. Otherwise,
+/// the identifier risks collision with a built-in spec type of the same name.
+///
+/// ### Errors
+/// - If `generics` has any parameters, as UDTs don't support generics
+/// - If `ident` cannot be parsed as a Rust type
+/// - If `ident` cannot be mapped to a type with `map_type`
+/// - If the type mapped from `ident` is not a UDT
+pub fn is_input_type_spec_safe(ident: &Ident, generics: &Generics) -> Result<(), Error> {
```
leighmcculloch: I don't think this fn is really about spec safety, it's about is the type valid for contracttype's.
```suggestion
pub fn is_type_contracttype_compatible(ident: &Ident, generics: &Generics) -> Result<(), Error> {
```

soroban-sdk-macros/src/map_type.rs
```diff
@@ -58,6 +91,11 @@ pub fn map_type(t: &Type, allow_ref: bool, allow_hash: bool) -> Result<ScSpecTyp
                     "MuxedAddress" => Ok(ScSpecTypeDef::MuxedAddress),
                     "Timepoint" => Ok(ScSpecTypeDef::Timepoint),
                     "Duration" => Ok(ScSpecTypeDef::Duration),
+                    // Check if types that require generics are being used without any path arguments
+                    "Result" | "Option" | "Vec" | "Map" | "BytesN" | "Hash" => Err(Error::new(
+                        ident.span(),
+                        format!("type {} requires generic arguments", ident),
+                    )),
```
leighmcculloch, line 116: Good call.

soroban-sdk-macros/src/map_type.rs
```diff
@@ -299,4 +337,95 @@ mod test {
...
+
+    #[test]
+    fn test_is_input_type_spec_safe_generic_type_errors() {
+        let input: DeriveInput = parse_quote!(
+            struct BytesN<T> {
+                pub key: T,
+            }
+        );
+        is_input_type_spec_safe(&input.ident, &input.generics)
+            .expect_err("generics unsupported on user-defined types");
```
leighmcculloch: This test says it is asserting on a user-defined type, but it isn't a user defined type, it is BytesN which is an sdk type.

mootz12: We check the generics requirement before testing with `map_type`, as otherwise we need to map `map_type` errors to something useful.

For example, consider `Result<T>`. The `map_type` error would be `incorrect number of generic arguments`, forcing use to string match / translate errors.

Am improving the wording here slightly, and I'll add an additional test case to cover both sdk types that contain generics and non-sdk types.

PR #1794, opened 2026-03-26
Add compiler error when single value event used with multiple data entries

### LINKED ISSUE #1793: `contractevent(data_format = "single-value")` does not reject multiple data fields
#### What version are you using?

soroban-sdk-macros v25.3.0

#### What did you do?

Defined a `#[contractevent]` struct with `data_format = "single-value"` and more than one data field:

```rust
pub struct MyEvent {
    #[topic]
    name: Symbol,
    value1: Symbol,
    value2: Symbol,
}
```

#### What did you expect to see?

A clear compile-time error explaining that `data_format = "single-value"` only supports 0 or 1 data fields.

#### What did you see instead?

The macro expands the data fields using a repetition with no separator. This results in confusing, unactionable compiler errors rather than a diagnostic pointing at the actual problem.

mootz12:
### What

Adds a compiler error during `derive_event` that informs the user single value events can't have multiple data fields.

### Why

The compiler error that surfaced if multiple data fields existed for a single value event was unclear.
```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found keyword `self`
   --> soroban-sdk/src/tests/contract_event.rs:224:5
    |
224 |     #[contractevent(data_format = "single-value")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected one of `.`, `;`, `?`, `}`, or an operator
    |
```

The new error is:
```
error: data_format = "single-value" requires exactly 0 or 1 data fields, but found more
   --> soroban-sdk/src/tests/contract_event.rs:224:5
    |
224 |     #[contractevent(data_format = "single-value")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

Fixes #1793

### Known limitations

None

### DIFF
--- a/soroban-sdk/src/storage.rs
+++ b/soroban-sdk/src/storage.rs
@@ -153,7 +153,7 @@ impl Storage {
     pub fn max_ttl(&self) -> u32 {
         let seq = self.env.ledger().sequence();
         let max = self.env.ledger().max_live_until_ledger();
-        max - seq
+        max.saturating_sub(seq)
     }
 
     /// Returns if there is a value stored for the given key in the currently


soroban-sdk-macros/src/derive_event.rs
```diff
@@ -266,10 +266,17 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
...
-        },
+        DataFormat::SingleValue => {
+            if data_params_count > 1 {
+                errors.push(Error::custom(
+                    "data_format = \"single-value\" requires exactly 0 or 1 data fields, but found more",
+                ));
+            }
+            quote! {
+                use #path::IntoVal;
+                #(self.#data_idents.into_val(env))*
```
leighmcculloch, line 277: What is the data field set to fro 0 data params? Void?

mootz12: Yeah, the match arm uses:

```rust
        DataFormat::SingleValue if data_params_count == 0 => quote! {
            #path::Val::VOID.to_val()
        },
```

PR #1798, opened 2026-03-27
Expand docs for mock auth usage in tests

### LINKED ISSUE #1797: Document potential testing pitfalls with auth

A test using `mock_all_auths` or `mock_all_auths_allowing_non_root_auth` without verifying the resulting auth tree via `env.auths` doesn't perform any verification of the auth logic.

Similarly, `mock_all_auths_allowing_non_root_auth` masks missing `authorize_as_current_contract` calls in cross-contract scenarios.

`mock_all_auths` recommend checking `env.auths`, and including it in the example.

`authorize_as_current_contract` noting the dangers of using `mock_all_auths_allowing_non_root_auth`

None

mootz12:
### What

Expand docs for the various mock authorizations to include additional doctests and callout testing pitfalls.

### Why

Closes #1797

### Known limitations

None

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -1332,7 +1332,7 @@ impl Env {
     ///     // Verify that the expected authorization was required.
     ///     assert_eq!(
     ///         env.auths(),
-    ///         std::vec![(
+    ///         [(
     ///             addr.clone(),
     ///             AuthorizedInvocation {
     ///                 function: AuthorizedFunction::Contract((
@@ -1340,7 +1340,7 @@ impl Env {
     ///                     symbol_short!("hello"),
     ///                     (&addr,).into_val(&env),
     ///                 )),
-    ///                 sub_invocations: std::vec![],
+    ///                 sub_invocations: [].into(),
     ///             }
     ///         )]
     ///     );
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/env.rs
```diff
@@ -1241,19 +1328,35 @@ impl Env {
     ///     let client = HelloContractClient::new(&env, &contract_id);
     ///     let addr = Address::generate(&env);
     ///     client.hello(&addr);
+    ///
+    ///     // Verify that the expected authorization was required.
+    ///     assert_eq!(
+    ///         env.auths(),
+    ///         std::vec![(
```
leighmcculloch: We shouldn't need to import std, it should be possible to compare with the array `[]` or `&[]` I think too, is that right?

soroban-sdk/src/env.rs
```diff
@@ -1241,19 +1328,35 @@ impl Env {
...
+    ///             addr.clone(),
+    ///             AuthorizedInvocation {
+    ///                 function: AuthorizedFunction::Contract((
+    ///                     contract_id,
+    ///                     symbol_short!("hello"),
+    ///                     (&addr,).into_val(&env),
+    ///                 )),
+    ///                 sub_invocations: std::vec![],
```
leighmcculloch: Similar here, I believe you can do [].into() so as not needing to extern crate std.

leighmcculloch (approved):

PR #1801, opened 2026-03-30
Implement CAP-82 checked {U,I}256 arith functions

jayz22:
### What

Implement CAP-82 host functions wrapper for U256 and I256
- `checked_{add, sub, mul, pow}`

as well as other `checked_` arithmetic functions (requiring no new host functions)
- `checked_{div, rem_euclid, shl, shr}`

these new functions return Option, which is `None` on overflow.

### DIFF
--- a/tests/fuzz/fuzz/Cargo.lock
+++ b/tests/fuzz/fuzz/Cargo.lock
@@ -1196,8 +1196,9 @@ checksum = "a507befe795404456341dfab10cef66ead4c041f62b8b11bbb92bffe5d0953e0"
 
 [[package]]
 name = "soroban-builtin-sdk-macros"
-version = "26.0.0-rc.1"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=60398e4cf317eccc8ed32802104834bd1d994928#60398e4cf317eccc8ed32802104834bd1d994928"
+version = "26.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e38f0269c20b3c4c3aa2198b5460c72837e0361295a46b914adbcad67fa83d48"
 dependencies = [
  "itertools",
  "proc-macro2",
@@ -1207,8 +1208,9 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-common"
-version = "26.0.0-rc.1"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=60398e4cf317eccc8ed32802104834bd1d994928#60398e4cf317eccc8ed32802104834bd1d994928"
+version = "26.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77b15a125a4ec51eeee94cfa58ac3e4f29e7f5426e7f4692a1b119ee60d83a06"
 dependencies = [
  "arbitrary",
  "crate-git-revision",
@@ -1225,8 +1227,9 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-guest"
-version = "26.0.0-rc.1"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=60398e4cf317eccc8ed32802104834bd1d994928#60398e4cf317eccc8ed32802104834bd1d994928"
+version = "26.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f85f2b5be7654eca65aa17333cb3fe6b0e6685bc7a40792e28e247e4fef6372e"
 dependencies = [
  "soroban-env-common",
  "static_assertions",
@@ -1234,8 +1237,9 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-host"
-version = "26.0.0-rc.1"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=60398e4cf317eccc8ed32802104834bd1d994928#60398e4cf317eccc8ed32802104834bd1d994928"
+version = "26.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb9c9c66b62af1179dc6265000bb4356caf0b67225568f64f39bd4194925917a"
 dependencies = [
  "ark-bls12-381",
  "ark-bn254",
@@ -1270,8 +1274,9 @@ dependencies = [
 
 [[package]]
 name = "soroban-env-macros"
-version = "26.0.0-rc.1"
-source = "git+https://github.com/stellar/rs-soroban-env?rev=60398e4cf317eccc8ed32802104834bd1d994928#60398e4cf317eccc8ed32802104834bd1d994928"
+version = "26.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d075b3654119b6dac81abd0f8c2f29cdba36edbda102b3c8c3d5f029f099790"
 dependencies = [
  "itertools",
  "proc-macro2",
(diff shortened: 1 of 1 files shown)


soroban-sdk/src/num.rs
```diff
@@ -311,6 +311,90 @@ impl U256 {
...
+    /// Performs checked exponentiation. Returns `None` if overflow occurred.
+    pub fn checked_pow(&self, pow: u32) -> Option<U256> {
+        let val = self
+            .env
+            .u256_checked_pow(self.val, pow.into())
+            .unwrap_infallible();
+        if val.is_void() {
+            None
+        } else {
+            Some(U256::try_from_val(&self.env, &val).unwrap_optimized())
```
leighmcculloch: If the values are guaranteed to be U256 then it should be find to do an unsafe conversion to the U256 value. We do that in places in the SDK where the type is known to be guaranteed.

soroban-sdk/src/num.rs
```diff
@@ -462,6 +546,110 @@ impl I256 {
...
+    /// Returns `true` if dividing `self` by `other` would overflow or divide by zero.
+    /// This covers: `other == 0`, or `self == I256::MIN && other == -1`.
+    fn is_div_overflow(&self, other: &I256) -> bool {
+        let zero = I256::from_i32(&self.env, 0);
+        if *other == zero {
+            return true;
+        }
+        let neg_one = I256::from_i32(&self.env, -1);
+        if *other == neg_one {
+            let min = I256::from_parts(&self.env, i64::MIN, 0, 0, 0);
```
leighmcculloch: Given that this line is somewhat unintuitive with the values for the parts, could we move it into a `I256::min()` fn and then call that from here?

soroban-sdk/src/num.rs
```diff
@@ -462,6 +546,110 @@ impl I256 {
...
+    pub fn checked_rem_euclid(&self, other: &I256) -> Option<I256> {
+        if self.is_div_overflow(other) {
+            return None;
+        }
+        Some(self.rem_euclid(other))
+    }
+
+    /// Performs checked left shift. Returns `None` if `bits >= 256`.
+    pub fn checked_shl(&self, bits: u32) -> Option<I256> {
+        if bits >= 256 {
```
leighmcculloch: Can we add an associated const to I256 to mirror `i32::BITS`. So I256::BITS will be 256, then reference that here?

tests/contracttrait_trait/src/lib.rs
```diff
@@ -200,7 +203,11 @@ mod test {
             "Test u32 values.\nReturns the input unchanged."
         );
 
-        let entry = ScSpecEntry::from_xdr(Contract::spec_xdr_test_i32(), Limits::none()).unwrap();
+        let entry = ScSpecEntry::read_xdr(&mut Limited::new(
+            &mut &Contract::spec_xdr_test_i32()[..],
+            Limits::none(),
+        ))
+        .unwrap();
```
leighmcculloch: Why is this changing?

soroban-sdk/src/num.rs
```diff
@@ -311,6 +311,90 @@ impl U256 {
...
+    pub fn checked_div(&self, other: &U256) -> Option<U256> {
+        if *other == U256::from_u32(&self.env, 0) {
+            return None;
+        }
+        Some(self.div(other))
+    }
+
+    /// Performs checked Euclidean remainder. Returns `None` if `other` is zero.
+    pub fn checked_rem_euclid(&self, other: &U256) -> Option<U256> {
+        if *other == U256::from_u32(&self.env, 0) {
```
leighmcculloch: Creating a U256 with a small value hardcoded like this should be something that we do at compile time, it should make the code slightly smaller I think. Maybe not worth the optimisation. The `from_u32` can't be const unfortunately because of the `&Env`. However, these internal functions can use `U256Val::from_u32(0)` in a const context to do that work at compile time.

I _think_ this would work:
```diff
diff --git c/soroban-sdk/src/num.rs i/soroban-sdk/src/num.rs
index 1d411e7..b62b54b 100644
--- c/soroban-sdk/src/num.rs
+++ i/soroban-sdk/src/num.rs
@@ -192,6 +192,11 @@ pub struct U256 {
 impl_num_wrapping_val_type!(U256, U256Val, U256Small);
 
 impl U256 {
+    fn is_zero(&self) -> bool {
+        const ZERO: U256Val = U256Val::from_u32(0);
+        self.val.as_val().get_payload() == ZERO.as_val().get_payload()
+    }
+
     pub fn from_u32(env: &Env, u: u32) -> Self {
         U256 {
             env: env.clone(),
@@ -366,7 +371,7 @@ impl U256 {
 
     /// Performs checked division. Returns `None` if `other` is zero.
     pub fn checked_div(&self, other: &U256) -> Option<U256> {
-        if *other == U256::from_u32(&self.env, 0) {
+        if other.is_zero() {
             return None;
         }
         Some(self.div(other))
@@ -374,7 +379,7 @@ impl U256 {
 
     /// Performs checked Euclidean remainder. Returns `None` if `other` is zero.
     pub fn checked_rem_euclid(&self, other: &U256) -> Option<U256> {
-        if *other == U256::from_u32(&self.env, 0) {
+        if other.is_zero() {
             return None;
         }
         Some(self.rem_euclid(other))
@@ -419,6 +424,16 @@ pub struct I256 {
 impl_num_wrapping_val_type!(I256, I256Val, I256Small);
 
 impl I256 {
+    fn is_zero(&self) -> bool {
+        const ZERO: I256Val = I256Val::from_i32(0);
+        self.val.as_val().get_payload() == ZERO.as_val().get_payload()
+    }
+
+    fn is_neg_one(&self) -> bool {
+        const NEG_ONE: I256Val = I256Val::from_i32(-1);
+        self.val.as_val().get_payload() == NEG_ONE.as_val().get_payload()
+    }
+
     pub fn from_i32(env: &Env, i: i32) -> Self {
         I256 {
             env: env.clone(),
@@ -602,12 +617,10 @@ impl I256 {
     /// Returns `true` if dividing `self` by `other` would overflow or divide by zero.
     /// This covers: `other == 0`, or `self == I256::MIN && other == -1`.
     fn is_div_overflow(&self, other: &I256) -> bool {
-        let zero = I256::from_i32(&self.env, 0);
-        if *other == zero {
+        if other.is_zero() {
             return true;
         }
-        let neg_one = I256::from_i32(&self.env, -1);
-        if *other == neg_one {
+        if other.is_neg_one() {
             let min = I256::from_parts(&self.env, i64::MIN, 0, 0, 0);
             if *self == min {
                 return true;

```

tests/contracttrait_trait/src/lib.rs
jayz22: The env bump to 26.0.1 pulls in stellar-xdr v26.0.0 where `ScSpecEntry::from_xdr()` was removed. The new API uses `ReadXdr::read_xdr()` with `Limited`.

soroban-sdk/src/num.rs
jayz22: good point. I've added an `unchecked_from_val` method to {U, I}256 that directly converts from `Val` via try_from (that only checks the tag value). I think this is the cheapest way to convert to a `I256Val` from `Val`.

soroban-sdk/src/num.rs
jayz22: good point. done.

soroban-sdk/src/num.rs
jayz22: done.

soroban-sdk/src/num.rs
jayz22: good point, done.

tests/contracttrait_trait/src/lib.rs
jayz22: my bad. I bumped the env version but didn't remove the git commit dep for the stellar-xdr, and this test was failing due to failed version resolution and I thought there was an API change so I asked claude to fix it. 
After resolving the version mismatch, the test passes now and I've reverted this change.

PR #1810, opened 2026-04-01
Return errors instead of panicking in soroban-spec-rust

### LINKED ISSUE #1765: soroban-spec-rust panics on invalid UTF-8 or invalid identifiers in spec entries

All versions through at least 25.3.0.

### What did you do?

Generated Rust bindings from a Wasm contract containing invalid UTF-8 bytes (or invalid Rust identifiers) in ScSpec entry names (function names, type names, field names).

Minimal reproduction as a test:

```rust
#[test]
#[should_panic]
fn test_invalid_utf8() {
    use stellar_xdr::curr::{ScSpecUdtStructV0, StringM, VecM};
    let spec = ScSpecUdtStructV0 {
        doc: "".try_into().unwrap(),
        lib: "".try_into().unwrap(),
        name: StringM::try_from(vec![0xff, 0xfe]).unwrap(), // invalid UTF-8
        fields: VecM::default(),
    };
    types::generate_struct(&spec); // panics
}
```

The generator returns an error when spec entries contain invalid UTF-8 or names that cannot form valid Rust identifiers.

### What did you see instead?

The generator panics (via `unwrap()` or `format_ident!`), crashing the tool or process.

leighmcculloch:
### What
Convert all code-generation functions in `soroban-spec-rust` (`generate_struct`, `generate_union`, `generate_enum`, `generate_error_enum`, `generate_event`, `generate_function`, and their `_with_options` variants) from infallible to fallible, returning `GenerateError`. Propagate `GenerateError` up through `generate_from_wasm_with_options` and other fns.

### Why
The generator panicked on Wasm contracts with invalid UTF-8 bytes or non-identifier strings in spec entry names. Existing TODOs in the source acknowledged this. 

Close #1765

### Targeting

v26


soroban-spec-rust/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+pub mod syn_ext;
```
mootz12, line 1: does this need to be pub? It looks like it's only used internally.

mootz12 (approved):

soroban-spec-rust/src/lib.rs
leighmcculloch: Oops no.

PR #1835, opened 2026-04-16
Update spec shaking v2 to keep alive subtype markers via reference

### LINKED ISSUE #1834: spec shaking v2: Support recursive types

main

### What did you do?

With `experimental_spec_shaking_v2` that recursively references itself through any built-in container (`Vec`, `Option`, `Map`, `Result`, a tuple, or a reference), and use it as a contract function parameter or return type:

  ```rust
  #![no_std]
  use soroban_sdk::{contract, contractimpl, contracttype, Symbol, Vec};

  #[contracttype]
  #[derive(Clone, Debug, Eq, PartialEq)]
  pub struct Recursive {
      pub a: Symbol,
      pub b: Vec<Recursive>,
  }

  #[contract]
  pub struct Contract;

  #[contractimpl]
  impl Contract {
      pub fn recursive(a: Recursive) -> Vec<Recursive> {
          a.b
      }
  }
```

Invoking `recursive` to function as expected

### What did you see instead?

Stack overflow

### PR DESCRIPTION
**Pull request by mootz12**
### What

Introduce `keep_reachable`, a small helper that preserves a function's symbol across the linker's DCE pass without invoking it at runtime.

The following type implementations in `soroban-sdk/src/spec_shaking.rs` have been rewritten from `T::spec_shaking_marker()` to `keep_reachable(T::spec_shaking_marker)`:
  - &T, &mut T
  - Vec<T>, Map<K, V>

This ensures spec markers for nested types don't get eliminated by DCE, but there is no risk of recursive function call paths in the final WASM.

### Why

Closes #1834

The overflow issue was causing by a loop within the `spec_shaking_marker` function. Each type invoked this on any subtypes, so if there was a loop, a stack overflow would occur during execution.

This only applies to `Vec` and `Map` currently. Due to some restrictions in `map_type` with applying lifetime arguments to struct field types, references currently can't be used to contain recursive type definitions. However, given this restriction is unrelated, the `keep_reachable` fix was still applied to those types. 

### Known limitations

This will increase contract size for each occurrence of `keep_reachable`, by ~15 bytes each. Per the SDK test contracts, the contacts impacted saw an increase of ~100 bytes.

- [x] ~Determine if we need to support all types, or if just `Vec<T>` and `Map<K,V>` is sufficient. I don't think we currently allow constructing recursive types without using a container type like `Vec` or `Map`, but this should be confirmed as intentional and documented before relying on it~

### CODE REVIEW

File: soroban-sdk/src/spec_shaking.rs
Comment (mootz12) [inline soroban-sdk/src/spec_shaking.rs:1]: All functional changes exist here. The rest are tests / expanded tests / snapshots / etc.

File: soroban-sdk/src/spec_shaking.rs [hunk 1]
```diff
@@ -34,6 +41,19 @@ pub trait SpecShakingMarker {
...
+///
+/// Same volatile-read technique the macro-generated marker bodies use to
+/// keep their `MARKER` byte statics alive (see `soroban-sdk-macros/src/shaking.rs`),
+/// applied to a function pointer instead of a data pointer.
+#[doc(hidden)]
+#[inline(always)]
+fn keep_reachable(f: fn()) {
+    let _ = unsafe { core::ptr::read_volatile(&f) };
+}
+
```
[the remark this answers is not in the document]
mootz12: I'd prefer to keep the Rust runtime as similar as possible to the WASM one. Ultimately, this doesn't really make a difference either way.

soroban-sdk/src/spec_shaking.rs
```diff
@@ -13,6 +13,15 @@
...
 //! spec XDR in the WASM's contractspecv0 section.
+//!
+//! Types whose layout includes their inner types inline (`Option<T>`,
+//! `Result<T, E>`, tuples) call `T::spec_shaking_marker()` directly for
+//! each inner type.
+//!
+//! Types whose size is independent of their inner types (`Vec<T>`,
+//! `Map<K, V>`, `&T`, `&mut T`) use [`keep_reachable`] to reference each
+//! inner type's marker without calling it. Recursive definitions are
+//! possible through these types, so a direct call would risk a runtime cycle.
```
leighmcculloch, line 24: This seems problematic if `Vec<T>` exists as a function parameter, and `T` references another type `U`. My understanding from reading the implementation is that `U`'s marker fn will not be called because `T`'s wasn't called.

leighmcculloch (commented):
One concern discussed inline.

soroban-sdk/src/spec_shaking.rs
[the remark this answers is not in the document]
leighmcculloch: We already scope a lot to the wasm target, so I don't think we need to avoid that here.

soroban-sdk/src/spec_shaking.rs
```diff
@@ -368,19 +391,19 @@ impl SpecShakingMarker for crate::Duration {}
 impl SpecShakingMarker for crate::Val {}
 impl SpecShakingMarker for crate::Error {}
 
-// Container types - propagate to inner types
+// SDK Container types - Vec and Map use `keep_reachable` to allow for recursive definitions.
 impl<T: SpecShakingMarker> SpecShakingMarker for crate::Vec<T> {
     #[inline(always)]
     fn spec_shaking_marker() {
-        T::spec_shaking_marker();
+        keep_reachable(T::spec_shaking_marker);
```
leighmcculloch, line 399: Instead of the keep_reachable, we could add another fn to `SpecShakingMarker` named something like `spec_shaking_marker_shallow` and Vec's `spec_shaking_marker` would call `T::spec_shaking_marker`. This relies on a little less magic of read_volatile on a fn pointer, and prevents the affect on optimisations of taking a fn pointer to the fn.

But, that aside, I think there's still the issue that this will cause some types like types within `T` to not have their spec marked.

soroban-sdk/src/spec_shaking.rs
mootz12: Updated the test, I think this works OK. The type `T` still has the `U::spec_shaking_marker` call, and `T` is kept alive via the fn-pointer reference.

leighmcculloch: Interesting, so should we just use this method of keeping alive instead of actually calling in all the cases? We don't need two approaches, we can use this fn ptr approach for all cases and just never call the fns?

leighmcculloch: Yeah if this way is more reliable and avoids an actual call at runtime, then lets use this approach for all marker creation, not just the Vec/Map ones. Is there a reason we can't do that?

mootz12: I originally had that, but it does inflate the WASM size more.

I recorded `test_associated_type_chained` between the versions:
main: 1985 bytes
just vec/map: 2067 bytes
all ptrs: 2218 bytes

I'll pull some data from larger contract sources and update here.

leighmcculloch: I guess this is because in wasm Rust forces any fn that has its pointer taken to be a fn in the fn table. So it can't be inlined. So every contract type's marker fn ends up in the fn table even if it is never called. 🤔

soroban-sdk/src/spec_shaking.rs
```diff
@@ -368,19 +391,19 @@ impl SpecShakingMarker for crate::Duration {}
 impl SpecShakingMarker for crate::Val {}
 impl SpecShakingMarker for crate::Error {}
 
-// Container types - propagate to inner types
+// SDK Container types - Vec and Map use `keep_reachable` to allow for recursive definitions.
 impl<T: SpecShakingMarker> SpecShakingMarker for crate::Vec<T> {
```
leighmcculloch, line 396: Do we need to do this for liballoc container types for when the alloc feature is enabled?

leighmcculloch (approved):
One ask inline, otherwise lgtm.

soroban-sdk/src/spec_shaking.rs
```diff
@@ -34,6 +43,16 @@ pub trait SpecShakingMarker {
...
 }
 
+/// Keeps `f`'s symbol alive through the linker's dead-code-elimination
+/// pass without invoking `f` at runtime. Takes `f` as a function pointer
+/// and volatile-reads it. The same technique is used to keep the `MARKER`
+/// byte statics alive (see `soroban-sdk-macros/src/shaking.rs`).
+#[doc(hidden)]
+#[inline(always)]
+fn keep_reachable(f: fn()) {
+    let _ = unsafe { core::ptr::read_volatile(&f) };
```
leighmcculloch: One more thing, I think we should gate this on WASM builds. There's really no reason to include this in a non-WASM build, WASM is such a specialty target that many things are different, and we gate plenty of things by WASM, so I don't think we need to worry about having a divergence there.

soroban-sdk/src/spec_shaking.rs
leighmcculloch: Ignore this question. Alloc types are not supported at contract boundaries like function inputs/outputs or storage/event fields, so they do not appear in specs and do not need supporting.

mootz12: Agree

PR #1837, opened 2026-04-20
Normalize raw identifiers (`r#`) in contract macros

### LINKED ISSUE #1836: Macros do not normalize raw identifiers (`r#`)

main

### What did you do?

Declared a contract type, event, and error using raw identifiers for field and variant names that collide with Rust keywords:

```rust
#![no_std]

use soroban_sdk::{contract, contractevent, contractimpl, Env};

#[contractevent]
pub struct Data {
    pub r#type: u32,
    pub value: u32,
}

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn emit(env: &Env) {
        Data {
            r#type: 1,
            value: 42,
        }
        .publish(&env);
    }
}
```

Either:

1. A compile-time error from the macro explaining that names with `r#` cannot be used
2. The macro normalizes the name by stripping the `r#` prefix

### What did you see instead?

The code compiles cleanly, but any call to `emit` panics.

mootz12:
### What

Normalizes raw identifiers (r#) across the Soroban macros so that every name, generated Rust ident, and downstream comparison/sort key uses the unraw'd form of a user's Ident.

For tokens that must refer back to the user's field / variant / fn, the raw Ident is preserved, so it remains valid Rust.

### Why

Fixes #1836 

### Known limitations

None

### DIFF
--- a/soroban-sdk-macros/src/derive_trait.rs
+++ b/soroban-sdk-macros/src/derive_trait.rs
@@ -38,11 +38,17 @@ fn derive_or_err(metadata: TokenStream2, input: TokenStream2) -> Result<TokenStr
     let input: ItemTrait = parse2(input)?;
 
     let path = &args.crate_path;
-    let spec_name = args.spec_name.unwrap_or(format!("{}Spec", input.ident));
+    let spec_name = args
+        .spec_name
+        .unwrap_or_else(|| format_ident!("{}Spec", input.ident).to_string());
     let spec_ident = format_ident!("{spec_name}");
     let spec_export = args.spec_export;
-    let args_name = args.args_name.unwrap_or(format!("{}Args", input.ident));
-    let client_name = args.client_name.unwrap_or(format!("{}Client", input.ident));
+    let args_name = args
+        .args_name
+        .unwrap_or_else(|| format_ident!("{}Args", input.ident).to_string());
+    let client_name = args
+        .client_name
+        .unwrap_or_else(|| format_ident!("{}Client", input.ident).to_string());
 
     Ok(quote! {
         pub struct #spec_ident;
--- a/soroban-sdk-macros/src/derive_trait.rs
+++ b/soroban-sdk-macros/src/derive_trait.rs
@@ -2,7 +2,7 @@ use crate::default_crate_path;
 use darling::{ast::NestedMeta, Error, FromMeta};
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{format_ident, quote};
-use syn::{parse2, ItemTrait, Path};
+use syn::{ext::IdentExt as _, parse2, ItemTrait, Path};
 
 // See soroban-sdk/docs/contracttrait.md for documentation on how this works.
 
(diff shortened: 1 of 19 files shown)


soroban-sdk/src/tests/contract_udt_raw_identifier.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub r#type: u32,
+    pub r#fn: u32,
+}
+
+#[contracttype]
+#[derive(Debug)]
+pub struct TestType {
+    pub r#type: u32,
+    pub r#value: u32,
+}
```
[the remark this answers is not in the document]
mootz12: fixed

leighmcculloch (commented):
This change is focused on functions, and other soroban spec emtities, and introduces the bespoke 'soroban_name', but I think what it is addressing is the more general problem that whenever composing Ident's or getting their string value we should be .unraw(). And that applies to more than just soroban types but Ident usage in general.

So I think we should start calling `.unraw()` instead of `.to_string()` everywhere Ident's are used.

Thoughts?

mootz12:
> This change is focused on functions, and other soroban spec emtities, and introduces the bespoke 'soroban_name', but I think what it is addressing is the more general problem that whenever composing Ident's or getting their string value we should be .unraw(). And that applies to more than just soroban types but Ident usage in general.
> 
> So I think we should start calling `.unraw()` instead of `.to_string()` everywhere Ident's are used.
> 
> Thoughts?

Yeah, good call. Updated this to just be a blanket update to Ident usage to ensure we properly use `unraw` and `format_ident!` everywhere.

soroban-sdk-macros/src/derive_trait.rs
```diff
@@ -38,11 +38,17 @@ fn derive_or_err(metadata: TokenStream2, input: TokenStream2) -> Result<TokenStr
...
     let spec_ident = format_ident!("{spec_name}");
     let spec_export = args.spec_export;
-    let args_name = args.args_name.unwrap_or(format!("{}Args", input.ident));
-    let client_name = args.client_name.unwrap_or(format!("{}Client", input.ident));
+    let args_name = args
+        .args_name
+        .unwrap_or_else(|| format_ident!("{}Args", input.ident).to_string());
+    let client_name = args
+        .client_name
+        .unwrap_or_else(|| format_ident!("{}Client", input.ident).to_string());
```
leighmcculloch: Why is .to_string() added here?

soroban-sdk-macros/src/lib.rs
```diff
@@ -152,13 +152,13 @@ pub fn contract(metadata: TokenStream, input: TokenStream) -> TokenStream {
     let item = parse_macro_input!(input as ItemStruct);
 
     let ty = &item.ident;
-    let ty_str = quote!(#ty).to_string();
+    let ty_str = ty.unraw().to_string();
 
-    let client_ident = format!("{ty_str}Client");
+    let client_ident = format_ident!("{}Client", ty).to_string();
```
leighmcculloch: Why is .to_string() added here?

soroban-sdk-macros/src/lib.rs
```diff
@@ -239,7 +239,7 @@ pub fn contractimpl(metadata: TokenStream, input: TokenStream) -> TokenStream {
         path.path
             .segments
             .last()
-            .map(|name| format!("{}Args", name.ident))
+            .map(|name| format_ident!("{}Args", name.ident).to_string())
```
leighmcculloch: Similar, why is .to_string() called here?

soroban-sdk-macros/src/lib.rs
```diff
@@ -152,13 +152,13 @@ pub fn contract(metadata: TokenStream, input: TokenStream) -> TokenStream {
...
-    let ty_str = quote!(#ty).to_string();
+    let ty_str = ty.unraw().to_string();
 
-    let client_ident = format!("{ty_str}Client");
+    let client_ident = format_ident!("{}Client", ty).to_string();
     let fn_set_registry_ident = format_ident!("__{}_fn_set_registry", ty_str.to_lowercase());
     let crate_path = &args.crate_path;
     let client = derive_client_type(&args.crate_path, &ty_str, &client_ident);
-    let args_ident = format!("{ty_str}Args");
+    let args_ident = format_ident!("{}Args", ty).to_string();
```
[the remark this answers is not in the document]
mootz12: format_ident performs `unraw` for idents

soroban-sdk-macros/src/derive_trait.rs
mootz12: Updated to prefer `format` and `unraw` when building strings with idents over `format_ident*.to_string`

soroban-sdk/src/tests/contract_udt_raw_identifier.rs
leighmcculloch, line 1: Nice test.

leighmcculloch (approved):

PR #1869, opened 2026-05-12
Preserve cfg attrs for contracttrait default functions

### LINKED ISSUE #1868: Function attributes are dropped for `#[contracttrait]` default methods used by `#[contractimpl(contracttrait)]`

main

### What did you do?

Defined a `#[contracttrait]` with a cfg-gated default function, then implemented the trait with `#[contractimpl(contracttrait)]`:

```rust
#![no_std]
use soroban_sdk::{contract, contractimpl, contracttrait, Env};

#[contracttrait]
pub trait T {
    #[cfg(any())]
    fn hidden(env: Env) -> u32 {
        let _ = env;
        7
    }

    fn shown(env: Env) -> u32 {
        let _ = env;
        8
    }
}

#[contract]
pub struct C;

impl T for C {}
```

The contract should compile. The cfg-disabled `hidden` function should be omitted from the generated wrappers/spec/client/args/test registration code, while `shown` should remain callable.

### What did you see instead?

Compilation fails because the `#[contracttrait]` helper macro serializes default function signatures without preserving `#[cfg]`. Later, `#[contractimpl(contracttrait)]` generates wrappers for `hidden` even though Rust has compiled that trait method out.

The failure looks like:

```text
cannot find method or associated constant `hidden` in trait `T`
```

mootz12:
### What

Preserve pass-through function attributes, including `#[cfg(...)]`, when `#[contracttrait]` records default trait functions for later `#[contractimpl(contracttrait)]` code generation.

This also makes generated args and native test registration code cfg-aware, and adds regression coverage for cfg-disabled defaults, cfg-enabled defaults, cfg-gated overrides, and regular `#[contractimpl]` cfg-gated functions.

### Why

`#[contracttrait]` previously serialized only doc attributes for default trait methods. If a default method was cfg-gated, the generated trait helper lost that cfg and later generated wrappers for a method that Rust had compiled out of the trait.

Cfg-gated overrides also need special handling: an override should suppress the trait default only when the override's cfg is active. Otherwise, the default implementation should remain available.

Native test registration needed the same treatment because registration ctors referenced generated raw wrapper functions. If a wrapper was cfg-disabled but its registration was not, the generated test code could reference a missing symbol.

Fixes #1868 

### Known limitations

`cfg_attr` is not interpreted for function gating. Use direct `#[cfg(...)]` on `#[contracttrait]` default functions and `#[contractimpl]` functions.

For cross-crate `#[contracttrait]` default functions, function-level cfgs are evaluated in the implementing contract crate's configuration context when generating wrappers. Shared trait crates should prefer cfg-gating the whole trait item, or require implementing contracts to define matching features for function-level cfgs.

### DIFF
--- a/soroban-sdk-macros/src/derive_fn.rs
+++ b/soroban-sdk-macros/src/derive_fn.rs
@@ -238,21 +238,13 @@ pub fn derive_contract_function_registration_ctor<'a>(
     crate_path: &Path,
     ty: &Type,
     trait_ident: Option<&Path>,
-    method_idents: impl Iterator<Item = &'a Ident>,
+    fns: impl IntoIterator<Item = &'a syn_ext::Fn>,
 ) -> TokenStream2 {
     if cfg!(not(feature = "testutils")) {
         return quote!();
     }
 
     let ty_str = ty_to_safe_ident_str(ty);
-    let (idents, wrap_idents): (Vec<_>, Vec<_>) = method_idents
-        .map(|ident| {
-            let ident_str = ident.unraw().to_string();
-            let wrap_ident = format_ident!("__{}__{}__invoke_raw_slice", ty_str, ident_str);
-            (ident_str, wrap_ident)
-        })
-        .multiunzip();
-
     let trait_str = trait_ident
         .map(|p| {
             p.segments
@@ -262,21 +254,37 @@ pub fn derive_contract_function_registration_ctor<'a>(
                 .join("_")
         })
         .unwrap_or_else(|| "".to_string());
-    let methods_hash = format!("{:x}", Sha256::digest(idents.join(",").as_bytes()));
-    let ctor_ident = format_ident!("__{ty_str}__{trait_str}__{methods_hash}_ctor");
+
+    let ctors = fns
+        .into_iter()
+        .map(|f| {
+            let ident = f.ident.unraw().to_string();
+            let wrap_ident = format_ident!("__{}__{}__invoke_raw_slice", ty_str, ident);
+            let attrs = f
+                .attrs
+                .iter()
+                .filter(|attr| pass_through_attr_to_gen_code(attr) && !attr.path().is_ident("doc"))
+                .collect::<Vec<_>>();
+            let methods_hash =
+                format!("{:x}", Sha256::digest(quote!(#ident #(#attrs)*).to_string()));
+            let ctor_ident = format_ident!("__{ty_str}__{trait_str}__{methods_hash}_ctor");
+            quote! {
+                #[doc(hidden)]
+                #(#attrs)*
+                #[#crate_path::reexports_for_macros::ctor::ctor(crate_path=#crate_path::reexports_for_macros::ctor)]
+                #[allow(non_snake_case)]
+                fn #ctor_ident() {
+                    <#ty as #crate_path::testutils::ContractFunctionRegister>::register(
+                        #ident,
+                        #[allow(deprecated)]
+                        &#wrap_ident,
+                    );
+                }
+            }
+        })
+        .collect::<Vec<_>>();
 
     quote! {
-        #[doc(hidden)]
-        #[#crate_path::reexports_for_macros::ctor::ctor(crate_path=#crate_path::reexports_for_macros::ctor)]
-        #[allow(non_snake_case)]
-        fn #ctor_ident() {
-            #(
-                <#ty as #crate_path::testutils::ContractFunctionRegister>::register(
-                    #idents,
-                    #[allow(deprecated)]
-                    &#wrap_idents,
-                );
-            )*
-        }
+        #(#ctors)*
     }
 }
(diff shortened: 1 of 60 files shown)


soroban-sdk-macros/src/derive_fn.rs
```diff
@@ -262,21 +254,37 @@ pub fn derive_contract_function_registration_ctor<'a>(
...
+        .map(|f| {
+            let ident = f.ident.unraw().to_string();
+            let wrap_ident = format_ident!("__{}__{}__invoke_raw_slice", ty_str, ident);
+            let attrs = f
+                .attrs
+                .iter()
+                .filter(|attr| pass_through_attr_to_gen_code(attr) && !attr.path().is_ident("doc"))
+                .collect::<Vec<_>>();
+            let methods_hash =
+                format!("{:x}", Sha256::digest(quote!(#ident #(#attrs)*).to_string()));
```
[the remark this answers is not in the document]
mootz12: This compiles, `String` implements `AsRef<[u8]>`

soroban-sdk-macros/src/derive_fn.rs
```diff
@@ -262,21 +254,37 @@ pub fn derive_contract_function_registration_ctor<'a>(
                 .join("_")
         })
         .unwrap_or_else(|| "".to_string());
-    let methods_hash = format!("{:x}", Sha256::digest(idents.join(",").as_bytes()));
-    let ctor_ident = format_ident!("__{ty_str}__{trait_str}__{methods_hash}_ctor");
+
+    let ctors = fns
```
leighmcculloch, line 258: What's driving the change here where a single call to this fn is generating multiple ctors instead of one ctor?

Is it because we need to apply the cfg attributes to the ctors as well?

soroban-sdk/docs/contracttrait.md
```diff
@@ -87,7 +87,11 @@ macro_rules! __contractimpl_for_pause {
...
+The `trait_default_fns` contains stringified signatures and pass-through attributes of all functions with default implementations. We serialize function signatures and supported attributes into strings because the tooling used to parse macros and values doesn't handle raw tokens well as parameters. This information is "remembered" by the declarative macro for later comparison.
+
+Direct `#[cfg(...)]` attributes on default functions are preserved in the generated code. For cross-crate traits, function-level cfgs inside a `#[contracttrait]` are evaluated in the implementing contract crate's configuration context when generating wrappers. Shared trait crates should prefer applying cfg to the whole trait item, or require implementing contracts to define matching features when function-level cfgs are needed.
+
+`cfg_attr` is not interpreted by the bridge. Avoid using `cfg_attr` to conditionally add `cfg` to `#[contracttrait]` default functions or `#[contractimpl]` functions, because generated wrappers may not agree with whether Rust compiles the original function. Use direct `#[cfg(...)]` instead.
```
leighmcculloch, line 94: What is the bridge?

tests-expanded/test_account_tests.rs
```diff
@@ -408,8 +409,9 @@ pub extern "C" fn __Contract____check_auth__invoke_raw_extern(
 }
 #[doc(hidden)]
 #[allow(non_snake_case)]
+#[allow(non_snake_case)]
```
leighmcculloch, line 412: :eyes:

soroban-sdk/docs/contracttrait.md
leighmcculloch, line 94: Thanks for updating the docs too. It took me a couple reads to understand what the two paragraphs about 'direct cfg' and 'cfg_attr' were saying.

I think we should be explicit and simply say "cfg" is supported, "cfg_attr" is unsupported. And further more, we should probably generate a compile error if cfg_attr is used.

leighmcculloch (commented):
👏🏻 Very nice surgical change.

Just comments on the docs to tighten that up a bit.

leighmcculloch:
> 3\. Cross-crate cfg evaluation context

From the AI review, this is a bit concerning. We shouldn't be lifting cfg's from another library, e.g. the OpenZeppelin library, into the contract crate. That's very likely to go sideways.

If we can't support that safely, the `contracttrait` macro should compile error on the presence of cfg's on default fns, and the `contractimpl` macro should compile error on their presence in the pass through as well (to prevent some sort of hijacking).

soroban-sdk-macros/src/derive_fn.rs
mootz12: Yep! Splitting it per function prevents a ctor from trying to register a wrapper function that was cfg’d out

soroban-sdk/docs/contracttrait.md
mootz12: It's weird wording I read over when proofing the agent doc updates. I will update, thanks for calling it out!

mootz12:
> We shouldn't be lifting cfg's from another library, e.g. the OpenZeppelin library, into the contract crate.

At the moment this is required due to the way default functions for traits are implemented, as the macro is evaluated at the `contractimpl` point. 

I'll put up an alternate PR making the `cfg` and `cfg_attr` usage on `contracttrait` definitions result in compiler errors.

mootz12:
Closing in favor of alternate PR: #1871

PR #1870, opened 2026-05-12
Clarify guarantees for Hash

mootz12:
### What

 Update the `Hash` docs to clarify what guarantees the type provides.
 
`Hash<N>` values returned by SDK crypto functions are produced by secure cryptographic hash functions, and host-managed `__check_auth` receives a host-constructed authentication payload hash. However, `Hash<N>` is represented at contract boundaries as `BytesN<N>`, so accepting it as a general contract argument does not prove that caller-supplied bytes came from a hash function.

### Why

The previous docs overstated the construction/provenance guarantees of `Hash<N>`.

### Known limitations

None


soroban-sdk/src/crypto.rs
```diff
@@ -14,15 +14,20 @@ pub(crate) mod utils;
...
+/// [`keccak256`][Crypto::keccak256], the SDK guarantees that the bytes came
+/// from the corresponding secure cryptographic hash function. The host also
+/// provides this guarantee for the first parameter of
+/// [`CustomAccountInterface::__check_auth`][crate::auth::CustomAccountInterface::__check_auth].
+///
+/// `Hash<N>` is represented at contract boundaries as `BytesN<N>`, so accepting
+/// it as a general contract entrypoint argument does not prove that the caller
+/// supplied bytes were produced by a secure cryptographic hash function. Use
+/// `BytesN<N>` for externally supplied digests unless the calling context
+/// provides an explicit provenance guarantee.
```
leighmcculloch: I'd truncate the last part. The last part leaves me wondering in what external situations I could trust, when the answer should be none, at least from the perspective of the platform/protocol.
```suggestion
/// `BytesN<N>` for externally supplied digests.
```

leighmcculloch (approved):
One tweak requested inline, but with that lgtm.

PR #1871, opened 2026-05-14
Fix cfg handling for contracttrait and contractimpl blocks

### LINKED ISSUE #1747: Fix #[cfg(test)] methods in #[contractimpl] blocks failing to compile with testutils
## Summary

When a `pub fn` inside a `#[contractimpl]` block is annotated with `#[cfg(test)]`, the macro correctly propagates that attribute onto the generated `__Contract__<method>__invoke_raw_slice` function (via `#(#attrs)*` passthrough in `derive_fn`). However, `derive_contract_function_registration_ctor` only received method idents — not their attributes — so the generated registration ctor referenced `__Contract__<method>__invoke_raw_slice` unconditionally. In a non-test build with the `testutils` feature enabled, this produces:

```
error[E0425]: cannot find value `__Contract__persisted__invoke_raw_slice` in this scope
```

## Fix

Change `derive_contract_function_registration_ctor` to accept `(&Ident, &[Attribute])` pairs instead of just `&Ident`. For each method, extract any `cfg` attributes and emit them on the corresponding registration call in the ctor body, so `#[cfg(test)]` (and other cfg guards) are respected.

## Impact

Narrow compile-time only: any crate with a `#[cfg(test)]` public method inside `#[contractimpl]` would fail to compile with `testutils` enabled. No runtime behavior is affected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #1868: Function attributes are dropped for `#[contracttrait]` default methods used by `#[contractimpl(contracttrait)]`

mootz12:
### What

Reject `cfg` and `cfg_attr` on `#[contracttrait]` default functions.

Support direct `cfg` on `#[contractimpl(contracttrait)]` override methods. Generate default-function code under the inverse of each cfg-gated override, preserving stacked `cfg` attrs as `all(...)`. Reject `cfg_attr` on these override methods because it is not normalized by the handoff.

Preserve direct `cfg` attrs on generated args helpers and test registration for `#[contractimpl]` methods.

Add compile-fail coverage for unsupported contracttrait handoff attrs, and add an attributes test contract that checks active, inactive, and stacked `cfg` methods in the generated Wasm spec.

### Why

`#[contracttrait]` captures default-function metadata when the trait is defined, but non-overridden default wrappers are generated later where the trait is implemented. Carrying function-level cfgs through that handoff would evaluate them in the implementing crate's cfg context instead of the trait-defining crate's context.

Cfg-gated `#[contractimpl]` methods also need generated args and test registration code to follow the same cfg, otherwise inactive methods can leave behind generated references to code that Rust removed.

Fixes: #1868 #1747

### Known Limitations

* Generated args helpers now preserve pass-through method attrs, including `cfg`, `allow`, `deny`, and `doc`, for all `#[contractimpl]` methods.

* Testutils registration ctor symbol names may change because cfg attrs are included in the internal hash. These symbols are not part of the contract ABI.

### Examples

See integration test for supported `#[cfg]` flows: https://github.com/stellar/rs-soroban-sdk/blob/fix-cfg-support/tests/attributes/src/lib.rs


soroban-sdk/tests/ui.rs
leighmcculloch, line 1: Can we name this compile_fails? or something like that since all the tests in here will be build failures.

soroban-sdk-macros/src/attribute.rs
```diff
@@ -16,6 +17,56 @@ pub fn pass_through_attr_to_gen_code(attr: &Attribute) -> bool {
...
+        attrs,
+        is_attr_cfg_attr,
+        "`cfg_attr` is not supported on `#[contractimpl(contracttrait)]` methods because the generated helper only supports direct `cfg` attrs for default override matching",
+    )
+}
+
+fn reject_cfg_attrs(attrs: &[Attribute], message: &'static str) -> Result<()> {
+    reject_attrs(
+        attrs,
+        |attr| is_attr_cfg(attr) || is_attr_cfg_attr(attr),
```
leighmcculloch: These helper fns have names that are inconsistent, which meant on first read I assume they did different things. `rfg_cfg_attrs` rejects both `cfg` and `cfg_attr`, while `reject_cfg_attr_on_contracttrait_impl_fn` only rejects `cfg_attr`.

soroban-sdk-macros/src/attribute.rs
```diff
@@ -16,6 +17,56 @@ pub fn pass_through_attr_to_gen_code(attr: &Attribute) -> bool {
...
+/// Rejects cfg attributes in the `#[contracttrait]` default-function handoff.
+pub fn reject_cfg_attrs_on_contracttrait_default_fn(attrs: &[Attribute]) -> Result<()> {
+    reject_cfg_attrs(
+        attrs,
+        "`cfg` and `cfg_attr` are not supported on `#[contracttrait]` default functions because they would be evaluated where the default implementation is generated, not where the trait is defined",
+    )
+}
+
+/// Rejects cfg_attr attributes in the `#[contractimpl(contracttrait)]` override handoff.
+pub fn reject_cfg_attr_on_contracttrait_impl_fn(attrs: &[Attribute]) -> Result<()> {
```
leighmcculloch: This fn is used in `contractimpl` not `contracttrait` but when I first read the name I thought it was used in `contracttrait`.

soroban-sdk-macros/src/attribute.rs
```diff
@@ -16,6 +17,56 @@ pub fn pass_through_attr_to_gen_code(attr: &Attribute) -> bool {
...
+) -> Result<()> {
+    let mut error: Option<Error> = None;
+    for attr in attrs.iter().filter(|attr| reject(attr)) {
+        match &mut error {
+            Some(error) => error.combine(Error::new_spanned(attr, message)),
+            None => error = Some(Error::new_spanned(attr, message)),
+        }
+    }
+    error.map_or(Ok(()), Err)
+}
```
leighmcculloch, line 46: The nested nature of these helpers had me swirling a little. Wdyt of organising them like this instead? I find the single-layer nature of it easier to navigate, and in the end there's only one call site for each case and so embedding the error into the call site automatically documents the behaviour with minimal lines of code.

```diff
diff --git i/soroban-sdk-macros/src/attribute.rs w/soroban-sdk-macros/src/attribute.rs
index 693379aa..1ba6894f 100644
--- i/soroban-sdk-macros/src/attribute.rs
+++ w/soroban-sdk-macros/src/attribute.rs
@@ -1,6 +1,6 @@
 use syn::{
-    ext::IdentExt as _, punctuated::Punctuated, Attribute, Data, Error, Fields, FieldsNamed,
-    FieldsUnnamed, Result,
+    ext::IdentExt as _, punctuated::Punctuated, spanned::Spanned, Attribute, Data, Error, Fields,
+    FieldsNamed, FieldsUnnamed, Result,
 };
 
 /// Returns true if the attribute is a doc attribute.
@@ -27,41 +27,18 @@ pub fn is_attr_cfg_attr(attr: &Attribute) -> bool {
     attr.path().is_ident("cfg_attr")
 }
 
-/// Rejects cfg attributes in the `#[contracttrait]` default-function handoff.
-pub fn reject_cfg_attrs_on_contracttrait_default_fn(attrs: &[Attribute]) -> Result<()> {
-    reject_cfg_attrs(
-        attrs,
-        "`cfg` and `cfg_attr` are not supported on `#[contracttrait]` default functions because they would be evaluated where the default implementation is generated, not where the trait is defined",
-    )
-}
-
-/// Rejects cfg_attr attributes in the `#[contractimpl(contracttrait)]` override handoff.
-pub fn reject_cfg_attr_on_contracttrait_impl_fn(attrs: &[Attribute]) -> Result<()> {
-    reject_attrs(
-        attrs,
-        is_attr_cfg_attr,
-        "`cfg_attr` is not supported on `#[contractimpl(contracttrait)]` methods because the generated helper only supports direct `cfg` attrs for default override matching",
-    )
-}
-
-fn reject_cfg_attrs(attrs: &[Attribute], message: &'static str) -> Result<()> {
-    reject_attrs(
-        attrs,
-        |attr| is_attr_cfg(attr) || is_attr_cfg_attr(attr),
-        message,
-    )
-}
-
-fn reject_attrs(
-    attrs: &[Attribute],
-    reject: fn(&Attribute) -> bool,
-    message: &'static str,
-) -> Result<()> {
+/// Combines an error spanned on each item with the given message into a single
+/// `Result`. Returns `Ok(())` if the iterator is empty.
+pub fn reject_items<I>(items: I, message: &str) -> Result<()>
+where
+    I: IntoIterator,
+    I::Item: Spanned,
+{
     let mut error: Option<Error> = None;
-    for attr in attrs.iter().filter(|attr| reject(attr)) {
+    for item in items {
         match &mut error {
-            Some(error) => error.combine(Error::new_spanned(attr, message)),
-            None => error = Some(Error::new_spanned(attr, message)),
+            Some(error) => error.combine(Error::new(item.span(), message)),
+            None => error = Some(Error::new(item.span(), message)),
         }
     }
     error.map_or(Ok(()), Err)
diff --git i/soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs w/soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
index d109dc03..500aa34f 100644
--- i/soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
+++ w/soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
@@ -1,8 +1,5 @@
 use crate::{
-    attribute::{
-        is_attr_cfg, is_attr_doc, reject_cfg_attr_on_contracttrait_impl_fn,
-        reject_cfg_attrs_on_contracttrait_default_fn,
-    },
+    attribute::{is_attr_cfg, is_attr_cfg_attr, is_attr_doc, reject_items},
     default_crate_path, syn_ext,
 };
 use darling::{ast::NestedMeta, Error, FromMeta};
@@ -58,7 +55,11 @@ fn derive(args: &Args, input: &ItemTrait) -> Result<TokenStream2, Error> {
             ..
         }) = i
         {
-            reject_cfg_attrs_on_contracttrait_default_fn(attrs).map_err(Error::from)?;
+            reject_items(
+                attrs.iter().filter(|a| is_attr_cfg(a) || is_attr_cfg_attr(a)),
+                "`cfg` and `cfg_attr` are not supported on `#[contracttrait]` default functions because they would be evaluated where the default implementation is generated, not where the trait is defined",
+            )
+            .map_err(Error::from)?;
             let doc_attrs: Vec<_> = attrs.iter().filter(|a| is_attr_doc(a)).collect();
             fns.push(quote!(#(#doc_attrs)* #sig).to_token_stream().to_string());
         }
@@ -107,7 +108,10 @@ pub fn generate_call_to_contractimpl_for_trait(
     spec_ident: &str,
 ) -> Result<TokenStream2, syn::Error> {
     for method in pub_methods {
-        reject_cfg_attr_on_contracttrait_impl_fn(&method.attrs)?;
+        reject_items(
+            method.attrs.iter().filter(|a| is_attr_cfg_attr(a)),
+            "`cfg_attr` is not supported on `#[contractimpl(contracttrait)]` methods because the generated helper only supports direct `cfg` attrs for default override matching",
+        )?;
     }
     let impl_fns = pub_methods.iter().map(|f| {
         let cfg_attrs = f.attrs.iter().filter(|attr| is_attr_cfg(attr));
```

soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
```diff
@@ -43,18 +49,20 @@ fn derive(args: &Args, input: &ItemTrait) -> Result<TokenStream2, Error> {
...
+    for i in &input.items {
+        if let TraitItem::Fn(TraitItemFn {
             default: Some(_),
             sig,
             attrs,
             ..
-        }) => {
+        }) = i
+        {
+            reject_cfg_attrs_on_contracttrait_default_fn(attrs).map_err(Error::from)?;
```
leighmcculloch: This was something that an agent noticed while I was looking at the PR locally:

When this rejection fires, the error propagates via `?` up through `derive_or_err`, and the top-level `match` at line 27-30 emits only `err.write_errors()` — dropping the original `#input` trait declaration. Any downstream `#[contractimpl(contracttrait)] impl Trait for C` then fails with a follow-on `cannot find macro \`Trait\` in this scope` cascade that buries the actionable diagnostic.

Repro: trait with `#[cfg(any())] fn x() {...}` plus `#[contractimpl(contracttrait)] impl T for C { fn x() {...} }` → user sees the intended cfg-reject AND a cascade error. The existing trybuild fixture defines `DefaultCfg` / `DefaultCfgAttr` but never uses them in an impl block, so this cascade is masked by the current `.stderr` snapshot.

Suggested fix: mirror the `contractimpl` pattern at `soroban-sdk-macros/src/lib.rs:299-303` so the trait body is still emitted on error, e.g. `Err(err) => quote! { #input #err }` (parsing `input` once up front). Also worth extending the trybuild fixture with `impl DefaultCfg for C { ... }` to pin no-cascade behavior.

soroban-sdk-macros/src/derive_fn.rs
```diff
@@ -234,22 +234,31 @@ pub fn derive_pub_fn(
...
+                .attrs
+                .iter()
+                .filter(|attr| is_attr_cfg(attr))
+                .collect::<Vec<_>>();
+            let ident_str = f.ident.unraw().to_string();
             let wrap_ident = format_ident!("__{}__{}__invoke_raw_slice", ty_str, ident_str);
-            (ident_str, wrap_ident)
+            // Include cfg attrs so cfg-gated registrations and their inverse-cfg
+            // defaults do not collide on the same internal ctor symbol.
+            let hash_part = quote!(#(#attrs)* #ident_str).to_string();
```
leighmcculloch: The below was something that an agent noticed while I was looking at the PR locally.

I guess this could be an interesting way of reducing the impact of this change.

---

`quote!(#(#attrs)* #ident_str)` interpolates `ident_str: String` through `String: ToTokens`, which emits it as a *string-literal token* (i.e., the 5 bytes `"add"` with surrounding quotes), not as a bare identifier. As a result, even a no-cfg method like `add` hashes the bytes `"add"` instead of `add`, and every ctor symbol in the repo changes vs. `main` even for contracts that use no `#[cfg]` attribute.

Confirmed in `tests-expanded/test_add_i128_tests.rs`: ctor renamed from `__Contract____7e9e5ac3..._ctor` → `__Contract____7b8a6f33..._ctor` for plain `add` with no cfg. The `_ctor` symbol is `#[doc(hidden)]` and only registered via the `ctor` crate under `feature = "testutils"` — it isn't wasm-exported, so impact is cosmetic/debug-info-only and the expanded-test golden churn is incidental.

Suggested fix: keep no-cfg contracts on their existing hash by building the hash input as bytes rather than going through `quote!`, e.g.:
```rust
let hash_part = if attrs.is_empty() {
    ident_str.clone()
} else {
    quote!(#(#attrs)* #ident_str).to_string()
};
```
or, cleaner, pre-format the cfg prefix into a `String` and concatenate. That preserves the `methods_hash` for the existing surface and only churns hashes for cfg-bearing methods (which is the intended scope of the fix).

leighmcculloch: There was a reason I used a string in the current implementation instead of the raw tokens. I'm sure there was some edge case I had where I'd hit an issue with a compile error. But given that no tests break with this change I might be misremembering, or possibly I had hit a n00b error that I later resolved another way and didn't need to keep it as a string.

soroban-sdk-macros/src/attribute.rs
leighmcculloch: This comment is probably redundant given:
- https://github.com/stellar/rs-soroban-sdk/pull/1871/changes#r3307888684

leighmcculloch (approved):
I'm good as is, but there are a few comments inline that are worth considering I think. If you go ahead and make these changes, get me (or someone else) to take another review because I think the complexity of what's changing here is worth a final review.

soroban-sdk/tests/ui.rs
mootz12: Name was from there docs. But I agree it's confusing. Fixed https://github.com/stellar/rs-soroban-sdk/pull/1871/commits/836a1d7a433d9c906c837d29c55eb75ea01c50af

soroban-sdk-macros/src/derive_contractimpl_trait_macro.rs
mootz12: Good catch. Fixed -> [7ec0f48](https://github.com/stellar/rs-soroban-sdk/pull/1871/commits/7ec0f4868030ab96804143ea7f6b881b9c2cfe12)

soroban-sdk-macros/src/derive_fn.rs
mootz12: Nice, didn't realize this quoted strings. The ident is a string as the test registration function expects a string, which is fine for now.

Fixed -> [0f54934](https://github.com/stellar/rs-soroban-sdk/pull/1871/commits/0f54934a71466f32914f43500ad20e48a1c0b88f)

soroban-sdk-macros/src/attribute.rs
mootz12: I think this is a good change, addresses your concerns with function naming too.

Fixed -> [e81cccb](https://github.com/stellar/rs-soroban-sdk/pull/1871/commits/e81cccb59a437f8c6479acb6d2fa519cf2d5e1bb)

leighmcculloch (approved):

PR #1872, opened 2026-05-14
Surface env!/include! deps in contractmeta! to dep-info

### LINKED ISSUE #1873: `contractmeta!` `env!`/`include!`/`include_str!` inputs not surfaced in dep-info, causing stale sccache artifacts

soroban-sdk v26.0.0 (current `main`), reproducible against `soroban-sdk-macros/src/lib.rs` at commit 98353851 and earlier.

### What did you do?

During recent testing of an upcoming `crate-git-revision` release I discovered some interesting behaviour in `contractmeta!`.

Used the `contractmeta!` macro with a `val` parameter that depends on the build environment, then rebuilt with sccache enabled after changing the environment variable or included file:

```rust
// Cargo.toml sets BUILD_ID via a build script, NOTES.md contents change between builds.
contractmeta!(key = "build_id", val = env!("BUILD_ID"));
contractmeta!(key = "notes", val = include_str!("../NOTES.md"));
```

Build the crate, change `BUILD_ID` (or modify `../NOTES.md`), then rebuild under sccache.

A fresh build artifact reflecting the new `BUILD_ID` / new `NOTES.md` contents. sccache should treat the change in environment variable or included file as a cache-invalidating input.

### What did you see instead?

sccache returns a cache hit on a stale artifact. The generated `contractmetav0` section reflects the previous `BUILD_ID` / previous `NOTES.md` contents.

Root cause: `contractmeta!` accepts an expression for `val` and runs it through `MacroString` (from the [`macro-string`](https://crates.io/crates/macro-string) crate) at proc-macro time to expand `env!`, `include!`, and `include_str!`, so the result can be encoded into the generated `contractmetav0` section. Because that expansion happens entirely inside a proc macro, rustc never observes the underlying environment variable read or file include, and they are not recorded in the crate's dep-info.

Under sccache, the cache key is derived from dep-info. With these inputs missing, two builds with different `env!` values or different included file contents produce the same dep-info, hash to the same sccache key, and a stale cached artifact is reused.

This has not been observed with vanilla `cargo` builds — only with sccache.

Affected:

- `soroban-sdk-macros/src/lib.rs` — the `contractmeta` proc macro

Related upstream issue in the `macro-string` crate: https://github.com/dtolnay/macro-string/issues/29

leighmcculloch:
### What

Update the `contractmeta!` macro to re-emit the raw `val` tokens inside a `const { let _ = { ... }; }` block alongside the existing `MacroString` evaluation so that any `env!`, `include!`, or `include_str!` calls in `val` are seen by rustc and recorded in the crate's dep-info.

### Why

`contractmeta!` accepts an expression for `val` and runs it through `MacroString` (from the `macro-string` crate) to expand `env!`, `include!`, and `include_str!` at proc-macro time so the result can be encoded into the generated `contractmetav0` section. Because that expansion happens inside a proc macro, rustc never observes the underlying environment variable reads or file includes and they are not recorded in the crate's dep-info.

Rebuilding with sccache surfaced this as artifact collisions: builds with different `env!` values or different included file contents produced the same dep-info, hashed to the same sccache key, and a stale cached artifact was reused instead of being rebuilt. We have not observed the same behaviour with vanilla `cargo` builds.

By re-emitting the original `val` tokens in a `const` block that rustc type-checks (but optimises away), the `env!` / `include!` / `include_str!` calls are surfaced to the compiler and their inputs flow through into dep-info, so sccache keys change when the environment variable or included file changes.

See https://github.com/dtolnay/macro-string/issues/29 for the upstream issue this works around.

### Example

```rust
// Cargo.toml sets BUILD_ID via build script, file contents change between builds.
contractmeta!(key = "build_id", val = env!("BUILD_ID"));
contractmeta!(key = "notes", val = include_str!("../NOTES.md"));
```

Before this change: rebuilding under sccache with a different `BUILD_ID` or modified `NOTES.md` could yield a cache hit on a stale artifact, because rustc's dep-info did not list the env var or file as an input.

After this change: rustc sees the `env!` / `include_str!` calls in the emitted `const` block, records them in dep-info, and sccache correctly invalidates.

### Known limitations

N/A


mootz12 (approved):
LG2M! Good catch.

PR #1876, opened 2026-05-14
Deprecate export for spec generation in spec shaking v2

### LINKED ISSUE #1875: spec shaking v2: `export` flag is no longer applicable

`main` with `experimental_spec_shaking_v2`.

### What did you do?

Enabled `experimental_spec_shaking_v2` and used `export = false` on a spec-bearing type:

```rust
#[cfg(feature = "experimental_spec_shaking_v2")]
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
#[contracttype(export = false)]
pub struct Hidden {
    pub x: u32,
}
```

The same problem can occur indirectly when a workspace enables `soroban-sdk/experimental_spec_shaking_v2` and also builds helper crates such as `soroban-token-spec` or `stellar-asset-spec` that contain `#[contractevent(..., export = false)]`.

Spec shaking v2 should derive the final contract spec from reachable contract functions, events, errors, and referenced types. A type reachable from the contract boundary should either participate in shaking or produce a clear diagnostic if the annotation is unsupported.

### What did you see instead?

`export = false` suppresses spec emission, so v2 marker generation is skipped for that type and reachable boundary types can fail trait-bound checks at compile time. `export = true` is also misleading under v2 because post-build shaking can still remove entries that are not reachable. Helper spec crates with `export = false` can fail to compile when the SDK macro dependency is unified with the v2 feature.

### Alternatives

If we still see a need to allow developers to insert entries into the `contractspecv0`, we could add a flag or attribute that emits a keep-alive marker regardless of if the spec is reachable. Keeping `export` as the flag to use would mean we still need to error when `export = false`, or document that this is not recommended behavior and could result in incorrect specs.

mootz12:
### What

Emit a deprecation warning if `export = ...` is used on `contracttype`, `contracterror`, and `contractevent` when `experimental_spec_shaking_v2` is enabled.

Update SDK internal auth and executable types to use normal spec annotations under v2 and keep `export = false` only for non-v2 builds. Remove the manual no-op auth `SpecShakingMarker` implementations because the macros now generate the v2 markers and specs for those types.

The deprecation warning looks like (support for proc macro diagnostics only in nightly, see [RFC 1566](https://github.com/rust-lang/rust/issues/54140))
```
warning: use of deprecated constant `export_false_used::__SOROBAN_EXPORT_ARG_DEPRECATED_FOR_UsedExportFalseEvent`: `export` is a no-op under `experimental_spec_shaking_v2` (specs are determined by reachability) and will be removed in a future release
   --> tests/spec_shaking_v2/src/lib.rs:272:16
    |
272 |     pub struct UsedExportFalseEvent {
```

### Why

`export = false` conflicts with exact spec shaking because it can hide a spec entry that remains reachable from a public contract boundary, causing missing marker coverage or compile-time trait-bound failures. `export = true` is also misleading under v2 because reachability, not a retention hint, determines the final spec.

When contracts used `contracttypes` defined in the `soroban-sdk` like `Context`, these types would not get included in the spec, and clients would need to assume that these were SDK types by name to decode them.

Fixes #1875 

### Known Limitations

* By default, specs include UDTs defined in `soroban-sdk` like `Context`, `Executable`, and get filtered if they remain unused.


soroban-token-spec/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use soroban_sdk::{contractevent, Address, BytesN, String};
 // purpose of generating the complete contract spec.
 
 /// Transfer event published when a classic payment uses the MEMO_TEXT memo type.
-#[contractevent(topics = ["transfer"], data_format = "map", export = false)]
+#[contractevent(topics = ["transfer"], data_format = "map")]
```
mootz12: I'm not sure why both `soroban-token-spec` and `stellar-asset-spec` were `export = false` originally, given the crates were not intended to be consumed by contracts.

Opted to not introduce the feature here as this shouldn't impact dev tooling.

cc @leighmcculloch if you remember otherwise.

soroban-sdk/src/address.rs
```diff
@@ -214,7 +214,14 @@ impl TryFrom<Address> for AccountId {
...
 
-#[contracttype(crate_path = "crate", export = false)]
+#[cfg_attr(
+    feature = "experimental_spec_shaking_v2",
+    contracttype(crate_path = "crate")
+)]
+#[cfg_attr(
+    not(feature = "experimental_spec_shaking_v2"),
+    contracttype(crate_path = "crate", export = false)
+)]
```
leighmcculloch, line 224: Every library, e.g. openzeppelin, will be required to add this cfg too, but the problem is that the library won't have a `experimental_spec_shaking_v2` feature and it's be awkward to expect the entire stack of dependencies would all have the feature and pass them on appropriately.

We are maybe lucky there are not that many libraries today, but it is still awkward.

I'm inclined to say because of this we should defer this change until after v2 is fully rolled out, and between now and then document that export doesn't do anything when v2 is enabled.

soroban-token-spec/src/lib.rs
leighmcculloch: I don't remember unfortunately.

soroban-sdk/src/address.rs
mootz12: Good call, a compiler error is too aggressive.

I still think including these SDK types is a necessary change for complete specs, so I am going to update this PR to emit a deprecation warning instead, and have `spec_shaking_v2` just ignore the flag.

leighmcculloch: Will the deprecation warning be present just when the feature is enabled? That would work. Although what if someone fixes the warnings then turns the feature back off and doesn't make the connection they also need to remove the deprecations again.

The interactions around this still seem awkward.

mootz12: For a contract using spec shaking v2, they will only see this warning if they use `export = false` is in their code. I don't expect contract developers to swap back and forth all that often, and the warning should only appear when they turn the feature on and use the flag.

For contracts not using spec shaking v2 but consume a lib with `spec shaking v2`, they inherit the feature, and should be made aware their export flags are being ignored.

For libraries, if they don't use spec shaking v2, this does nothing. If they do (like openzeppelin), they will see a warning and can remove `export = ...` flags before the flag is removed.

soroban-token-spec/src/lib.rs
mootz12: Now that this is just deprecated, I reverted this change.

soroban-sdk/src/address.rs
leighmcculloch: > For libraries, if they don't use spec shaking v2, this does nothing.

Not quite. If a library doesn't use spec shaking v2, but a contract that imports the library does, they will see the deprecation? Is that correct? Or no, not when building the library just as an import?

leighmcculloch (approved):
Only one concern remaining. But not sure if it is correct or not. See inline. If we can prove or know that it behaves correctly in that situation, then lgtm.

soroban-sdk/src/address.rs
leighmcculloch: I think I'm wrong here, because Rust doesn't surface deprecations from dependencies when building a primary crate. So the deprecation will be injected into the library, but it will be hidden by build because it's a dependency.

mootz12: Yup, had an agent build out a mock setup for this and confirmed the contract dev won't see the deprecation warnings. Note that if they import the dependency from a local version, they will see the warnings.

PR #1879, opened 2026-05-28
Warn when building contracts without stellar-cli

leighmcculloch:
### What

The build script now emits a warning when building wasm contracts without the Stellar CLI environment signal, indicating that the CLI will be required in future versions. The check is always performed on wasm targets, but the warning only triggers when neither the experimental_spec_shaking_v2 feature nor the CLI signal is present.

### Why

This prepares contract developers for an upcoming change where the Stellar CLI will be mandatory for building contracts with soroban-sdk. By warning users now, they have time to adopt `stellar contract build` in their workflows before it becomes a requirement.

### Known limitations

N/A

### DIFF
--- a/soroban-sdk/build.rs
+++ b/soroban-sdk/build.rs
@@ -43,13 +43,7 @@ pub fn main() {
             );
             std::process::exit(1);
         } else {
-            println!("cargo::warning=stellar-cli will be required in a future version of soroban-sdk to build contracts");
-            println!("cargo::warning=");
-            println!("cargo::warning=Building contracts with `cargo build` directly will not be supported in a");
-            println!("cargo::warning=future version of soroban-sdk.");
-            println!("cargo::warning=");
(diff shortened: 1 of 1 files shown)


leighmcculloch:
> cargo::warning is only shown for path-source dependencies by default

This is a good observation about the limitation of the warning in build scripts.

> Print to `stderr` via `eprintln!` (as the error path does) — this is visible regardless of dep source, though it's also "louder."

This is incorrect and not the way we should fix it. Cargo will capture and hide any stderr output of the build script, so it doesn't make it better.

I'll push a fix.

leighmcculloch:
Wow this is almost impossible to do well with cargo. A compile error in a build script is easy, but surfacing a warning or a deprecation style warning is non-trivial and either isn't rendered in the common case or in the hacky cases I got to work looks quite terrible and confusing. 😞

PR #1880, opened 2026-05-29
Require stellar-cli for wasm contract builds

leighmcculloch:
### What

Fail builds of the soroban-sdk that are not performed with `stellar contract build`, and provide an escape hatch when the `SOROBAN_SDK_ALLOW_BUILD_WITHOUT_STELLAR_CLI` env var is present to temporarily allow developers to get past the error during a period of transition.

### Why

The soroban-sdk will soon rely on `stellar contract build` to run post-build steps such as spec shaking v2, and builds that bypass the stellar-cli produce incomplete or incorrect contract artifacts.

We wanted to surface earlier that this requirement is coming so that developers not using the stellar-cli wouldn't get completely blocked at a new release. (Thanks @mootz12 for the idea.)

It would be ideal if this was a warning and not a hard error but unfortunately cargo doesn't provide good ways to display a warning in this case because warnings in dependencies are hidden during builds. There was an attempt to do that in https://github.com/stellar/rs-soroban-sdk/pull/1879.

### Known limitations

The escape hatch suppresses the error unconditionally and does not verify that the resulting wasm has been post-processed; contracts built with `SOROBAN_SDK_ALLOW_BUILD_WITHOUT_STELLAR_CLI=1` may be missing spec shaking and other future post-build steps, and should not be deployed to mainnet.

### DIFF
--- a/soroban-sdk/src/_migrating/v27_stellar_cli.rs
+++ b/soroban-sdk/src/_migrating/v27_stellar_cli.rs
@@ -16,10 +16,7 @@
 //! stellar contract build
 //! ```
 //!
-//! `stellar contract build` sets the
-//! `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2` environment variable
-//! that the soroban-sdk's build script checks for. No further changes to the
-//! contract crate or its `Cargo.toml` are required.
+//! No further changes to the contract crate or its `Cargo.toml` are required.
 //!
 //! [stellar-cli]: https://github.com/stellar/stellar-cli
 //!
@@ -33,8 +30,9 @@
 //! ```
 //!
 //! This escape hatch is provided for the transition and will be removed in a
-//! near future version of soroban-sdk. Contracts built with the escape hatch
-//! may be missing post-build steps and should not be deployed to mainnet.
+//! near future version of soroban-sdk once support for building without the
+//! stellar-cli is completely removed. Contracts built with the escape hatch may
+//! be missing post-build steps and should not be deployed to mainnet.
 //!
 //! ## Non-Wasm Builds
 //!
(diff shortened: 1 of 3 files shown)


soroban-sdk/src/_migrating/v27_stellar_cli.rs
```diff
@@ -0,0 +1,40 @@
...
+//! [stellar-cli]: https://github.com/stellar/stellar-cli
+//!
+//! ## Temporary Escape Hatch
+//!
+//! For environments that cannot yet use the stellar-cli, the error can be
+//! suppressed by setting:
+//!
+//! ```sh
+//! SOROBAN_SDK_ALLOW_BUILD_WITHOUT_STELLAR_CLI=1
+//! ```
```
mootz12, line 30: This is only true for v1 in the current impl. We should include in the documentation that this only applies without `experimental_spec_shaking_v2`

soroban-sdk/src/_migrating/v27_stellar_cli.rs
```diff
@@ -0,0 +1,40 @@
+//! Building contracts requires the stellar-cli.
+//!
+//! ## Breaking Change: Wasm Builds Fail Without stellar-cli
+//!
+//! Building a contract for a wasm target without using the stellar-cli now
+//! results in a compile error. The soroban-sdk relies on the stellar-cli's
+//! `stellar contract build` to perform post-build steps such as spec
+//! optimization. Builds that bypass the stellar-cli produce incomplete or
+//! incorrect contract artifacts.
```
mootz12, line 9: We might want to lighten the wording here, some1 might get the idea that spec_shaking_v1 produced incomplete contract artifacts.

mootz12 (approved):
Comments are just around some wording to clarify that the escape hatch is v1/v2 dependent.

soroban-sdk/src/_migrating/v27_stellar_cli.rs
leighmcculloch: Hmm good point. This is getting pretty complicated. I'm beginning to question how much we need to alert folks ahead of a new SDK release that they need the cli. Because now I think about it someone will only be affected when they upgrade to the new sdk, which is at a time that people expect breakages, expect change.

leighmcculloch:
Closed because of my comment here: https://github.com/stellar/rs-soroban-sdk/pull/1880#discussion_r3328491778

Issue #1723, opened 2026-02-09
Tests: `soroban-examples` `privacy-pools` example fails to build in CI – duplicate `panic_impl` lang item


leighmcculloch:
### What version are you using?

main branch (commit c92d7a6)

### What did you do?

The "Test with soroban-examples" CI workflow fails to build the `privacy-pools` and `privacy-pools/contract` examples. Both the workspace-level and contract-level builds fail with the same error.

Failing jobs:
- [`test-example (privacy-pools)`](https://github.com/stellar/rs-soroban-sdk/actions/runs/21806958059/job/62911864820?pr=1722)
- [`test-example (privacy-pools/contract)`](https://github.com/stellar/rs-soroban-sdk/actions/runs/21806958059/job/62911864811?pr=1722)

### What did you expect to see?

A successful build of the `privacy-pools` example.

### What did you see instead?

```
error[E0152]: duplicate lang item in crate `soroban_sdk` (which `lean_imt` depends on): `panic_impl`
  |
  = note: the lang item is first defined in crate `soroban_sdk` (which `soroban_poseidon` depends on)
  = note: first definition in `soroban_sdk` loaded from .../libsoroban_sdk-5138a0543f72aac4.rmeta
  = note: second definition in `soroban_sdk` loaded from .../libsoroban_sdk-f5ac024b8db104b4.rmeta

error: could not compile `lean-imt` (lib) due to 1 previous error
```

The CI patches SDK versions in the soroban-examples, but `lean-imt` and `soroban-poseidon` end up resolving to different copies of `soroban-sdk`, causing a duplicate `panic_impl` lang item conflict. The patching likely doesn't account for all the transitive dependency paths in the `privacy-pools` workspace.

/cc @jayz22

PR #1884, opened 2026-06-02
Fix failing OpenZeppelin CI jobs

### LINKED ISSUE #1885: Test with OpenZeppelin Contracts CI fails with duplicate panic_impl lang item

rs-soroban-sdk main (commit cdae1e3), building against OpenZeppelin/stellar-contracts main. soroban-sdk 26.0.1 (local checkout) and soroban-poseidon 26.0.0 (crates.io).

### What did you do?

Ran the "Test with OpenZeppelin Contracts" workflow (.github/workflows/test-with-openzeppelin-stellar-contracts.yml), which builds the OpenZeppelin stellar-contracts against the local rs-soroban-sdk checkout for the wasm32v1-none target. It redirects soroban-sdk to the local checkout by rewriting the OpenZeppelin workspace-root Cargo.toml [workspace.dependencies] entries into path overrides, then builds each contract, for example examples/confidential/auditor.

Each OpenZeppelin contract builds against a single copy of soroban-sdk (the local checkout under test), and the workflow passes.

### What did you see instead?

The build fails with `error[E0152]: duplicate lang item in crate soroban_sdk: panic_impl` (example failing run: https://github.com/stellar/rs-soroban-sdk/actions/runs/26823581513/job/79084681622).

leighmcculloch:
### What

Add a `[patch.crates-io]` table to the OpenZeppelin `stellar-contracts` workspace manifest during the "Test with OpenZeppelin Contracts" workflow, so that copies of `soroban-sdk` (and the other publishable rs-soroban-sdk crates) pulled in transitively through published crates resolve to the local checkout rather than crates.io.

### Why

The workflow makes the OpenZeppelin contracts build against the local `soroban-sdk` by rewriting `[workspace.dependencies]` entries into path overrides, but a path override only redirects directly-named dependency edges. After `stellar-tokens` gained a dependency on the crates.io `soroban-poseidon`, which itself depends on `soroban-sdk` from crates.io, that transitive copy was left unpatched. The wasm build then linked two copies of `soroban-sdk` and failed with a duplicate `panic_impl` lang item error (E0152), the Cargo limitation tracked in #1723, breaking every contract that depends on `stellar-tokens` (for example `examples/confidential/auditor`). A `[patch.crates-io]` table redirects transitive copies as well, collapsing the graph onto a single local `soroban-sdk`.

Close #1885

### Known limitations

The patch only unifies the copies while the local `soroban-sdk` version is semver-compatible with the version required by the transitive dependency; if they diverge across an incompatible major, two copies would again be linked and the build would need a different remedy, as described in #1723.

### DIFF
--- a/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
+++ b/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
@@ -70,12 +70,18 @@ jobs:
     - name: Patch workspace dependencies
       working-directory: stellar-contracts
       run: |
-        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
-        for crate in $crates; do
-          rel_path=$(realpath --relative-to="." ../rs-soroban-sdk/$crate)
+        # Collect the publishable rs-soroban-sdk crates and their directories once,
+        # from the authoritative cargo metadata, deriving the path from manifest_path
+        # rather than assuming the package name matches the directory name. The list is
+        # reused by the "Patch transitive crates.io dependencies" step below.
+        (cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps) \
+          | jq -r '.packages[] | select(.publish != []) | "\(.name)\t\(.manifest_path | sub("/Cargo.toml$"; ""))"' \
+          > "$RUNNER_TEMP/sdk-crates.tsv"
+        while IFS=$'\t' read -r crate dir; do
+          rel_path=$(realpath --relative-to="." "$dir")
           sed -i 's|'"$crate"' = "\([^"]*\)"|'"$crate"' = { path = "'"$rel_path"'" }|g' Cargo.toml
           sed -i 's|'"$crate"' = { \(.*\)version = "[^"]*"\(.*\)|'"$crate"' = { \1path = "'"$rel_path"'" \2|g' Cargo.toml
-        done
+        done < "$RUNNER_TEMP/sdk-crates.tsv"
 
     - name: Enable experimental_spec_shaking_v2 feature
       if: matrix.experimental_spec_shaking_v2
@@ -95,13 +101,15 @@ jobs:
         # linked, producing duplicate `panic_impl` lang item errors (E0152, #1723).
         # A [patch.crates-io] table redirects transitive copies to the local checkout
         # (valid while the local version is semver-compatible with the required one).
-        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
-        echo "" >> Cargo.toml
-        echo "[patch.crates-io]" >> Cargo.toml
-        for crate in $crates; do
-          rel_path=$(realpath --relative-to="." ../rs-soroban-sdk/$crate)
-          echo "$crate = { path = \"$rel_path\" }" >> Cargo.toml
-        done
+        # Reuse the crate list collected in the "Patch workspace dependencies" step.
+        if [ -s "$RUNNER_TEMP/sdk-crates.tsv" ]; then
+          # Don't emit a second [patch.crates-io] header if one already exists upstream.
+          grep -q '^\[patch.crates-io\]' Cargo.toml || printf '\n[patch.crates-io]\n' >> Cargo.toml
+          while IFS=$'\t' read -r crate dir; do
+            rel_path=$(realpath --relative-to="." "$dir")
+            echo "$crate = { path = \"$rel_path\" }" >> Cargo.toml
+          done < "$RUNNER_TEMP/sdk-crates.tsv"
+        fi
 
     - name: Diff
       run: (! git diff --exit-code) || (echo 'A diff is expected'; exit 1)
--- a/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
+++ b/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
@@ -86,6 +86,23 @@ jobs:
         # Add features field to path-patched entries without features
         sed -i '/soroban-sdk = {.*path = /{/features/!s| }|, features = ["experimental_spec_shaking_v2"] }|}' Cargo.toml
 
+    - name: Patch transitive crates.io dependencies
+      working-directory: stellar-contracts
+      run: |
+        # Crates such as soroban-poseidon are pulled from crates.io and depend on
+        # soroban-sdk from crates.io. The [workspace.dependencies] path rewrite above
+        # cannot redirect those transitive copies, so two soroban-sdk crates get
+        # linked, producing duplicate `panic_impl` lang item errors (E0152, #1723).
+        # A [patch.crates-io] table redirects transitive copies to the local checkout
+        # (valid while the local version is semver-compatible with the required one).
+        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
+        echo "" >> Cargo.toml
+        echo "[patch.crates-io]" >> Cargo.toml
+        for crate in $crates; do
+          rel_path=$(realpath --relative-to="." ../rs-soroban-sdk/$crate)
+          echo "$crate = { path = \"$rel_path\" }" >> Cargo.toml
+        done
+
     - name: Diff
       run: (! git diff --exit-code) || (echo 'A diff is expected'; exit 1)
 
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> if they diverge across an incompatible major

Fwiw this is still relatively brittle, but it plugs the leak for now and we can think about how to make this more resilient if possible.

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -86,6 +86,23 @@ jobs:
...
+      run: |
+        # Crates such as soroban-poseidon are pulled from crates.io and depend on
+        # soroban-sdk from crates.io. The [workspace.dependencies] path rewrite above
+        # cannot redirect those transitive copies, so two soroban-sdk crates get
+        # linked, producing duplicate `panic_impl` lang item errors (E0152, #1723).
+        # A [patch.crates-io] table redirects transitive copies to the local checkout
+        # (valid while the local version is semver-compatible with the required one).
+        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
+        echo "" >> Cargo.toml
+        echo "[patch.crates-io]" >> Cargo.toml
```
leighmcculloch: **Bug: unconditional `[patch.crates-io]` append will break if OZ already has that section**

TOML forbids duplicate table headers. This step appends `[patch.crates-io]` unconditionally; if OpenZeppelin ever adds their own `[patch.crates-io]` to their workspace `Cargo.toml` (a perfectly normal thing to do), Cargo will fail with `Cannot declare ('patch', 'crates-io') twice` and every matrix job will break with a cryptic parse error.

The current OZ `main` has no such section, so this is latent — but the fix is straightforward:

```suggestion
        grep -q '^\[patch\.crates-io\]' Cargo.toml || echo "[patch.crates-io]" >> Cargo.toml
```

Or use a proper TOML-aware merge. At minimum the guard avoids silent corruption.

---

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -86,6 +86,23 @@ jobs:
...
+        for crate in $crates; do
+          rel_path=$(realpath --relative-to="." ../rs-soroban-sdk/$crate)
```
leighmcculloch: **Fragile: assumes crate package name == directory name**

`$crate` is the Cargo package name from `cargo metadata`, used directly as `../rs-soroban-sdk/$crate`. Cargo does not require the package name and directory name to match. If any crate is ever placed in a differently-named directory, `realpath` silently returns an exit 0 on a non-existent path, the wrong path ends up in `[patch.crates-io]`, and Cargo fails later with a confusing "no Cargo.toml found" error rather than a clear message here.

The same assumption exists in the "Patch workspace dependencies" step (line 75), so this inherits a pre-existing fragility — but it's worth fixing in both places. The robust approach is to derive the path from `manifest_path` in the metadata output:

```bash
crates_and_paths=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | \
  jq -r '.packages[] | select(.publish != []) | "\(.name)\t\(.manifest_path | sub("/Cargo.toml$";""))"')
```

Then use that to compute both the sed rewrites and the `[patch.crates-io]` entries from a single authoritative source.

---

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -86,6 +86,23 @@ jobs:
...
+      working-directory: stellar-contracts
+      run: |
+        # Crates such as soroban-poseidon are pulled from crates.io and depend on
+        # soroban-sdk from crates.io. The [workspace.dependencies] path rewrite above
+        # cannot redirect those transitive copies, so two soroban-sdk crates get
+        # linked, producing duplicate `panic_impl` lang item errors (E0152, #1723).
+        # A [patch.crates-io] table redirects transitive copies to the local checkout
+        # (valid while the local version is semver-compatible with the required one).
+        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
+        echo "" >> Cargo.toml
```
leighmcculloch: **Minor: section header is written before the loop, so it's appended even if `$crates` is empty**

If `cargo metadata` fails or unexpectedly returns no publishable packages, `$crates` is empty and the loop body never runs — but a bare `[patch.crates-io]` section is still appended. This is harmless on its own (valid TOML), but it compounds the duplicate-table risk from the comment on line 100: even a no-op run contaminates the file. Consider writing the header and entries atomically only when `$crates` is non-empty:

```bash
if [ -n "$crates" ]; then
  grep -q '^\[patch\.crates-io\]' Cargo.toml || echo "[patch.crates-io]" >> Cargo.toml
  for crate in $crates; do
    ...
  done
fi
```

---

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -86,6 +86,23 @@ jobs:
...
+    - name: Patch transitive crates.io dependencies
+      working-directory: stellar-contracts
+      run: |
+        # Crates such as soroban-poseidon are pulled from crates.io and depend on
+        # soroban-sdk from crates.io. The [workspace.dependencies] path rewrite above
+        # cannot redirect those transitive copies, so two soroban-sdk crates get
+        # linked, producing duplicate `panic_impl` lang item errors (E0152, #1723).
+        # A [patch.crates-io] table redirects transitive copies to the local checkout
+        # (valid while the local version is semver-compatible with the required one).
+        crates=$(cd ../rs-soroban-sdk && cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.publish != []) | .name')
```
leighmcculloch: **Cleanup: `cargo metadata` is called a second time (same command as line 73)**

The "Patch workspace dependencies" step (line 73) runs the exact same `cargo metadata --format-version 1 --no-deps | jq ...` invocation. `cargo metadata` spawns a full Cargo resolver pass — on a cold runner this can take 10–30 s and downloads the registry index. Running it twice per matrix cell is pure waste.

Since both steps run in sequence and share the same shell only within a single step, the most straightforward fix is to merge the two patching steps into one so the `$crates` variable is computed once and reused. An alternative is to write it to a temp file between steps.

---

leighmcculloch (commented):
Reviewed the new "Patch transitive crates.io dependencies" step. The approach is correct and the fix for #1723 makes sense. Four findings, ranked by severity:

1. **Duplicate `[patch.crates-io]` table** (line 100) — If OpenZeppelin ever adds a `[patch.crates-io]` section upstream, this creates invalid TOML and breaks every matrix job silently. Needs a guard before appending the header.

2. **Package name used as directory path** (line 102) — `../rs-soroban-sdk/$crate` assumes crate name == directory name; all 9 current crates match today, but this will silently produce wrong paths if that ever changes. Better to derive paths from `manifest_path` in the metadata JSON.

3. **Section header written unconditionally before the loop** (line 99) — If `$crates` is empty (e.g. metadata failure), a bare `[patch.crates-io]` is appended even though no entries follow, compounding issue 1.

4. **Redundant `cargo metadata` call** (line 98) — Identical to the call in the "Patch workspace dependencies" step; merging both patching steps into one would eliminate the duplicate and make the overall Cargo.toml transformation easier to reason about.

---

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
leighmcculloch: Took the temp-file alternative rather than merging the two steps. Merging would put the `[patch.crates-io]` append before the `Enable experimental_spec_shaking_v2` step, whose `sed` matches every `soroban-sdk = {.*path = ` line and would then inject `features = [...]` into the patch entry too — features in a `[patch]` entry are meaningless and would muddy the table. Keeping the patch append after that step preserves the ordering, so the crate list is now collected once into `$RUNNER_TEMP/sdk-crates.tsv` in the first step and reused here, with paths derived from `manifest_path`. That also resolves the package-name-vs-directory fragility in both places.

---

mootz12 (approved):
LG2M. We might want to consider adding this to stellar examples tests as well.

leighmcculloch:
@leighmcculloch Done — applied the same `[patch.crates-io]` transitive patch to the `Test with soroban-examples` workflow (`test-with-soroban-examples.yml`), mirroring the OpenZeppelin one: collect the crate list once from `manifest_path`, then append a guarded `[patch.crates-io]` table to the example's workspace-root `Cargo.toml` after the spec-shaking step. With transitive `soroban-sdk` now unified onto the local checkout, the `privacy-pools` example no longer hits the duplicate `panic_impl` error, so I removed its `exclude:` entry and it builds against the local SDK again.

Also folded in the optional nit from the re-review (escaped the `.` in the `^\[patch\.crates-io\]` guard) in both workflows since I was pushing anyway.

---

PR #1886, opened 2026-06-02
Add migration docs for functional changes to cfg attributes and export args

mootz12:
### What

Add migration docs for #1871 and #1876 

### Why

Both of these changes could impact existing contracts. The documentation provides guidance about how to migrate.

### Known limitations

None


soroban-sdk/src/_migrating/v27_contracttrait.rs
```diff
@@ -0,0 +1,123 @@
...
+//! therefore be evaluated in the implementing crate's configuration rather than the trait-defining
+//! crate's, so these attributes are now rejected on default functions.
+//!
+//! On `#[contractimpl(contracttrait)]` override methods, a direct `cfg` is supported — the default
+//! wrapper is generated under the inverse condition — but `cfg_attr` is rejected because it is not
+//! normalized by this handoff.
+//!
+//! ## `cfg`/`cfg_attr` on default functions
+//!
+//! `cfg` and `cfg_attr` are no longer accepted on the default functions of a [`contracttrait`]:
```
leighmcculloch: This statement makes it sound like `cfg` and `cfg_attr` were supported and fully functional, and now that functionality is being taken away. But my understanding was that these attributes never really worked before, and what's changed is an explicit compile error has been introduced to make it really clear to the user that they shouldn't use it, or it's not supported, and if they do, it's not going to work the way they expect. 

Am I understanding that correctly?

soroban-sdk/src/_migrating/v27_contracttrait.rs
```diff
@@ -0,0 +1,123 @@
...
+//! # fn main() {}
+//! ```
+//!
+//! To make a single function conditional, declare it without a default and gate the implementing
+//! method at each `#[contractimpl(contracttrait)]` site, where a direct `cfg` is supported (see
+//! below).
+//!
+//! ## `cfg_attr` on override methods
+//!
+//! On a `#[contractimpl(contracttrait)]` impl, a direct `cfg` is supported but `cfg_attr` is not:
```
leighmcculloch: A similar comment for this statement like the previous comment I left. Is this something that was truly supported before and now we're taking away that support, or are we making the compile errors clearer?

soroban-sdk/src/_migrating.rs
```diff
@@ -5,8 +5,29 @@
...
+//! 2. The `export` argument on [`contracttype`], [`contracterror`], and
+//!    [`contractevent`] is deprecated under the
+//!    [`experimental_spec_shaking_v2`][crate::_features#experimental_spec_shaking_v2]
+//!    feature. Under spec shaking v2 the final spec is determined by
+//!    reachability from the contract boundary, so `export` no longer has any
+//!    effect and now emits a deprecation warning. Remove `export = ...` from
+//!    these annotations when the feature is enabled; the argument will be
+//!    removed entirely in a future release. This affects only builds that opt
+//!    in to the experimental feature — default builds are unchanged. See
+//!    [`_features`][crate::_features#changed-behaviour] for details.
```
leighmcculloch: I think we should have a separate page for this one, like the others, and we should go into a little bit more detail about why export is not supported.

leighmcculloch (commented):
I really like the way this is written. I have a couple of asks inline.

soroban-sdk/src/_migrating/v27_contracttrait.rs
mootz12: Yes, this shouldn't be in migration docs. This was a bug fix - good call out.

leighmcculloch (approved):

PR #1887, opened 2026-06-03
Delete .github/workflows/claude-review.yml

leighmcculloch:
### What

Delete .github/workflows/claude-review.yml.

### Why

It needs some iteration. Fires too frequently. It's noisy and too verbose. There's no way to silence it on long iterative PRs. And we should iterate on the pull request target usage as well.

In the interim we've enabled Codex review, which does not have many of these downsides.

The shared claude-review workflow we can keep iterating on, and test it on one repo before bringing it to a bunch and creating too much noise:
- https://github.com/stellar/actions/pull/103


mootz12 (approved):
For the next attempt, it could be nice if this fires when requested. We could use the comment trigger:

> /claude-review {append additional prompt}

PR #1888, opened 2026-06-03
Zero Copy improvement: `BytesN::from` to use `BytesN::to_array`

dkcumming:
### What

Updated `BytesN::from` for `[u8; N]` to use `BytesN::to_array`.

### Why

This reduces the need to allocate memory and increases performance. This is benchmarked on [dkcumming/rs-soroban-sdk branch dc/bytes-from-zc-bench](https://github.com/stellar/rs-soroban-sdk/compare/main...dkcumming:rs-soroban-sdk:dc/bytes-from-zc-bench). For completeness the key end-to-end results are communicated in this table:

| Bench              | Before CPU | After CPU |    Δ % | Before Mem | After Mem |   Δ % |
|--------------------|-----------:|----------:|-------:|-----------:|----------:|------:|
| `baseline_*`       |    263,201 |   256,273 |  -2.6% |  1,158,730 | 1,154,758 | -0.3% |
| `to_array_32`      |    443,381 |   257,511 | -41.9% |  1,163,066 | 1,154,758 | -0.7% |
| `to_array_ref_32`  |    443,381 |   257,511 | -41.9% |  1,163,066 | 1,154,758 | -0.7% |
| `to_array_48`      |    533,025 |   258,573 | -51.5% |  1,165,618 | 1,154,758 | -0.9% |
| `to_array_ref_48`  |    533,025 |   258,573 | -51.5% |  1,165,618 | 1,154,758 | -0.9% |
| `to_array_96`      |    799,665 |   259,107 | -67.6% |  1,174,810 | 1,154,758 | -1.7% |
| `to_array_ref_96`  |    799,665 |   259,107 | -67.6% |  1,174,810 | 1,154,758 | -1.7% |
| `to_array_192`     |  1,334,673 |   260,175 | -80.5% |  1,200,106 | 1,154,758 | -3.8% |
| `to_array_ref_192` |  1,334,675 |   260,177 | -80.5% |  1,200,106 | 1,154,758 | -3.8% |

The tests to run these benchmarks were:

<details>

The contract:

```rust
#![no_std]
use soroban_sdk::{contract, contractimpl, BytesN, Env};

#[contract]
pub struct Contract;

// Each conversion returns a value derived from the array so the conversion
// cannot be optimized out of the guest Wasm.
//
// Each `to_array_N` has a matching `baseline_N` with the same signature, so
// that subtracting it accounts for dispatch overhead.

#[contractimpl]
impl Contract {
[...]

PR #1889, opened 2026-06-04
Test BytesN into [u8; N] conversions

leighmcculloch:
### What

Add unit tests exercising the `From<BytesN<N>> for [u8; N]` and `From<&BytesN<N>> for [u8; N]` conversions, covering the consuming form, the by-reference form (asserting the reference conversion does not consume the value), and the zero-length edge case.

### Why

These two public `From` conversions had no direct test coverage, and the `TryFrom<Bytes>/<&Bytes> for [u8; N]` impls that depend on them via `fixed.into()` were untested as well, so the `.into()` path had no regression guard. This gap surfaced while reviewing #1888, which rewrites both impls to delegate to `BytesN::to_array`; the tests pin the observable behaviour so that refactor (and future ones) can be verified to preserve it. They pass against both the current implementation and the one proposed in #1888.

### DIFF
--- a/soroban-sdk/src/tests/bytesn.rs
+++ b/soroban-sdk/src/tests/bytesn.rs
@@ -15,3 +15,27 @@ fn test_bytesn_is_empty_nonzero_length() {
     assert_eq!(b.len(), 4);
     assert!(!b.is_empty());
 }
+
+#[test]
+fn test_bytesn_into_array() {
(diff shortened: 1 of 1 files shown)


leighmcculloch (commented):
.

mootz12 (approved):

PR #1896, opened 2026-06-05
Update SDK to p27 and add CAP-71 auth delegation support.

dmkozh:
### What

Update SDK to p27 and add CAP-71 auth delegation support.

### Why

Adapting to p27 changes.

### Known limitations

I haven't added targeted tests for CAP-71 functions, as these are just simple wrappers over host functions. I think a full-fledged example in soroban-examples would be more appropriate.


soroban-sdk/src/address.rs
```diff
@@ -269,6 +269,41 @@ impl Address {
...
+    /// to the transaction via `Env::get_delegated_signers_for_current_auth_check`,
+    /// verify that these delegates actually belong to the account and are valid
+    /// for this transaction, and then call `delegate_account_auth` for each of
+    /// these delegates.
+    ///
+    /// ### Panics
+    ///
+    /// If called outside of `__check_auth`, or if this address has not
+    /// authorized the invocation.
+    pub fn delegate_account_auth(&self) {
```
leighmcculloch: This function needs a working example within the docs.

soroban-sdk/src/env.rs
```diff
@@ -381,6 +381,31 @@ impl Env {
...
+    /// Typical usage flow is to get all the delegated signers passed by the user
+    /// to the transaction via `Env::get_delegated_signers_for_current_auth_check`,
+    /// verify that these delegates actually belong to the account and are valid
+    /// for this transaction, and then call `delegate_account_auth` for each of
+    /// these delegates.
+    ///
+    /// ### Panics
+    ///
+    /// If called outside of `__check_auth`.
+    pub fn get_delegated_signers_for_current_auth_check(&self) -> Vec<Address> {
```
leighmcculloch: Similar needs a rust doc working example.

leighmcculloch: Can we add tests for using these functions along with a test vector contract?

soroban-sdk/src/env.rs
```diff
@@ -381,6 +381,31 @@ impl Env {
...
+    ///
+    /// Typical usage flow is to get all the delegated signers passed by the user
+    /// to the transaction via `Env::get_delegated_signers_for_current_auth_check`,
+    /// verify that these delegates actually belong to the account and are valid
+    /// for this transaction, and then call `delegate_account_auth` for each of
+    /// these delegates.
+    ///
+    /// ### Panics
+    ///
+    /// If called outside of `__check_auth`.
```
leighmcculloch: 🤔 I'm pondering how we can structure these functions so they are only callable within the check auth function. cc @mootz12

leighmcculloch (commented):
I've left some comments inline from a cursory look over and will do a deeper relook next week.

soroban-sdk/src/address.rs
dmkozh: I don't think so. We don't really have docs examples for most of the host functions, and for these specific ones the examples would be too complex. I would much rather prefer making a proper e2e example, as short code snippets don't really do the justice here.

soroban-sdk/src/env.rs
dmkozh: Same as above, I don't think that makes much sense. Furthermore, we might need to add more testutils for this. I can look more into that, but I don't think it has to be done in this PR.

dmkozh: I'm not sure what these tests would cover. It would make more sense if we introduce the test utils, but for now there isn't much to test.

soroban-sdk/src/address.rs
```diff
@@ -269,6 +269,41 @@ impl Address {
         self.env.require_auth(self);
     }
 
+    /// Delegates the current `__check_auth` authorization to this address.
+    ///
+    /// This may only be called from the `__check_auth` function of a custom
+    /// account contract.
```
leighmcculloch: Suggest tweak, because this comment is in a Rust context, and the function can be called outside of the __check_auth Rust function as long as it is inside the same invocation frame. There may be a better way to word what I suggest here.
```suggestion
    /// This may only be called from within the contract invocation that called the `__check_auth` function of a custom
    /// account contract.
```

soroban-sdk/src/address.rs
```diff
@@ -269,6 +269,41 @@ impl Address {
...
+    /// `Env::get_delegated_signers_for_current_auth_check` instead of
+    /// `require_auth` when implementing modular custom accounts. These accounts
+    /// don't have to perform the authentication themselves, but instead can
+    /// delegate the authentication logic to a different address (G- or C-) that
+    /// performs the actual authentication logic. The delegate may rely on its
+    /// own delegates as well, i.e. delegation can be nested recursively if
+    /// necessary.
+    ///
+    /// Typical usage flow is to get all the delegated signers passed by the user
+    /// to the transaction via `Env::get_delegated_signers_for_current_auth_check`,
```
leighmcculloch: Afaiu not all signers for the whole transaction get passed.
```suggestion
    /// Typical usage flow is to get all the delegated signers passed by the user
    /// to the invocation via `Env::get_delegated_signers_for_current_auth_check`,
```

soroban-sdk/src/env.rs
leighmcculloch: We can't merge this without tests, something using the functions and asserting something about their affect.

At the very least a test can look like the example I posted at https://github.com/stellar/rs-soroban-sdk/pull/1896/changes#r3371611207 which does a pretty good job of testing the function.

Ideally tests have better mechanism for asserting on the delegated address that authorised, but that can come later and an almost-the-same assertion can be made like in the example with what we have today where you treat the delegate like a 'spy' and record what it sees.

We need both unit tests in the soroban-sdk's test/ folder, and a test vector contract.

soroban-sdk/src/address.rs
leighmcculloch: Can this function be renamed to `delegate_auth`? The account is superfluous and creates a disjoint vocab/narrative because the same term is not used in the `require_auth` fn when both apply to the same scope of addresses.

leighmcculloch: We have examples for most auth related functions due to the complexity and difficulty in understanding how to use them. See https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Env.html.

The size of the example is a good hint we should have an example. We can of course have larger examples in docs too, but an example here both captures the flow and shows how the function should be used in the Rust docs. And it's compiled and run on every build too verifying it works.

```suggestion
    ///
    /// ### Examples
    ///
    /// A "modular" custom account that performs no authentication itself, and
    /// instead delegates it to a set of registered signer addresses. The user
    /// chooses which of the registered signers to authenticate with, by
    /// attaching them to the transaction as delegated signers.
    ///
    /// ```
    /// use soroban_sdk::{
    ///     auth::{Context, CustomAccountInterface},
    ///     contract, contracterror, contractimpl, contracttype,
    ///     crypto::Hash, vec, Address, Env, Vec,
    /// };
    ///
    /// #[contracterror]
    /// #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
    /// #[repr(u32)]
    /// pub enum Error {
    ///     UnknownDelegate = 1,
    /// }
    ///
    /// #[contracttype]
    /// pub enum DataKey {
    ///     Signers,
    ///     ApprovedContexts,
    /// }
    ///
    /// #[contract]
    /// pub struct ModularAccount;
    ///
    /// #[contractimpl]
    /// impl ModularAccount {
    ///     // Registers the addresses allowed to authenticate for this account.
    ///     pub fn __constructor(env: Env, signers: Vec<Address>) {
    ///         env.storage().instance().set(&DataKey::Signers, &signers);
    ///     }
    /// }
    ///
    /// #[contractimpl]
    /// impl CustomAccountInterface for ModularAccount {
    ///     // The account verifies no signature of its own, so it carries no
    ///     // signature to check.
    ///     type Signature = ();
    ///     type Error = Error;
    ///
    ///     fn __check_auth(
    ///         env: Env,
    ///         _signature_payload: Hash<32>,
    ///         _signatures: (),
    ///         _auth_contexts: Vec<Context>,
    ///     ) -> Result<(), Error> {
    ///         // The signers the user attached to the transaction for this
    ///         // account's authorization.
    ///         let delegates = env.get_delegated_signers_for_current_auth_check();
    ///
    ///         // The host does not validate the delegates, so the account must
    ///         // check that each one is actually a registered signer before
    ///         // trusting it.
    ///         let signers: Vec<Address> =
    ///             env.storage().instance().get(&DataKey::Signers).unwrap();
    ///         for delegate in delegates.iter() {
    ///             if !signers.contains(&delegate) {
    ///                 return Err(Error::UnknownDelegate);
    ///             }
    ///             // Forward the current authorization to the delegate. Its own
    ///             // authentication is checked as if it had required auth
    ///             // directly, but without needing a separate auth entry.
    ///             delegate.delegate_account_auth();
    ///         }
    ///         Ok(())
    ///     }
    /// }
    ///
    /// // A delegate account. Stands in for any address (a Stellar account or
    /// // another contract) that knows how to authenticate itself. Here it
    /// // records the context it approved so the test can verify the delegation
    /// // reached it.
    /// #[contract]
    /// pub struct DelegateAccount;
    ///
    /// #[contractimpl]
    /// impl CustomAccountInterface for DelegateAccount {
    ///     type Signature = ();
    ///     type Error = Error;
    ///     fn __check_auth(
    ///         env: Env,
    ///         _signature_payload: Hash<32>,
    ///         _signatures: (),
    ///         auth_contexts: Vec<Context>,
    ///     ) -> Result<(), Error> {
    ///         env.storage()
    ///             .instance()
    ///             .set(&DataKey::ApprovedContexts, &auth_contexts);
    ///         Ok(())
    ///     }
    /// }
    ///
    /// // A contract with an operation that requires the account's authorization.
    /// #[contract]
    /// pub struct Protected;
    ///
    /// #[contractimpl]
    /// impl Protected {
    ///     pub fn protected(account: Address) {
    ///         account.require_auth();
    ///     }
    /// }
    ///
    /// #[test]
    /// fn test() {
    /// # }
    /// # #[cfg(feature = "testutils")]
    /// # fn main() {
    ///     use soroban_sdk::xdr::{
    ///         InvokeContractArgs, ScAddress, ScVal, SorobanAddressCredentials,
    ///         SorobanAddressCredentialsWithDelegates, SorobanAuthorizationEntry,
    ///         SorobanAuthorizedFunction, SorobanAuthorizedInvocation,
    ///         SorobanCredentials, SorobanDelegateSignature, StringM, VecM,
    ///     };
    ///
    ///     let env = Env::default();
    ///     let delegate = env.register(DelegateAccount, ());
    ///     // Register the modular account with `delegate` as an allowed signer.
    ///     let account = env.register(ModularAccount, (vec![&env, delegate.clone()],));
    ///     let protected = env.register(Protected, ());
    ///
    ///     let account_addr: ScAddress = account.clone().try_into().unwrap();
    ///     let delegate_addr: ScAddress = delegate.clone().try_into().unwrap();
    ///
    ///     // This authorization entry is normally built by the user's
    ///     // wallet/tooling and attached to the transaction. It authorizes
    ///     // `protected` on behalf of the account, and attaches `delegate` as a
    ///     // delegated signer. Delegates must be sorted by address with no
    ///     // duplicates.
    ///     env.set_auths(&[SorobanAuthorizationEntry {
    ///         credentials: SorobanCredentials::AddressWithDelegates(
    ///             SorobanAddressCredentialsWithDelegates {
    ///                 address_credentials: SorobanAddressCredentials {
    ///                     address: account_addr.clone(),
    ///                     nonce: 1,
    ///                     signature_expiration_ledger: 100,
    ///                     // The account verifies no signature of its own.
    ///                     signature: ScVal::Void,
    ///                 },
    ///                 delegates: std::vec![SorobanDelegateSignature {
    ///                     address: delegate_addr,
    ///                     signature: ScVal::Void,
    ///                     nested_delegates: VecM::default(),
    ///                 }]
    ///                 .try_into()
    ///                 .unwrap(),
    ///             },
    ///         ),
    ///         root_invocation: SorobanAuthorizedInvocation {
    ///             function: SorobanAuthorizedFunction::ContractFn(InvokeContractArgs {
    ///                 contract_address: protected.clone().try_into().unwrap(),
    ///                 function_name: StringM::try_from("protected").unwrap().into(),
    ///                 args: std::vec![ScVal::Address(account_addr)].try_into().unwrap(),
    ///             }),
    ///             sub_invocations: VecM::default(),
    ///         },
    ///     }]);
    ///
    ///     // The call succeeds: the account delegates its authentication to
    ///     // `delegate`, which approves it.
    ///     ProtectedClient::new(&env, &protected).protected(&account);
    ///
    ///     // The account authorized the `protected` call. Delegating to
    ///     // `delegate` is not recorded as a separate authorization.
    ///     use soroban_sdk::{
    ///         auth::ContractContext,
    ///         testutils::{AuthorizedFunction, AuthorizedInvocation},
    ///         IntoVal, Symbol,
    ///     };
    ///     assert_eq!(
    ///         env.auths(),
    ///         std::vec![(
    ///             account.clone(),
    ///             AuthorizedInvocation {
    ///                 function: AuthorizedFunction::Contract((
    ///                     protected.clone(),
    ///                     Symbol::new(&env, "protected"),
    ///                     (account.clone(),).into_val(&env),
    ///                 )),
    ///                 sub_invocations: std::vec![],
    ///             }
    ///         )]
    ///     );
    ///
    ///     // The delegation actually reached `delegate`, which approved the
    ///     // same invocation that was authorized above.
    ///     let approved: Vec<Context> = env.as_contract(&delegate, || {
    ///         env.storage().instance().get(&DataKey::ApprovedContexts).unwrap()
    ///     });
    ///     assert_eq!(approved.len(), 1);
    ///     match approved.get(0).unwrap() {
    ///         Context::Contract(ContractContext { contract, fn_name, args }) => {
    ///             assert_eq!(contract, protected);
    ///             assert_eq!(fn_name, Symbol::new(&env, "protected"));
    ///             assert_eq!(args, (account.clone(),).into_val(&env));
    ///         }
    ///         _ => panic!("expected a contract invocation context"),
    ///     }
    /// }
    /// # #[cfg(not(feature = "testutils"))]
    /// # fn main() { }
    /// ```
    pub fn delegate_account_auth(&self) {
```

soroban-sdk/src/env.rs
leighmcculloch: This function can reference an example on the other method.
```suggestion
    ///
    /// ### Examples
    ///
    /// See [`Address::delegate_account_auth`] for a complete, runnable example
    /// of a modular custom account that retrieves its delegated signers with
    /// this function, validates them, and delegates authentication to them.
    pub fn get_delegated_signers_for_current_auth_check(&self) -> Vec<Address> {
```

leighmcculloch (dismissed):
Few changes requested inline.

I've included inline a working example that can be included, an example doc for the second function that references the first, and the same pattern can be used for a unit test and a test vector.

Let me know if I can help with the unit tests and test vector.

soroban-sdk/src/env.rs
leighmcculloch: If there was a type for the Signature that wasn't generic we could attach it to that, but without that there's really no where to attach it within the __check_auth implementation.

So disregarding this.

soroban-sdk/src/env.rs
leighmcculloch: Can we remove the suffix? It seems redundant in this function. Everything is in the context of the current invocation. cc @mootz12 
```suggestion
    pub fn get_delegated_signers(&self) -> Vec<Address> {
```

soroban-sdk/src/address.rs
mootz12: (picky) This is still a little unclear to me. This reads like I could call this host function at any point, as I don't really have control on when `__check_auth` is invoked. I just got `require_auth`'d by something else.

maybe "This may only be called during `__check_auth`."

I'd also remove `of a custom account contract`, cause any contract could opt to `delegate_account_auth`. Smart account is just the one this was designed for.

soroban-sdk/src/address.rs
dmkozh: The length of the example is actually the reason for why I don't think it's a good idea to embed it into docs. 
The example is also not comprehensive and OTOH has a lot of noise (like contract errors, unrelated comments etc.). That all would be relevant to a proper e2e example, but in the docs it seems inappropriate to me.
I could give a more comprehensive example a shot. We could embed it into docs if you insist, but my opinion stays the same - it's seemingly too much info to digest for a function doc.

dmkozh: It's supposed to be different from `require_auth` because it is only usable in scope of account auth. I don't have too strong opinion here though, just wanted to double-check if we're on the same page here.

soroban-sdk/src/env.rs
dmkozh: Same reason as with delegate_account_auth - this must be in the context of the auth check, not just an arbitrary invocation. I don't mind the rename, but it would be more appealing if we had a better way of scoping this function to check_auth. But since we don't have such way, it's seemingly the best we can do for now.

soroban-sdk/src/address.rs
mootz12: +1 to `delegate_auth`

By "only usable in scope of account auth", we just mean within `__check_auth` right? IMO keeping the naming consistent makes it clearer what this is doing.

soroban-sdk/src/env.rs
mootz12: I agree. It's in context of the `__check_auth` process already.

soroban-sdk/src/address.rs
dmkozh: > I'd also remove of a custom account contract, cause any contract could opt to delegate_account_auth. Smart account is just the one this was designed for.

That's not true, is it? Contract with `__check_auth` == 'custom account contract'.

soroban-sdk/src/env.rs
dmkozh: How exactly is it in the context of `__check_auth` though?

soroban-sdk/src/address.rs
mootz12: Sure, it's just a minor wording nit. Point is that 'custom account contract' is not really defined, so saying it can only be used in that scope could be confusing.

A DAO/treasury could use `__check_auth` to:
- Validate some proposal passed that aligns with the call context
- Delegate auth to some executor role for the DAO

If we are consistent with the wording it's not a big deal.

dmkozh: In my mind 'custom account contract' is a contract with `__check_auth` defined (which in turn means that it can handle externally provided auth, as opposed to the regular invoker-based contract auth), but if there is no consensus on this definition I guess we could drop it.

dmkozh: Anyways, reworded this a bit.

soroban-sdk/src/address.rs
dmkozh: True, but invocation is kind of ambiguous as well (like it could be treated as e.g. an input argument). I've used auth entry instead, which is more precise than either option.

soroban-sdk/src/address.rs
dmkozh: I've renamed it for now, though I still feel like it's a bit confusing.

soroban-sdk/src/env.rs
dmkozh: Anyways, renamed this for now.

soroban-sdk/src/address.rs
mootz12: I suppose I'd lean to however we named the host function for now, and just make it consistent from that.

soroban-sdk/src/env.rs
mootz12: > it would be more appealing if we had a better way of scoping this function to check_auth

Hm, I assumed if your call stack didn't have `__check_auth` on it, this function would fail. So I assumed if you called this function you'd be within that scope.

We could also just remove "current" -> `get_delegated_signers_for_auth_check`. That detail can just exist in the function docs.

soroban-sdk/src/address.rs
dmkozh: I've added a test. It's pretty large (300+ LOC). It could be trimmed a bit (by removing a negative scenario), but I still feel like it's too much for the function doc.

soroban-sdk/src/env.rs
dmkozh: I've added a test, not sure about a test vector contract though - what exactly do you want to be covered?

soroban-sdk/src/address.rs
leighmcculloch: Auth entry is better, thanks.

soroban-sdk/src/env.rs
dmkozh: > Hm, I assumed if your call stack didn't have __check_auth on it, this function would fail. So I assumed if you called this function you'd be within that scope.

Sure, but the question was rather if the name should convey the fact that this function will not work outside of __check_auth at all (to be clear, the requirement is not just have it on call stack, it's being strictly in the __check_auth frame). I'm not sure if there is a great to make this more obvious statically, for host there is really no other way but to use the name (because we just expose a huge list of functions), but for the SDK we might be able to do something else.

soroban-sdk/src/address.rs
leighmcculloch: I don't disagree with the confusion, in general of how auth works, but I don't think the `_account_` makes a measurable impact to that.

soroban-sdk/src/tests/delegate_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+        auth_contexts: Vec<Context>,
+    ) -> Result<(), ModularAccountError> {
+        // Even though we use delegated authentication, the account can still
+        // perform the regular verification if necessary.
+        let public_key: BytesN<32> = env.storage().instance().get(&DataKey::PublicKey).unwrap();
+        env.crypto()
+            .ed25519_verify(&public_key, &signature_payload.into(), &signature);
+        record_authorized_calls(&env, &auth_contexts);
+
+        // The signers the user attached to the transaction for this account's
```
leighmcculloch:
```suggestion
        // The signers the user attached to the auth entry for this account's
```

soroban-sdk/src/env.rs
```diff
@@ -381,6 +381,31 @@ impl Env {
...
+    /// Typical usage flow is to use `Env::get_delegated_signers` to get all the
+    /// delegated signers passed by the user inside the matching auth entry, then
+    /// verify that these delegates actually belong to the account and are valid
+    /// for this transaction, and then call `delegate_auth` for each of
+    /// these delegates.
+    ///
+    /// ### Panics
+    ///
+    /// If called outside of `__check_auth`.
+    pub fn get_delegated_signers(&self) -> Vec<Address> {
```
leighmcculloch: In the XDR the values returned by this function are stored as part of the credential, along with a signature, and referred to as 'delegates'. But here we refer to them as 'delegated signers'. Which is similar but it's not a consistent vocab 🤔 

I'm thinking about the names of things to try and tease apart a clearer relationship between the XDR that a developer constructs and looks at and the functions it marries up with in contracts. As well as to make it more or less obvious when the functions can be called.

An auth entry has credentials, the credential is an address and a signature. In the case of delegation, the credential is an address, signature, and a list of delegates, that also contain their signature.

It occurs to me now we could have designed delegates so that they were passed as a function parameter, alongside signatures. It would have been more consistent with the other parameters that get passed to __check_auth, or maybe a __check_auth_with_delegates.

It's too late to do that at the host level, but we could still do that in the SDK. The SDK macros are capable of detecting the presence of the delegate parameter, and fill it by calling the host function. In this way the SDK doesn't have a dangling global function for something with narrow use and the input parameters in the XDR are still grouped together in the same way in the contract code.

```
impl CustomAccountInterfaceWithDelegates for ModularAccount {
    type Signature = BytesN<64>;
    type Error = ModularAccountError;

    fn __check_auth(
        env: Env,
        signature_payload: Hash<32>,
        signature: BytesN<64>,
        delegates: Vec<Address>,
        auth_contexts: Vec<Context>,
    ) -> Result<(), ModularAccountError> {
```

Thoughts?

leighmcculloch (commented):
Continuing to play around with and think about this and have posted some more thoughts inline.

soroban-sdk/src/env.rs
dmkozh: > It occurs to me now we could have designed delegates so that they were passed as a function parameter, alongside signatures

Yeah, I haven't considered that unfortunately, but OTOH it seems like it could be a bit too disruptive for the protocol layer. It would break the existing contracts even if empty delegates are passed which might make migration annoying for those who are interested in using this. 

> The SDK macros are capable of detecting the presence of the delegate parameter, and fill it by calling the host function.

This is a good idea conceptually, though I'm not sure how it will work in practice. If we can implement this in a non-confusing way that doesn't mess up contract specs, testing etc., then it seems like a fine option to go with.

dmkozh: A less elegant, but also less complex option would be to just introduce a wrapper struct for the custom-account-only functions (similar to `deployer`) that exposes the functions that are only meant for the custom accounts.

mootz12: > It occurs to me now we could have designed delegates so that they were passed as a function parameter, alongside signatures

I agree this would probably be better.

> It's too late to do that at the host level, but we could still do that in the SDK. The SDK macros are capable of detecting the presence of the delegate parameter, and fill it by calling the host function.

While I think the interface is better, this does perform a host call that the contract writer might not need. Not sure I am a huge fan of that.

> A less elegant, but also less complex option would be to just introduce a wrapper struct for the custom-account-only functions (similar to `deployer`) that exposes the functions that are only meant for the custom accounts.

This feels like the correct compromise. Suggesting `custom_account` or `delegate(s)` as the wrapper name.

dmkozh: >  this does perform a host call that the contract writer might not need. 

That's not true, is it? If you care about the delegates, you have to make a host call. If you don't care, you don't use the new interface.

mootz12: But nobody would _expect_ a different interface to make an additional host call.

I realize they need to if they want this to work (they do it in their code), but I think making host calls without the user explicitly writing them feels weird. I don't think we have done that anywhere else in the SDK.

dmkozh: >  I don't think we have done that anywhere else in the SDK.

We actually do that quite a lot, SDK abstracts away host and it doesn't have any specific contract w.r.t. host calls performed. E.g. parsing a `contracttype` results in an arbitrary number of the host calls, so any time you call a function with UDT in the signature, you do a bunch of host calls. Some SDK functions do several calls (like storage functions that do has+get calls). So I don't think it's a big deal (not to mention that the actual costs are negligible).

mootz12: I suppose IMO those feel expected. But I see your point.

Regardless, I still think your option of the wrapper-struct for these functions is my preference.

If we want to extend the interface with a new ` __check_auth_with_delegates` version, this would not be blocked. And, we can document it just gets converted to:
```rust
fn __check_auth(
        env: Env,
        signature_payload: Hash<32>,
        signature: BytesN<64>,
        auth_contexts: Vec<Context>,
    ) -> Result<(), ModularAccountError> {
        let delegates = env.custom_account().get_delegated_signers();
        {code}
}
```

dmkozh: Alright, I agree that it's not worth blocking this PR on the additional interface, I've introduced the accessor for now.

leighmcculloch: I don't think we should do what I suggested earlier because it misaligns the rust fn name and/or parameters with the actual name/parameters. Moving the fns to a group so it isn't on the top-level env sounds good to me.

leighmcculloch (commented):
Looks good, except I think a file is missing. I don't see the `custom_account.rs` file committed.

dmkozh:
> Looks good, except I think a file is missing. I don't see the custom_account.rs file committed.

Sorry, fixed

mootz12 (approved):

soroban-sdk/src/address.rs
dmkozh: I'll close this for now - we can update the docs as a followup if necessary.

PR #1905, opened 2026-06-10
Filter empty wasm hash from snapshot lookups

### LINKED ISSUE #1635: Empty WASM hash leaks to SnapshotSource and test snapshot files

soroban-sdk v23.2.1 (and prior versions)

### What did you do?

When registering a native test contract with `Env::register()`, the Host requests entries from the `SnapshotSource`, including a `ContractCode` entry with the empty WASM hash (`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` - the SHA256 of zero bytes).

This can be observed by implementing a custom `SnapshotSource` that logs all `get` requests:

```rust
struct LoggingSnapshotSource;

impl SnapshotSource for LoggingSnapshotSource {
    fn get(&self, key: &Rc<LedgerKey>) -> Result<Option<(Rc<LedgerEntry>, Option<u32>)>, HostError> {
        eprintln!("Attempted to get ledger key: {:?}", key);
        Ok(None)
    }
}

#[test]
fn test() {
    let e = Env::from_ledger_snapshot(LoggingSnapshotSource);
    let contract_id = e.register(Contract, ());
    // ...
}
```

Output:
```
Attempted to get ledger key: ContractCode(LedgerKeyContractCode { hash: Hash(e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855) })
Attempted to get ledger key: ContractData(LedgerKeyContractData { contract: Contract(...), key: LedgerKeyContractInstance, durability: Persistent })
```

Additionally, the empty WASM hash entry appears in test snapshot JSON files written to disk:

```json
{
  "contract_code": {
    "hash": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
  }
}
```

The empty WASM hash is an implementation detail for native test contracts. It should be:
1. **Filtered from `SnapshotSource` requests** - The SDK should intercept requests for the empty WASM hash and return `None` without forwarding to the underlying source
2. **Excluded from ledger snapshots** - The empty WASM code entry should not appear in snapshot JSON files written to disk

### What did you see instead?

The empty WASM hash leaks through both interfaces:
- Custom `SnapshotSource` implementations receive requests for it
- Test snapshot files contain the empty WASM entry

## Why This Is Problematic

The `SnapshotSource` trait enables powerful use cases like loading ledger state from external sources - RPC endpoints, network archives, or other remote data stores. These sources may involve:

- **Network requests** that add latency
- **Rate-limited APIs** where each request counts
- **Expensive lookups** across distributed systems

When the Host requests the empty WASM hash from a `SnapshotSource`, it forces these implementations to make a request for an entry that **will never exist** on any real network. This is a waste of resources and adds unnecessary latency to test setup.

The empty WASM hash is purely an internal implementation detail of how the Host and SDK handle native test contracts. Contract developers implementing custom `SnapshotSource`s should not need to know about or handle this edge case - it should be invisible to them.

Similarly, the empty WASM entry appearing in snapshot JSON files is confusing noise. These files may be checked into source control, shared between teams, or used for debugging. The empty WASM entry adds clutter and raises questions about what it is and why it exists.

## Current Behavior

```mermaid
sequenceDiagram
    participant Test
    participant SDK as Soroban SDK
    participant Host as Soroban Host
    participant SS as SnapshotSource
    participant Net as Network/RPC

    Test->>SDK: Env::from_ledger_snapshot(source)
    SDK->>Host: Create Host with SnapshotSource
    Test->>SDK: env.register(Contract, ())
    SDK->>Host: Register native contract
    Host->>SS: get(ContractCode{hash: empty})
    SS->>Net: Request empty WASM (wasteful!)
    Net-->>SS: Not found
    SS-->>Host: None
    Host->>SS: get(ContractData{instance})
    SS->>Net: Request contract instance
    Net-->>SS: Entry or not found
    SS-->>Host: Result
    
    Note over Test,Net: Later, when saving snapshot...
    
    Test->>SDK: env.to_ledger_snapshot()
    SDK->>Host: get_stored_entries()
    Host-->>SDK: All entries including empty WASM
    SDK-->>Test: LedgerSnapshot with empty WASM entry
```

## Proposed Behavior

```mermaid
sequenceDiagram
    participant Test
    participant SDK as Soroban SDK
    participant Host as Soroban Host
    participant Wrapper as FilteringSnapshotSource
    participant SS as SnapshotSource
    participant Net as Network/RPC

    Test->>SDK: Env::from_ledger_snapshot(source)
    SDK->>SDK: Wrap source in FilteringSnapshotSource
    SDK->>Host: Create Host with FilteringSnapshotSource
    Test->>SDK: env.register(Contract, ())
    SDK->>Host: Register native contract
    Host->>Wrapper: get(ContractCode{hash: empty})
    Wrapper-->>Host: None (intercepted, not forwarded)
    Host->>Wrapper: get(ContractData{instance})
    Wrapper->>SS: get(ContractData{instance})
    SS->>Net: Request contract instance
    Net-->>SS: Entry or not found
    SS-->>Wrapper: Result
    Wrapper-->>Host: Result
    
    Note over Test,Net: Later, when saving snapshot...
    
    Test->>SDK: env.to_ledger_snapshot()
    SDK->>Host: get_stored_entries()
    Host-->>SDK: All entries including empty WASM
    SDK->>SDK: Filter out empty WASM entry
    SDK-->>Test: LedgerSnapshot without empty WASM entry
```

## Proposed Solution

1. **Wrap the `SnapshotSource`**: In `Env::new_for_testutils`, wrap the provided `SnapshotSource` in a filtering wrapper that intercepts requests for `ContractCode` entries with the empty WASM hash and returns `None` immediately without forwarding to the underlying source.

2. **Filter `to_ledger_snapshot()` output**: When generating a `LedgerSnapshot` via `to_ledger_snapshot()`, filter out entries with the empty WASM hash before returning.

## Why not fix this in the Host?

The Host exposes its raw storage to the SDK via `get_stored_entries()`. Even if the Host handled the `SnapshotSource` side internally, the SDK would still need to filter the entries when generating snapshots. Handling both in the SDK keeps the solution contained in one place.

It would be possible to fix this entirely in the Host, but it would require either not exposing the full storage to the SDK or changing how storage is exposed to the SDK. While doable, this would increase the amount of specialized test code in the Host, which is also not desirable.

leighmcculloch:
### What

Wrap the snapshot source used by the test environment so that read-throughs for the ContractCode entry of the empty WASM hash (the SHA256 of zero bytes that the host uses as the synthetic wasm hash for native test contracts) are short circuited and return None without ever reaching the underlying source.

### Why

When registering a native test contract with `Env::register()`, the host reads a ContractCode entry for the empty WASM hash, and because the test environment uses recording-footprint storage that read falls through to the configured `SnapshotSource`. For custom `SnapshotSource` implementations that fetch ledger state from remote or RPC endpoints, this forces a request for an entry that can never exist on any real network, adding latency and consuming rate-limited lookups during test setup for what is purely an internal implementation detail of native test contracts.

### Known limitations

The empty WASM ContractCode entry is intentionally left in saved ledger snapshots, so `to_ledger_snapshot()` output is unchanged. A persisted native contract instance still references that hash, so removing the entry would break reloading the snapshot, and keeping it matches how every other contract carries a corresponding code entry. The fix only suppresses the redundant `SnapshotSource` lookup, which is safe because native contracts dispatch through their registered functions rather than the wasm code, so the empty code entry is never executed on reload.

Close #1635


leighmcculloch:
Solving this problem will become more important once we had capabilities like:
- https://github.com/stellar/rs-soroban-sdk/issues/1945

Nothing needs changing today to support that, but this filtering snapshot source added in here will probably as part of that issue expand to filter out all natively installed contracts.

dmkozh (approved):

PR #1908, opened 2026-06-14
Add external read-only storage SDK API

mihaieremia:
Discussion: https://github.com/stellar/stellar-protocol/discussions/1958

### What

Adds `next` SDK APIs for read-only access to another contract's storage:

- `env.storage().persistent().has_external(&contract, &key)`
- `env.storage().persistent().get_external::<_, V>(&contract, &key)`
- `env.storage().temporary().has_external(&contract, &key)`
- `env.storage().temporary().get_external::<_, V>(&contract, &key)`
- `env.storage().instance().has_external(&contract, &key)`
- `env.storage().instance().get_external::<_, V>(&contract, &key)`

The WASM path imports the protocol-28 host ABI directly. Optional reads use `try_get_external_contract_data` so a stored `Void` remains distinguishable from a missing value. Local storage `get` also uses the new optional host path under `next` for the same reason.

This branch intentionally stays on the current upstream SDK/env/XDR package versions and only adds `next` feature hooks. It is not an official protocol release/version bump.

### Why

This lets contracts read intentionally public state from another contract without a cross-contract view invocation.

For lending, this enables a cleaner split where a governance, market, or oracle registry contract owns audit-critical read-only state such as market configuration, oracle references, reserve factors, pause flags, and collateral factors. The controller contract can execute the lending flow and read those keys directly, which helps keep controller and pool WASMs smaller while preserving modular contracts that are easier to audit.

Expected benefits from the host PR benchmark are lower CPU, memory, and wall-time cost than invoking a target contract just to return the same stored value.

### Known limitations

- Draft SDK PR; it depends on the protocol-28 env ABI PR before the imports are usable on-chain.
- APIs are behind `next` and are not for protocol <= 27 contracts.
- Native SDK test execution of these methods currently panics because the unreleased protocol-28 host ABI is not available through the published env dependency. The WASM import path is the intended path for this PR.
- External instance reads load the target contract instance entry and inherit the host-side cost behavior of decoding that instance map.
- No schema, permission, auth, write, or TTL-extension APIs are added.

### Validation

- `cargo check -p soroban-sdk`
- `cargo check -p soroban-sdk --features next`
- `cargo test -p soroban-sdk --features next,testutils --test external_storage_next`
- `cargo fmt -p soroban-sdk -p soroban-sdk-macros -- --check`
- `git diff --check`

### DIFF
--- a/soroban-sdk-macros/Cargo.toml
+++ b/soroban-sdk-macros/Cargo.toml
@@ -29,5 +29,6 @@ sha2 = "0.10.7"
 heck = "0.5.0"
 
 [features]
+next = ["soroban-env-common/next"]
 testutils = []
 experimental_spec_shaking_v2 = []
--- a/soroban-sdk/Cargo.toml
+++ b/soroban-sdk/Cargo.toml
@@ -66,6 +66,7 @@ trybuild = "1.0.115"
 
 [features]
 alloc = []
+next = ["soroban-sdk-macros/next", "soroban-env-guest/next", "soroban-env-host/next"]
 testutils = ["soroban-sdk-macros/testutils", "soroban-env-host/testutils", "soroban-ledger-snapshot/testutils", "dep:ed25519-dalek", "dep:arbitrary", "dep:derive_arbitrary", "dep:ctor", "dep:soroban-ledger-snapshot"]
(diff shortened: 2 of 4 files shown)


tupui:
This would basically allow us to not have the need to make getters for everything 👍

willemneal:
@tupui once we standardized exporting storage types and layout.

PR #1914, opened 2026-06-17
Fix External CI tests so they run even for major version bumps

leighmcculloch:
### What

Add a step to both external-contract CI workflows ("Test with OpenZeppelin Contracts" and "Test with soroban-examples") that relabels the locally checked-out rs-soroban-sdk crates to the soroban-sdk version the destination already resolves, running before the existing path-patching so the `[patch.crates-io]` redirect collapses every transitive soroban-sdk edge onto the local checkout even when the local crates have bumped to a newer major. With the relabel in place, also drop the privacy-pools matrix exclusion in the soroban-examples workflow, since that example is now buildable against the local SDK.

### Why

Both workflows build external contracts against the local soroban-sdk by rewriting direct dependency edges to path overrides and adding a `[patch.crates-io]` table, but Cargo only applies a patch when the local version satisfies the requirement. The external repos pull soroban-poseidon from crates.io, which pins soroban-sdk to a specific major (`^26` for OpenZeppelin's stellar-tokens, `^25` for the privacy-pools example), so once the release bumped the local checkout to 27.0.0-rc.1 the `cargo update --precise` step could no longer redirect those transitive soroban-sdk edges and failed to resolve — breaking every OpenZeppelin job (turning the OZ check red on the 27.0.0-rc.1 release PR #1912) and previously forcing privacy-pools to be excluded from the soroban-examples matrix. Reading the destination's resolved soroban-sdk version with `cargo metadata` and relabelling the local crates to it keeps those requirements satisfiable, so the patch resolves and the contracts compile against the local soroban-sdk code regardless of the version bump.

### Known limitations

The relabel only changes the version Cargo resolves against; it does not address a destination whose source is genuinely incompatible with the newer SDK's API, which would still surface as a normal build or test failure. However, we try really hard not to introduce breaking changes into sdk's, even though there are major version bumps, so most of the time this should result in a passing build and if not then we've detected the type of breakage we want to avoid.

### DIFF
--- a/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
+++ b/.github/workflows/test-with-openzeppelin-stellar-contracts.yml
@@ -67,6 +67,36 @@ jobs:
         is_a_contract=$(cargo metadata --no-deps --format-version 1 | jq -r --arg pwd "$(pwd)" '.packages[] | select(.manifest_path == "\($pwd)/Cargo.toml") | (any(.dependencies[]; .name == "soroban-sdk") and any(.targets[]; any(.crate_types[]; . == "cdylib")))')
         echo "is-a-contract=$is_a_contract" | tee -a $GITHUB_OUTPUT
 
+    - name: Align the local soroban-sdk version with the destination
+      working-directory: ${{ github.workspace }}
+      run: |
+        # The destination workspace pins soroban-sdk (and the other published
+        # rs-soroban-sdk crates) to a released version, and pulls crates such as
+        # soroban-poseidon from crates.io that depend on that same major. When the
+        # local checkout has bumped to a newer major (for example a 27.0.0-rc
+        # release), the [patch.crates-io] + `cargo update --precise` step below can't
+        # redirect those transitive soroban-sdk edges, because Cargo only applies a
+        # patch when the local version satisfies the requirement (soroban-poseidon's
+        # ^26). Relabel the local crates with the version the destination already
+        # resolves so the patch satisfies the requirement, collapsing every
+        # soroban-sdk edge onto the local checkout's code regardless of the bump.
+        dest_version=$(
+          cargo metadata --manifest-path stellar-contracts/Cargo.toml --format-version 1 \
+            | jq -r '.packages[] | select(.name == "soroban-sdk") | .version' \
+            | head -n1
+        )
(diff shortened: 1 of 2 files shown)


.github/workflows/test-with-openzeppelin-stellar-contracts.yml
```diff
@@ -67,6 +67,36 @@ jobs:
...
+        # release), the [patch.crates-io] + `cargo update --precise` step below can't
+        # redirect those transitive soroban-sdk edges, because Cargo only applies a
+        # patch when the local version satisfies the requirement (soroban-poseidon's
+        # ^26). Relabel the local crates with the version the destination already
+        # resolves so the patch satisfies the requirement, collapsing every
+        # soroban-sdk edge onto the local checkout's code regardless of the bump.
+        dest_version=$(
+          cargo metadata --manifest-path stellar-contracts/Cargo.toml --format-version 1 \
+            | jq -r '.packages[] | select(.name == "soroban-sdk") | .version' \
+            | head -n1
```
mootz12, line 86: Could there be more than 1 `soroban-sdk` version in the graph? This looks to only grab the first one.

I suppose it would fail if there were two anyway, so this is fine for now.

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
mootz12, line 174: If I understand correctly this is probably redundant now? I suppose it's harmless and potentially protects against cargo getting confused if a newer release is being used.

mootz12 (approved):
LG2M. Left a few comments inline with questions. Going to merge this through to unblock the release as neither of the comments are blockers.

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
leighmcculloch: In theory yes but the likelihood is low.

.github/workflows/test-with-openzeppelin-stellar-contracts.yml
leighmcculloch: Yup, we can fix this on main separately after the rc release. Also further up in the workflow we both modify the deps and add a patch, where-as we only need to add a patch now.

PR #1915, opened 2026-06-18
Add a delegated auth rustdoc example

leighmcculloch:
### What

Add a complete, compiled rustdoc example to the custom_account module showing how a modular custom account uses `CustomAccount::get_delegated_signers` and `CustomAccount::delegate_auth` to delegate its `__check_auth` authentication to other accounts, and reference that example from both functions. Derive `Debug` for the `auth::Context` types so the example can compare the authorization contexts a delegate observed against an expected value with `assert_eq!`.

### Why

PR #1896 introduced the CAP-71 auth delegation API but deferred adding documentation examples for it. The delegation flow is intricate and easy to misuse — it spans constructing an `AddressWithDelegates` authorization entry, retrieving and validating the delegated signers, and forwarding the authorization to each — so a worked example that is compiled and run on every build both teaches the intended flow and guards it against regressions, in keeping with the SDK convention of providing examples for the auth-related functions.

### Known limitations

N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_01CdHVScFCj5PZ3anXAikhV4)_

### DIFF
--- a/soroban-sdk/src/custom_account.rs
+++ b/soroban-sdk/src/custom_account.rs
@@ -82,8 +82,8 @@ impl CustomAccount {
     ///
     /// ### Examples
     ///
-    /// A "modular" custom account that performs no authentication itself, and
-    /// instead delegates it to a set of registered signer addresses. The user
+    /// A module custom account that performs authentication by delegating to
+    /// other accounts instead of doing the authentication itself. The user
     /// chooses which of the registered signers to authenticate with, by
     /// attaching them to the transaction as delegated signers.
     ///
@@ -101,10 +101,10 @@ impl CustomAccount {
     ///     UnknownDelegate = 1,
     /// }
     ///
-    /// // Marks an address as a signer allowed to authenticate for the modular
-    /// // account.
     /// #[contracttype]
     /// enum ModularAccountDataKey {
+    ///     // Marks an address as a signer allowed to authenticate for the
+    ///     // modular account.
     ///     Signer(Address),
     /// }
     ///
@@ -140,9 +140,7 @@ impl CustomAccount {
     ///         // account's authorization.
     ///         let delegates = env.custom_account().get_delegated_signers();
     ///
-    ///         // The host does not validate the delegates, so the account must
-    ///         // check that each one is actually a registered signer before
-    ///         // trusting it.
+    ///         // Check if the delegates are accepted by the modular account.
     ///         for delegate in delegates.iter() {
     ///             if !env
     ///                 .storage()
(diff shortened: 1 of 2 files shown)


soroban-sdk/src/custom_account.rs
```diff
@@ -2,6 +2,231 @@
...
+//! #[contracttype]
+//! enum DelegateAccountDataKey {
+//!     // Records the contexts the delegate approved so the test can verify
+//!     // the delegation reached it.
+//!     ApprovedContexts,
+//! }
+//!
+//! // A simple account that the ModularAccount can delegate to for auth.
+//! //
+//! // It will always authorise an auth request, and store a copy of the auth
```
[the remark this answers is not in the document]
mootz12: Topical considering USA plays Australia tomorrow. Maybe we should let the game decide the spelling outcome 👀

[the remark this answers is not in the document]
leighmcculloch: lol. It's fixed in 3050f3e. It should be US spelling in the code. I use AU spelling in github issues/comments and chat, and I guess I either made this mistake myself or I tricked Claude into using it.

soroban-sdk/src/custom_account.rs
```diff
@@ -2,6 +2,233 @@
...
+//!     // `delegate`, which approves it.
+//!     ProtectedClient::new(&env, &protected).protected(&account);
+//!
+//!     // The account authorized the `protected` call. Delegating to
+//!     // `delegate` is not recorded as a separate authorization.
+//!     use soroban_sdk::{
+//!         auth::ContractContext,
+//!         testutils::{AuthorizedFunction, AuthorizedInvocation},
+//!         IntoVal, Symbol,
+//!     };
```
mootz12: (picky) can we move this up into the top level use statement?

soroban-sdk/src/custom_account.rs
```diff
@@ -2,6 +2,233 @@
...
+//!             sub_invocations: VecM::default(),
+//!         },
+//!     }]);
+//!
+//!     // The call succeeds: the account delegates its authentication to
+//!     // `delegate`, which approves it.
+//!     ProtectedClient::new(&env, &protected).protected(&account);
+//!
+//!     // The account authorized the `protected` call. Delegating to
+//!     // `delegate` is not recorded as a separate authorization.
```
mootz12, line 193: I wonder if this is something we should consider improving. As it stands, this seems like a fairly convoluted way to verify delegated auth is correctly tracked.

What about a new function `env.delegated_auths()` where we can track what the env would have checked? On brief glance it looks like the host tracks this it just needs a getter like `auths` currently has.

Thoughts @dmkozh ?

mootz12 (approved):
LG2M. A few comments inline.

soroban-sdk/src/custom_account.rs
leighmcculloch: Agreed and I think we discussed this idea already on the previous PR that added the CAP-71 functionality. It requires some planning with changes to the env too, so we can do it, but it shouldn't block us exampling what's available today even if the way to assert requires a bit of hoop jumping.

soroban-sdk/src/custom_account.rs
leighmcculloch: 5801e33

soroban-sdk/src/custom_account.rs
```diff
@@ -2,6 +2,233 @@
...
+//!     assert_eq!(
+//!         env.auths(),
```
dmkozh, line 195: This assertion is misleading and unnecessary. In the enforcing mode env.auths() would return the contents of set_auths, so it doesn't actually assert anything useful.

soroban-sdk/src/custom_account.rs
dmkozh: > What about a new function env.delegated_auths() where we can track what the env would have checked?

Yeah, we can do that, though it's going to be a bit awkward, as user still needs to provide at least the delegate addresses, and also we need to use the enforcing auth mode, which is normally not compatible with auth recording.

I guess what we could do for the unit tests is to use try_invoke_check_auth in combination with recording auth and a setter for the delegate addresses. This way  the recording semantics is preserved. More comprehensive e2e tests could use the current approach with enforcing auth.

PR #1917, opened 2026-06-18
Support unit structs as contract events

### LINKED ISSUE #1571: contractevent macro should support empty structs without fields
The contractevent macro currently only works on types defined with fields like `pub struct MyEvent { /* fields */ }`, but it should also work on empty structs defined like `pub struct MyEvent;` to allow for simple event markers without additional data.

leighmcculloch:
### What

The contractevent attribute macro now accepts fieldless unit structs such as `pub struct MyEvent;`, treating them as events that carry no topic or data fields and emit only the default prefix topic derived from the struct name.

### Why

Issue #1571 reports that the macro only worked on structs with named fields and rejected unit structs with a hard compile error, which prevented defining simple event markers that carry no additional data. Empty braced structs (`pub struct MyEvent {}`) already worked, so the gap was specifically the unit-struct form.

### Known limitations

When a contract's spec is regenerated via contractimport!, a fieldless event is rendered as an empty braced struct (`pub struct MyEvent {}`) rather than a unit struct, because the on-wasm spec records only that the event has no parameters and does not preserve the unit-versus-braced distinction. The generated type produces an identical on-chain event so that difference is of little consequence.

Close #1571

---
_Generated by [Claude Code](https://claude.ai/code/session_011mKQ41rU7Ze9hxQAUEdfmk)_


soroban-sdk-macros/src/derive_event.rs
```diff
@@ -114,10 +114,7 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
                     "structs with unnamed fields are not supported as contract events",
                 )
                 .with_span(&struct_.fields.span()))?,
-                Fields::Unit => Err(Error::custom(
-                    "structs with no fields are not supported as contract events",
-                )
-                .with_span(&struct_.fields.span()))?,
+                Fields::Unit => Vec::new(),
```
mootz12, line 117: Is an empty `Vec` the right form for this? I supposed I'd expect this to be `ScVal::Void` for both the unit and empty event.

Does this match the data form for:
```
    #[contractevent]
    pub struct MyEvent {
        #[topic]
        name: Symbol,
    }
```

and
```
env.events().publish(vec![&env, symbol_short!("MyEvent")].into(), ());
```

If so, it's consistent so is fine.

mootz12 (approved):

soroban-sdk-macros/src/derive_event.rs
leighmcculloch: The `Vec` here is just the list of fields, it's not the data type in the resulting event.

See below in the test where it shows the final XDR form, it actually maps to an empty `Map`.

I was thinking it would be a better tradeoff to not introduce a special case, since all events have a Map data by default, and that makes the event intuitively extendible. And it is the same form in the example you posted above too with the MyEvent that has a topic and no data fields, that is still a Map too.

Since the Map is empty I think it is still very effecient, it will be 4 bytes long. A Void would be slightly more efficient being that 4 bytes shorter, but that's not much.

I guess a Void maybe tells a better narrative? And everyone loves an optimisation.

How about we consider the optimisation separate to this change, the optimisation being zero fields results in a Void instead of a Map, which would cover your example above.

mootz12: I think consistency is best. And yeah, all forms above would map to the same event, so lg2m

PR #1939, opened 2026-07-20
Derive as_contract frame func from closure name

### LINKED ISSUE #1673: Env `as_contract` function uses dummy variable
The `as_contract` function on the env adds an [empty string](https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk/src/env.rs#L1542) as the "func" arg to `with_test_contract_frame`.

Is this needed? If not, the env should remove the "func" arg. If it is, we should add the func argument to `as_contract`.

It appears the "func" arg impacts the result of `require_auth`, so it might be useful for people to have when testing custom auth. For an example, see: https://github.com/stellar/rs-soroban-sdk/blob/main/tests/account/src/lib.rs

This was brought up in discussion on this PR: https://github.com/stellar/rs-soroban-env/pull/1628

leighmcculloch:
### What

`Env::as_contract` and `Env::try_as_contract` now derive the test contract frame's function name from the function or closure passed to them, instead of always using an empty symbol. The name is taken from `core::any::type_name`: the final path segment is isolated (respecting `::` that appears inside generic arguments or an enclosing generic context) and its own generic arguments are dropped, so a named function resolves to its own name, a name longer than a Symbol allows is truncated to the 32 character limit, and closures or otherwise unrepresentable names continue to fall back to the empty symbol. The public signatures are unchanged; the type of the passed function is inferred through a private helper so no turbofish call sites break.

### Why

The frame's `func` symbol was previously hardcoded to the empty symbol, so the test frame carried no record of which function the closure stood in for. Deriving it from the passed function gives diagnostic frames a meaningful name at no cost to the historical empty-name behavior for closures, and lets a caller stand in for a specific named function when running code inside a contract frame.

Close https://github.com/stellar/rs-soroban-sdk/issues/1673

### Known limitations

`core::any::type_name` is best-effort and its exact format is unstable, so the derived name is only ever a diagnostic hint. It also feeds the `func` passed to the test contract frame, which gates the `__check_auth`-only host functions (`CustomAccount::get_delegated_signers`/`delegate_auth`), so a future change to the `type_name` format could change which name is captured and, for a caller that relies on a specific derived name, whether such a gated call is permitted. If behavior ever needs to be keyed reliably on `func`, an explicit function-name parameter should be added rather than relying on the inferred name. The derived name has no direct getter, but it is the function name recorded for authorizations performed inside the frame, so the tests assert it through `MockAuth`: they run named, generic, generic-enclosing, closure, and over-long-named callables inside a frame and check that a `require_auth` there is authorized only under the expected symbol (the function's own name, the generic base name, the empty symbol, and the 32-character truncation), with a non-matching name confirmed to fail.


mootz12 (commented):
I have a few concerns on merging this now:

Closures ->
```
env.as_contract(&id, || {
    // some code
});
```
 Most contracts that use this `as_contract` are often doing a function call + validation, and opt for a closure. AFAIK this won't impact those?

Breaking change ->
This will break tests that use `set_auth` where require_auth is used within a closure.

---

I agree this is a better way to do it, but I'm not sure it's worth a breaking change. Thoughts?

leighmcculloch:
> This will break tests that use `set_auth` where require_auth is used within a closure.

Can you provide an example of what the breaking change would be? As far as I understood for how I designed this is that when used with an enclosure nothing should change. Behaviour only changes when you pass in a function with a name. Any breakage for functions with a name will only exist in tests. And given the edge case nature of passing in a function with a name, that seems tolerable.

mootz12:
> Can you provide an example of what the breaking change would be? As far as I understood for how I designed this is that when used with an enclosure nothing should change.

Sorry, I agree with you here. Original comment had a typo. Meant "used without a closure.".

Per https://github.com/stellar/rs-soroban-sdk/issues/1673, this was mostly written to make it clearer what the func arg was for folks testing auth via `as_contract`. I'm not sure this direction does that.

I originally had a change queued that just set the string to `__as_contract` or `__try_as_contract` instead of an empty string. But ultimately left it due to it breaking the test quoted in the issue.

leighmcculloch:
> Per #1673, this was mostly written to make it clearer what the func arg was for folks testing auth via `as_contract`. I'm not sure this direction does that.

This was a great prompt thanks. I had intended to include examples for how to use this for auth and forgot. In the process of trying to do so, this is more horrible to use than I expected. In my testing I had only done trivial things inside a named function, but the moment you try to do something with Addresses, or SDK types needing an Env, it falls apart because in Rust a named function can't reference state from the existing function it is defined. This makes it pretty hard to use in a test.

So going to close this.

If we want as_contract with fn names we'll need to add a new function that has the name as a parameter.

PR #1943, opened 2026-07-21
Switch to recording auth for native constructors

### LINKED ISSUE #1738: `register_at` does not switch to recording auth for WASM constructors

Latest main branch (`soroban-sdk/src/env.rs`)

### What did you do?

Called `register_at(addr, WASM, constructor_args)` with a WASM contract whose constructor calls `require_auth`.

The `register_at` code path (`register_contract_with_contract_id_and_executable` at `env.rs:1502-1539`) calls the constructor without switching to recording auth:

```rust
// register_contract_with_contract_id_and_executable
// Used by register_at — NO auth switching
self.env_impl
    .call_constructor_for_stored_contract_unsafe(&contract_id, constructor_args.to_object())
    .unwrap();
```

Compare with the `register` code path (`register_contract_with_source` at `env.rs:1086-1089`) which does switch:

```rust
// register_contract_with_source
// Used by register — HAS auth switching
let prev_auth_manager = self.env_impl.snapshot_auth_manager().unwrap();
self.env_impl
    .switch_to_recording_auth_inherited_from_snapshot(&prev_auth_manager)
    .unwrap();
// ... constructor runs with recording auth ...
self.env_impl.set_auth_manager(prev_auth_manager).unwrap();
```

`register_at` should behave the same as `register` with respect to authorization during the constructor call. The two APIs are documented to differ only in whether the contract ID is auto-generated or user-supplied. The constructor should run under recording auth so that `require_auth` calls succeed.

### What did you see instead?

`register_at` calls the constructor under whatever auth mode is currently active. Any WASM constructor calling `require_auth` fails with an unexpected auth error when registered via `register_at`, while the same contract works when registered via `register`.

leighmcculloch:
### What

Switch to recording auth before invoking a native (non-Wasm) contract's constructor during `register` and `register_at`, matching the behavior already used for the Wasm registration paths.

### Why

A native contract whose constructor calls `require_auth` could not be registered with `register` or `register_at` unless authorization was mocked, because `register_contract_with_constructor` invoked the constructor under the default enforcing auth manager. The Wasm paths already snapshot the auth manager, switch to recording auth inherited from that snapshot, and restore it afterward so a constructor's `require_auth` is auto-authorized during registration; the native path did not, so the same contract behaved differently depending on whether it was registered natively or as Wasm. This closes that gap, which was called out as a known limitation of #1933. The auth manager is restored before the registration result is unwrapped so that a failing constructor does not leave the `Env` stuck in recording mode.

Close https://github.com/stellar/rs-soroban-sdk/issues/1738


dmkozh (approved):

leighmcculloch:
Discussing further at https://github.com/stellar/rs-soroban-sdk/issues/1738#issuecomment-5040496581 before merging.

leighmcculloch:
We will treat this like a bug fix and so it can be shipped in a patch release.

PR #1946, opened 2026-07-22
Refresh mainnet() resource limits and fees

### LINKED ISSUE #1931: `InvocationResourceLimits::mainnet()` is out of date with actual Mainnet network settings

soroban-sdk 27.0.0 (current `main`, e5cb4b52). The values appear to date back to #1677, which introduced resource limit enforcement.

### What did you do?

Compared the hardcoded limits in [`InvocationResourceLimits::mainnet()`](https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk/src/testutils/cost_estimate.rs), which `Env::default()` has enforced on every test invocation since v25, with live Mainnet settings fetched on 2026-07-10 (`stellar network settings --network mainnet`; also visible on [Stellar Lab: Network Limits](https://lab.stellar.org/network-limits)).

The SDK limits should match Mainnet's per-transaction limits, since the enforcement exists to warn developers when a contract "might be too resource-heavy to run on a real network".

### What did you see instead?

These fields differ:

- `instructions`: the SDK uses 600,000,000 instead of 400,000,000
- `disk_read_entries`: the SDK uses 100 instead of 200
- `write_entries`: the SDK uses 50 instead of 200
- `ledger_entries`: the SDK uses 100 instead of 400

The remaining fields match: `mem_bytes`, `disk_read_bytes`, `write_bytes`, `contract_events_size_bytes`, `max_contract_data_key_size_bytes`, `max_contract_data_entry_size_bytes`, `max_contract_code_entry_size_bytes`.

The two directions of staleness have different consequences:

- **`instructions` too high:** a contract can pass `Env::default()` tests yet exceed Mainnet's transaction limit in production, defeating the enforcement's early-warning purpose.
- **Entry limits too low:** tests can panic for contracts that are valid on Mainnet, pushing developers toward `disable_resource_limits()` and removing the safety net entirely.

### Why is this important?

AI coding assistants often treat SDK values and rustdocs as ground truth for network behavior. I have seen assistants rely on these stale limits when judging whether a contract fits within Mainnet's budget, misleading both the model and the developers using it.

### Related staleness in the same file

`CostEstimate::fee()` uses a Pubnet fee snapshot from 2024-12-11 that could be refreshed in the same pass:

| Fee | SDK snapshot | Mainnet (2026-07-10) |
|---|---:|---:|
| `fee_per_instruction_increment` | 25 | 7 |
| `fee_per_disk_read_entry` | 6,250 | 1,563 |
| `fee_per_write_entry` | 10,000 | 2,500 |
| `fee_per_disk_read_1kb` | 1,786 | 447 |
| `fee_per_write_1kb` | 3,500 | 875 (`fee_write1_kb`) |
| `fee_per_historical_1kb` | 16,235 | 4,059 |
| `fee_per_contract_event_1kb` | 10,000 | 5,000 |
| `fee_per_transaction_size_1kb` | 1,624 | 406 |

### Suggested fix

1. Update the `mainnet()` rustdocs to point readers — human and AI alike — to a live source (`stellar network settings --network mainnet` or [Stellar Lab: Network Limits](https://lab.stellar.org/network-limits)) rather than relying on the hardcoded snapshot.
2. Update `mainnet()` to the current values (`instructions: 400_000_000`, `disk_read_entries: 200`, `write_entries: 200`, `ledger_entries: 400`).
3. Refresh the `fee()` snapshot and its "as of" comment.
4. Consider a scheduled CI job, or a release-checklist item, that diffs `mainnet()` against `stellar network settings --network mainnet` — nothing currently catches drift. Since Mainnet limits are validator-voted and change infrequently, a weekly or monthly run would be enough.

leighmcculloch:
### What

Refresh the `InvocationResourceLimits::mainnet()` limits and the fee-rate snapshot used by `CostEstimate::fee()` to the Stellar Mainnet settings as of 2026-07-10, and update their rustdocs (including the duplicated limits table in the v25 resource-limits migration guide) to point readers to `stellar network settings --network mainnet` and Stellar Lab as the live source of truth.

### Why

`InvocationResourceLimits::mainnet()` has been enforced by `Env::default()` on every test invocation since v25 but still carried the values from when resource-limit enforcement was first introduced, which had drifted from Mainnet in both directions: the instruction ceiling was too high (600,000,000 vs 400,000,000), letting a contract pass tests yet exceed Mainnet's per-transaction limit in production and defeating the early-warning purpose of the enforcement, while the entry ceilings were too low (disk read entries 100 vs 200, write entries 50 vs 200, ledger entries 100 vs 400).

### Known limitations

The values remain a hardcoded snapshot rather than being pulled dynamically, so they will drift again.

On the one hand this is bad, but also, if they kept moving tests could become flakey, so timing this update with new major releases on some cadence is probably a good idea.

Follow up to introduce CI to help us keep it updated:

- https://github.com/stellar/rs-soroban-sdk/issues/1948

### DIFF
--- a/soroban-sdk/src/tests/cost_estimate.rs
+++ b/soroban-sdk/src/tests/cost_estimate.rs
@@ -37,14 +37,14 @@ fn test_cost_estimate_with_storage() {
     .assert_eq(format!("{:#?}", e.cost_estimate().resources()).as_str());
     expect![[r#"
         FeeEstimate {
-            total: 29122,
-            instructions: 607,
-            disk_read_entries: 6250,
-            write_entries: 10000,
+            total: 10004,
+            instructions: 170,
+            disk_read_entries: 1563,
+            write_entries: 2500,
             disk_read_bytes: 0,
-            write_bytes: 274,
+            write_bytes: 69,
             contract_events: 0,
-            persistent_entry_rent: 11991,
+            persistent_entry_rent: 5702,
             temporary_entry_rent: 0,
         }"#]]
     .assert_eq(format!("{:#?}", e.cost_estimate().fee()).as_str());
@@ -70,8 +70,8 @@ fn test_cost_estimate_with_storage() {
     .assert_eq(format!("{:#?}", e.cost_estimate().resources()).as_str());
     expect![[r#"
         FeeEstimate {
-            total: 603,
-            instructions: 603,
+            total: 169,
+            instructions: 169,
             disk_read_entries: 0,
             write_entries: 0,
             disk_read_bytes: 0,
@@ -103,10 +103,10 @@ fn test_cost_estimate_with_storage() {
     .assert_eq(format!("{:#?}", e.cost_estimate().resources()).as_str());
     expect![[r#"
         FeeEstimate {
-            total: 16847,
-            instructions: 597,
-            disk_read_entries: 6250,
-            write_entries: 10000,
+            total: 4231,
+            instructions: 168,
+            disk_read_entries: 1563,
+            write_entries: 2500,
             disk_read_bytes: 0,
             write_bytes: 0,
             contract_events: 0,
@@ -136,8 +136,8 @@ fn test_cost_estimate_with_storage() {
     .assert_eq(format!("{:#?}", e.cost_estimate().resources()).as_str());
     expect![[r#"
         FeeEstimate {
-            total: 598,
-            instructions: 598,
+            total: 168,
+            instructions: 168,
             disk_read_entries: 0,
             write_entries: 0,
             disk_read_bytes: 0,
--- a/soroban-sdk/src/testutils/cost_estimate.rs
+++ b/soroban-sdk/src/testutils/cost_estimate.rs
@@ -40,7 +40,10 @@ impl CostEstimate {
     /// on 2026-07-10. Because the fees are hardcoded rather than pulled
     /// dynamically, they may drift from the live network over time; the current
     /// values can be checked via `stellar network settings --network mainnet`
-    /// or on Stellar Lab: <https://lab.stellar.org/network-limits>.
+    /// or on Stellar Lab: <https://lab.stellar.org/network-limits>. The one
+    /// exception is the per-1KB storage rent rate, which is a deliberate
+    /// conservative overestimate rather than the snapshot value, so storage
+    /// rent estimates may be higher than the live network charges.
     ///
     /// Take the return value with a grain of salt as both the resource estimate
     /// and the fee rates may be imprecise.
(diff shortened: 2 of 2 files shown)


leighmcculloch:
CI note: the `test-example (import_ark_bn254, false)` and `(…, true)` jobs fail, and this is the change working as intended rather than a regression to paper over. That example's `test_running_contract_as_wasm` consumes 404,552,943 instructions, which now trips the refreshed Mainnet ceiling of 400,000,000 (it fit under the previous, too-high 600,000,000). Surfacing exactly this kind of over-limit contract — one that passes today but would exceed Mainnet's per-transaction instruction limit — is the point of #1931.

The example lives in `stellar/soroban-examples` (pulled in by `.github/workflows/test-with-soroban-examples.yml`, which checks out that repo's `main`), so it can't be fixed from this PR. Resolving the job needs a decision there — e.g. calling `disable_resource_limits()` or setting a custom higher limit in that heavy-pairing example's test, or accepting that a single bn254 pairing at ~404M isn't meant to fit within a Mainnet transaction. Other heavy-crypto examples could hit the same ceiling. Flagging for a maintainer call rather than lowering the SDK value back toward 600M, which would defeat the purpose of the change.

---

leighmcculloch:
I'm opening to merging this in a patch release given an experience of this is that it is surprising. But also, it could break peoples tests as it does for the soroban-examples, in which case it would be better to hold this for the v28 release and merge it just prior.

leighmcculloch:
The soroban-example failures are being fixed in:

mootz12 (approved):
Good catch on this drift.

Should we cache this for the user during test runs or something instead of requiring a release? We can expire it on some window.

This value changes unrelated to protocol upgrades, so I don't think tying it to SDK releases is the best.

leighmcculloch:
> Should we cache this for the user during test runs or something instead of requiring a release?

This is an interesting idea. There would need to be some built in defaults to apply when used offline. And test reproducibility is really important so it would be important that once a test ran once with limits that it ran every future time with the same values until a user took action to update it. I guess we could have built in defaults and write them to a .json file that gets committed with the project.

Is that what you had in mind?

leighmcculloch:
> This value changes unrelated to protocol upgrades, so I don't think tying it to SDK releases is the best.

For this reason I don't think it should be tied to major releases, and so we can release it with a patch release as a bug fix / tweak.

leighmcculloch:
If there was a way to get the limits at a ledger number, then it could be convenience to cache a past limit in, but then you could get the limit for a given ledger number in your test that would be cached. That makes the values less arbitrary.

PR #1951, opened 2026-07-24
Deprecate lib arg on contracttype/error/event

### LINKED ISSUE #1952: Deprecate the `lib` arg on contracttype/error/event

The `lib` parameter on `contracttype`/`contracterror`/`contractevent` macros is a hidden, unused feature. It's undocumented and never turned out to be something that was usable. On [mainnet there are zero contracts using it](https://github.com/stellar-experimental/contract-wasms/blob/main/analysis/spec-contracttype-lib-usage.csv) and there are no situations where the feature would be recommended for use.

If someone discovers it and starts using it, they probably shouldn't.

Deprecate it, and remove it once it has been deprecated for a little while.

Remove it now, or leave it forever.

leighmcculloch:
### What

`lib` on `contracttype`/`contracterror`/`contractevent` now warns: deprecated.

### Why

Nobody uses it. Undocumented, never shipped, [zero on mainnet](https://github.com/stellar-experimental/contract-wasms/blob/main/analysis/spec-contracttype-lib-usage.csv). Warn now, remove later.

Close https://github.com/stellar/rs-soroban-sdk/issues/1952


mootz12 (approved):

leighmcculloch:
Because this introduces a deprecation lint which is a possible breaking change, Rust conventions require this as a minor version, so I'll hold this until it seems like we aren't pushing any more bug fixes for the 27.0 stream.

PR #1959, opened 2026-07-28
Reset budget in disable_resource_limits

### LINKED ISSUE #1949: Change disable_resource_limits API to be more intuitive

27.0.2

### What did you do?

Called env.cost_estimate().disable_resource_limits() in a contract test when I needed unlimited limits:

The call to disable_resource_limits be sufficient to test without limits.

The API to be less confusing, like what is reset_unlimited() vs disable_resource_limits().

### What did you see instead?

```
Error(Budget, ExceededLimit)
```

I needed to also call `env.cost_estimate().budget().reset_unlimited()`.

leighmcculloch:
### What

`CostEstimate::disable_resource_limits()` now also resets the CPU instruction and memory budget to unlimited, so a single call is enough to run invocations without any limits.

### Why

The CPU/memory budget is metered and enforced separately from the invocation resource limits, so disabling the latter alone still left heavy invocations failing with `Error(Budget, ExceededLimit)` until `budget().reset_unlimited()` was also called, which was surprising and unintuitive.

### Known limitations

N/A

Close #1949

### DIFF
--- a/soroban-sdk/src/tests/cost_estimate.rs
+++ b/soroban-sdk/src/tests/cost_estimate.rs
@@ -149,6 +149,29 @@ fn test_cost_estimate_with_storage() {
     .assert_eq(format!("{:#?}", e.cost_estimate().fee()).as_str());
 }
 
+// A single call to `disable_resource_limits()` should be sufficient to run
+// invocations without any limits, including the CPU/memory budget that is
+// enforced separately from the invocation resource limits.
+#[test]
+fn test_disable_resource_limits_also_disables_budget() {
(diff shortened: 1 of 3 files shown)


leighmcculloch:
After playing with this I don't like this solution to the problem. If we reset the limits to unlimited automatically, then when enabling invocation limit checks the limits have changed and need to be set back. This creates an odd series of fn calls required if I have set specific limits, I need to keep setting them.

This makes me think we should rethink how to use these functions, or maybe simply add more docs noting that you need to call reset unlimited.

PR #1961, opened 2026-07-29
Document resource limit and budget interplay

### LINKED ISSUE #1949: Change disable_resource_limits API to be more intuitive

leighmcculloch:
### What

Expand the `CostEstimate` and `Budget` docs to explain that the CPU/memory budget and the invocation resource limits are two independent limits, each enforced at a different point and each needing its own call to change, with contract-test examples for the common cases.

### Why

After `disable_resource_limits()`, code written to exceed the limits still fails with `Error(Budget, ExceededLimit)`, because the call turns off only the post-invocation check and leaves the budget's own limits in place, and nothing in the docs said the budget was a separate limit at all.

### Known limitations

No API change, so running an invocation with no limits still takes two calls. Automatically lifting the budget from `disable_resource_limits()` was tried and rejected, because it resulted in still an unintuitive behaviour. To address the API awkwardness the API really needs changing as a whole.

Close #1949


leighmcculloch:
@dmkozh Keen on your thoughts on this if you think there is something we can do non-breaking with the API that makes sense. Otherwise I think this docs improvements is beneficial.

leighmcculloch:
Chatted to @dmkozh offline and he's got an idea for how to restructure this API, so I'm going to close this in favour of that happening instead.

PR #1963, opened 2026-07-29
Extract shared contract type spec emission

leighmcculloch:
### What

Move the identical spec-emission block copied across the struct, tuple struct, enum, enum int, and error enum derives into one `spec::type_spec` helper.

### Why

Five copies of the same `__SPEC_XDR_TYPE_*` static and `spec_xdr()` const fn had to be edited in lockstep on every change to how type specs are emitted.

### Known limitations

N/A


leighmcculloch:
Superseded by:
- https://github.com/stellar/rs-soroban-sdk/pull/1965

PR #1964, opened 2026-07-29
Generate contract specs as const ScSpecEntryRef

leighmcculloch:
### What

Generate every contract spec entry — types, events, and functions — as a const `ScSpecEntryRef` value that the contract crate encodes to XDR at compile time, instead of encoding it inside the macro and emitting an opaque byte-string literal.

### Why

The spec a macro produces is currently invisible in expanded code and unreachable from Rust, so it can only be inspected by decoding the bytes out of the built wasm; a const value can be read in `cargo expand` and reused by other generated code. Spec section bytes are unchanged for all 43 test contracts.

### Known limitations

Depends on stellar/rs-stellar-xdr#562 for the borrowing `*Ref` types and their const XDR serializers, so this carries a `[patch.crates-io]` pin on a stellar-xdr git rev until that lands and is released. Spec shaking still encodes the spec at macro time to compute its marker hash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1686,13 +1686,12 @@ dependencies = [
 [[package]]
 name = "stellar-xdr"
 version = "27.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05ff843326969bdf1ef673dcdba94c08f4a3c8f1e58d6e6ef39b1bd4f749179a"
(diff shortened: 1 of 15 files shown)


leighmcculloch:
Superseded by:
- https://github.com/stellar/rs-soroban-sdk/pull/1965

Issue #1857, opened 2026-04-30
Type alias for function signatures produces invalid specs


mootz12:
### What version are you using?

main with feature `expiremental_spec_shaking_v2`

### What did you do?

Used a Rust type alias in a contract function signature. This applies if the aliased type is a primitive like `i128` or a custom type like in the example below.

Minimal example:

```rust
#![no_std]

use soroban_sdk::{contract, contractimpl, contracttype};

#[contracttype]
#[derive(Clone)]
pub struct Item {
    pub amount: i128,
}

pub type ItemAlias = Item;

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn echo(item: ItemAlias) -> ItemAlias {
        item
    }
}
```

Under spec shaking v2, generated boundary code roots reachable types by calling the marker method for the function parameter and return types, conceptually:

```rust
<ItemAlias as SpecShakingMarker>::spec_shaking_marker();
```

But `ItemAlias` is only a Rust type alias. Rust trait lookup normalizes it to the aliased type, so this call resolves as if it were:

```rust
<Item as SpecShakingMarker>::spec_shaking_marker();
```

That means the surviving marker is the marker for the `Item` spec entry.

At the same time, the contract spec macro builds the public function spec from the source syntax. It sees the token `ItemAlias` in the function signature and emits a function spec that references UDT name `ItemAlias`.

This creates two different views of the same signature:

- `contractspecv0` function entry says the function uses UDT `ItemAlias`.
- `T::spec_shaking_marker()` roots the spec entry for UDT `Item`.

### What did you expect to see?

Either of these would be acceptable:

1. The SDK rejects source-level type aliases in contract-facing signatures with a clear compile-time error.
2. The SDK resolves the alias consistently, so both the public function spec and the spec-shaking marker refer to the same UDT, `Item`.

In either case, spec shaking should not produce a final spec where a function references a UDT name that has no corresponding UDT spec entry.

### What did you see instead?

For the example above, the marker machinery keeps `Item`, but the public function spec can still reference `ItemAlias`. After shaking, this can produce an invalid or unusable contract spec: generated import/client code expects a type named `ItemAlias`, but the actual UDT spec entry is named `Item`.

leighmcculloch:
Note that this is a problem without spec shaking v2, as is illustrated by:
- https://github.com/stellar/rs-soroban-sdk/pull/1971

leighmcculloch:
[...]

Issue #1570, opened 2025-09-19
Improved identity of types in specs


leighmcculloch:
I think we need to explore how to improve the identity of types captured in the contract spec.

Today types are captured in the contract spec as only a name. This works for many cases, but problems arise in complex contracts, or even simple contracts that use multiple types with the same names in different modules.

In that latter case a contract developer will not themselves experience any problems, but another contract attempting to import their contract will encounter the dilemma that their contract will have multiple types with the same name and importing will fail.

In other cases it becomes impossible to identify which type functions in the interface are referencing.

In cases where types have been defined in a library, it can get even more confusing when those types lift into the contract, such as what has been observed in:
- https://github.com/stellar/rs-soroban-sdk/issues/1569

leighmcculloch:
Whilst I wonder if we can still do something here, I think the problems have been largely mitigated by the following changes that made it much more predictable and sensible what types get included into specs as well as introducing warnings when types with the same names are included in the spec:
- https://github.com/stellar/rs-soroban-sdk/pull/1672
- https://github.com/stellar/stellar-cli/pull/2353
- https://github.com/stellar/stellar-cli/pull/2426

PR #1965, opened 2026-07-29
Encode contract spec XDR at compile time

leighmcculloch:
> [!NOTE]
>
> Discussion is here:
>
> 1. stellar/rs-soroban-sdk#1965 ← this PR
>
> 1. stellar/stellar-xdr#312
> 1. stellar/stellar-cli#2674
>
>
> 1. stellar/stellar-cli#2720

### What

Change when and where contract spec XDR is encoded. Before, the proc-macro encoded the spec and embedded the resulting bytes as an opaque literal in the generated code. Now, the proc-macro constructs a `ScSpecEntryView` over static data and the encoding happens at const evaluation time while the generated code compiles, using the new const XDR encoding support.

### Example
Screenshot 2026-08-07 at 3 25 26 pm

### Why

It will make it possible to embed information into the spec that is only available later in the compilation process, like the fully qualified name of types.

As a bonus the spec stops being an opaque blob and becomes a value: readable in `cargo expand`, referenceable by other generated code, and no longer only inspectable by decoding bytes back out of a built wasm. It is also the foundation for connecting type IDs across types in a future change.


soroban-sdk/src/spec_shaking.rs
```diff
@@ -23,6 +23,159 @@
...
+    0xa2bfe8a1, 0xa81a664b, 0xc24b8b70, 0xc76c51a3, 0xd192e819, 0xd6990624, 0xf40e3585, 0x106aa070,
+    0x19a4c116, 0x1e376c08, 0x2748774c, 0x34b0bcb5, 0x391c0cb3, 0x4ed8aa4a, 0x5b9cca4f, 0x682e6ff3,
+    0x748f82ee, 0x78a5636f, 0x84c87814, 0x8cc70208, 0x90befffa, 0xa4506ceb, 0xbef9a3f7, 0xc67178f2,
+];
+
+/// SHA-256 of `input`, evaluatable at compile time.
+///
+/// A const implementation is needed because the marker is built while the
+/// generated code compiles; `sha2` is not usable in a const context.
+const fn sha256(input: &[u8]) -> [u8; 32] {
```
leighmcculloch: TODO: Find a way to do the sha256 marker generation in a const context without writing our own sha256 impl.

leighmcculloch: Discussion here about adding const fns to the sha2 crate indicate dependence on some still very unstable nightly features becoming stabilized:
- https://github.com/RustCrypto/hashes/issues/288

leighmcculloch: There is also:
- https://crates.io/crates/sha2-const

Cargo.toml
```diff
@@ -53,7 +53,11 @@ version = "=28.0.0"
 #rev = "a749b69b3471aec0c20ec431b0297f3414d66421"
 default-features = false
 
-#[patch.crates-io]
+[patch.crates-io]
+# Pending https://github.com/stellar/rs-stellar-xdr/pull/562, which adds the
+# borrowing `*View` types and their const XDR serializers that the contract spec
+# generation in soroban-sdk-macros builds on.
+stellar-xdr = { git = "https://github.com/stellar/rs-stellar-xdr", rev = "e73df1f6793fe5916dce7a833e14c0718786ab04" }
```
leighmcculloch: CI note on this pin.

`build-fuzz` and `fuzz-corpus` were failing with:

```
package `soroban-sdk` depends on `stellar-xdr` with feature `const`
but `stellar-xdr` does not have that feature
```

`tests/fuzz/fuzz` is its own workspace root (`[workspace] members = ["."]`), so it never inherited this `[patch.crates-io]` and resolved published stellar-xdr 28.0.0. Fixed by copying the same pin into `tests/fuzz/fuzz/Cargo.toml`; `make build-fuzz` and `make fuzz-corpus` both pass now. **The two pins must be removed together** once rs-stellar-xdr#562 merges and publishes.

Still red and not fixable from this repo: the ~120 `test-crate (…)` / `test-example (…)` jobs, same resolution error. Those build external example/package repos as the Cargo build root, so no `[patch.crates-io]` here applies to them. They stay red until 562 merges and a stellar-xdr release carrying the `const` feature is on crates.io — there is no workaround on this side.

---

soroban-sdk-macros/src/derive_contractimpl_trait_default_fns_not_overridden.rs
```diff
@@ -89,7 +89,12 @@ fn derive(args: &Args) -> Result<TokenStream2, Error> {
         Some(trait_ident),
         &args.client_name,
     ));
-    output.extend(derive_fns_spec(&args.spec_name, &fns, spec_export));
+    output.extend(derive_fns_spec(
+        &args.crate_path,
+        &args.spec_name,
+        &fns,
+        spec_export,
+    ));
```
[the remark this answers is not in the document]
mootz12: Hm, I think we do this elsewhere, and it hasn't come up before.

mootz12 (approved):
LG2M!

PR #1968, opened 2026-07-30
Always enable spec shaking

### LINKED ISSUE #1759: Remove `experimental_spec_shaking_v2` feature flag and make it always on

The `experimental_spec_shaking_v2` feature flag gates spec shaking v2 behaviour behind an opt-in flag in v25 ever since v25.2.0. Once it has been validated, it should be made the default and only behaviour, removing the need for users to opt in.

For the v26 major version:

- Remove the `experimental_spec_shaking_v2` feature flag.
- Make the spec shaking v2 behavior enabled by default.
- Add a `disable_spec_shaking_v2` feature flag that allows users to opt out if issues arise during the v26 period.

Depending on feedback received during v26, for v27:

- Remove the `disable_spec_shaking_v2` feature flag entirely.
- Make spec shaking v2 always on with no option to disable.

- Keep the `experimental_spec_shaking_v2` opt-in flag, but this delays adoption.
- Skip the `disable_spec_shaking_v2` escape hatch and make it always on immediately in v26. This should be considered if the feedback during v25 is positive.

### LINKED ISSUE #1844: spec shaking v2 rollout

Spec shaking v2 brings a significant improvement to contract spec generation. Enabling this by default will ensure WASMs are built in a consistent and minimized way.

Spec shaking v2 also requires users to build with a supported build tool like the Stellar CLI to perform the spec filtering step.

#### Protocol 26
* Spec shaking v2 remains opt-in via the `experimental_spec_shaking_v2` feature flag
* Building WASMs without the CLI and with the `experimental_spec_shaking_v2` feature flag results in a build error
* ~Building WASMs without the CLI and without `experimental_spec_shaking_v2` feature flag results in a compiler warning that users should migrate to using a supported build tool like the CLI.~
* All known issues are addressed and patched

#### Protocol 27
* `experimental_spec_shaking_v2` becomes a default feature, enabling spec shaking v2 by default
* Developers can still use spec shaking v1 if they don't depend on `soroban-sdk` as a transitive dependency and disable default features

#### Protocol 28
* `experimental_spec_shaking_v2` is removed, SDK always builds with spec shaking v2

* For protocol 27, we can opt to add a `disable_spec_shaking_v2` feature, that makes it easier to disable if a transitive dependency uses default features.

leighmcculloch:
### What

Remove the `experimental_spec_shaking_v2` Cargo feature and make the behaviour it gated unconditional, so every contract emits spec markers and relies on the build system (i.e. stellar-cli) to shake its spec.

There are some subsequent impacts: Remove the `export` parameter on contract macros.

### Why

The feature had been enabled on all contracts built with the OpenZeppelin/stellar-contracts v0.7.0+ that was released in April 2026, four months ago. In terms of contracts on mainnet, there is no concrete meta that guarantees we can identify OZ contracts, but analysis of mainnet contract specs and identifying the use of types that are unique to the OZ libs suggests that there are 349 wasms on mainnet dependent on OZ's libs, 49 wasms using types only available in 0.7.0+.

Response has been positive.

Close https://github.com/stellar/rs-soroban-sdk/issues/1759
Close https://github.com/stellar/rs-soroban-sdk/issues/1844

### Known limitations

This targets v28 and is breaking in a few two ways.

Contracts must now be built by a build system that shakes the spec, so `cargo build --target wasm32v1-none` on a crate depending on soroban-sdk exits 101 with `error: soroban-sdk requires stellar-cli v25.2.0+ to build a contract`. Native builds and unit tests are unaffected, and `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2` still signals a build system that does shake the spec, though the error no longer advertises it.

The `export` argument on `contracttype`, `contracterror`, and `contractevent` was deprecated in v27 and is now a compile error pointing at the argument itself. `soroban-spec-rust` keeps `GenerateOptions` and every `*_with_options` entry point, but the only field on the struct was `export` and is gone, leaving it empty and `non_exhaustive` so fields can be added later without breaking callers.

Migration notes are in the new `_migrating::v28_spec_shaking` page, which also absorbs the spec shaking reference material that previously lived in `_features`. No version bump is included, so the release that carries this needs to account for one.


leighmcculloch:
The semver-checks job that is failing on this pr is being removed in https://github.com/stellar/rs-soroban-sdk/pull/1969.

dmkozh:
> feature_missing lint (soroban-sdk): the experimental_spec_shaking_v2 feature was removed; any downstream crate enabling it will break ([cargo-semver ref]((doc.rustlang.org/redacted)

Unrelated to this PR, but the check output looks broken

dmkozh (approved):

mootz12 (approved):

PR #1969, opened 2026-07-30
Remove semver-checks job from rust workflow

leighmcculloch:
### What
Removes the `semver-checks` job from `.github/workflows/rust.yml`.

### Why
The `semver-checks` job runs `cargo semver-checks` as a required CI check, blocking merges on detected semver violations. This has been replaced by the new `semver` agentic workflow (`.github/workflows/semver.md` / `semver.lock.yml`), which runs the same `cargo semver-checks` analysis but labels PRs with their semver category (`semver:major`/`semver:minor`/`semver:patch`) instead of gating the merge. This is a better fit: it surfaces the same information to help maintainers make the right merge decisions, without blocking merges outright.

Follow up of https://github.com/stellar/rs-soroban-sdk/pull/1960

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,7 +22,7 @@ jobs:
 
   complete:
     if: always()
-    needs: [fmt, check-git-rev-deps, semver-checks, build, test, expand-test-wasms, build-fuzz, docs, readme, migration-docs]
+    needs: [fmt, check-git-rev-deps, build, test, expand-test-wasms, build-fuzz, docs, readme, migration-docs]
     runs-on: ubuntu-slim
     steps:
     - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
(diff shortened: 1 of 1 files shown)


dmkozh:
> This has been replaced by the new semver agentic workflow

How much do we trust this workflow? Can it silently break or e.g. report the stale results? I can see a ton of steps in Semver Impact action, but I couldn't easily find the actual semver check results.

Would it be possible to run the semver check in automated fashion (as we did before), do not block CI on its result, and pass the result to the agentic workflow? This way we'll have a static, authoritative semver check result, and then a 'massaged' classification done by the agent.

leighmcculloch:
> I couldn't easily find the actual semver check results

The cargo-semver tool gets run in the `agent` step, for example:
- https://github.com/stellar/rs-soroban-sdk/actions/runs/30580362882/job/90999006737?pr=1961#step:18:1

leighmcculloch:
We can keep the job here and just make it not block the build. But the job is not going to detect as many breaks as what the new job will, because some breakages are conceptual or behavioural, or relate to macros, which the agent has a better chance of detecting. So I would rely on the new job more than the old one.

dmkozh:
> The cargo-semver tool gets run in the agent step, for example:

I think that's already pretty good in case if we need to debug some issues (I just recall that semver tool itself breaks every few months due to updates), and I hope these won't be swallowed silently. 

I agree that the agent may be able to find more subtle breakages, but it may also may have imperfect precision/recall, whereas an automated tool has pretty consistent 'reference' behavior.

dmkozh (approved):

leighmcculloch:
> it may also may have imperfect precision

The semver tool is already very imperfect even in its own assessment of compatibility. So I think this is okay. It's a best effort, and we expect ourselves to still assess as best we can. All of this is really to catch those edge cases where it's easy to think this is a minor change where in Rust because of peculiar type constraints it's actually a definite major.

PR #1970, opened 2026-07-30
Qualify type and event names

### LINKED ISSUE #1063: Cannot use shared code between contracts

### LINKED ISSUE #1570: Improved identity of types in specs

### LINKED ISSUE #1857: Type alias for function signatures produces invalid specs

### LINKED ISSUE #1978: Optimise spec shaking v2 with the use of a graph
I propose we optimise spec shaking v2 by changing _when_ contract specs get built, and changing the contract spec XDR so that it accurately describes the graph of relationships of the functions, types, events and errors.

**Context:**

Spec shaking v2 is in the final steps of rolling out:
- https://github.com/stellar/rs-soroban-sdk/pull/1968

Spec shaking v2 is functional, but it introduced for every type, event, and error, a new marker that gets stored in the Wasm data section. The data section contains a soup of data used by the contract so it is not a simple task to post-process the section and remove the data, and so with spec shaking v2 data sections have grown. That growth was an acceptable tradeoff because spec shaking v2 significantly reduced the size of Wasms because it reduces the spec entries down to only the entries used by the contract when large libraries like OpenZeppelin's library are used.

@mootz12 identified and demonstrated that the spec shaking v2 markers could be reduced from every type, event, error down to only every event and error by building a graph data structure during proc-macro execution and embedding it into a new custom section and traversing that to identify types required then stripping that custom section during build:
- https://github.com/stellar/rs-soroban-sdk/pull/1861

**Proposal How:**

1. Change when the SDK constructs contract specs from proc-macro execution to const evaluation.

   Today they are constructed during proc-macro execution, which limits the contents of the specs to information available at the call site of a type, not the definition site.

   If we change it to const evaluation, then the specs can include information that is only known at the definition site of a type.

2. Change the names stored in the spec for each type to be fully qualified.

   Today the names stored in the spec are only the simple name, e.g. `MyType`, because at the call site that is usually the only information available.

   Once specs are constructed during const evaluation, specs can call a const function on `MyType` and ask `MyType` for it's fully qualified name, e.g. `mycrate::mymod::MyType`.

3. Change the [stellar-cli](https://github.com/stellar/stellar-cli) so that it not only spec shakes based on the presence of spec markers, but transform the contract specs into a graph, and traverses the graph to identify which types must be retained even if their marker is not present in the Wasm data section.

4. Change the [soroban-sdk](https://github.com/stellar/rs-soroban-sdk) so that it only emits spec markers when absolutely necessary, and not emitting them in all the cases that can be identified by traversing the graph.

5. Updating client code generators to handle the fully qualified names, either truncating them, or preserving them, whichever is most suitable for the environment that client SDK is used in.

**Contract Spec XDR Changes:**

This proposal requires changing the contract spec XDR, increasing the size limit of type names from 60 to upwards of 255 so that the fully qualified namespace in a typical Rust program can be included.

This proposal also requires clients to be updated to handle that the type names are fully qualified.

**Expected saving:**

TBD

leighmcculloch:
> [!NOTE]
>
> Discussion is here:
>
>
> 1. stellar/stellar-xdr#312
> 1. stellar/rs-soroban-sdk#1970 ← this PR
> 1. stellar/stellar-cli#2674
>
>
> 1. stellar/stellar-cli#2720

### What

Name a user-defined type in the contract spec by its Rust path rather than by its identifier. Each type gains a `spec_type_name` const fn returning `module_path!()` then its own name, and both the type's own definition entry and every reference to it take the name from that fn.

#### Generated Type Qualified Names and Spec XDR

Screenshot 2026-08-07 at 11 31 46 pm

### Why

A bare name is not enough to tell two user-defined types apart: two crates, or two modules of one crate, can each define a type with the same name, and a spec that names both simply cannot distinguish between them.

This was once rare, but as contracts on Soroban are getting larger, and as libraries are showing up in the ecosystem, this is becoming more common. This does happen with error types more frequently but the impact has been minimal to none in practice.

Close https://github.com/stellar/rs-soroban-sdk/issues/1570
Close https://github.com/stellar/rs-soroban-sdk/issues/1857
Close https://github.com/stellar/rs-soroban-sdk/issues/1063
Close https://github.com/stellar/rs-soroban-sdk/issues/1978


soroban-sdk/src/tests/contract_udt_struct_aliased_import.rs
```diff
@@ -0,0 +1,152 @@
...
+            ScSpecUdtStructFieldV0 {
+                doc: "".try_into().unwrap(),
+                name: "inner".try_into().unwrap(),
+                type_: ScSpecTypeDef::Udt(ScSpecTypeUdt {
+                    // See module doc comment: resolved through the type
+                    // system to the referenced type's fully qualified
+                    // definition path — the alias `Renamed` never appears.
+                    name: "soroban_sdk::tests::contract_udt_struct_aliased_import::inner::Inner"
+                        .try_into()
+                        .unwrap(),
```
leighmcculloch, line 118: @mootz12 Using type_name means that this PR also fixes the type aliasing issue demonstrated in the following PR that you pointed out:
- #1971

Issue #1944, opened 2026-07-21
Approve event field `expiration_ledger` should be named `live_until_ledger`


leighmcculloch:
### What version are you using?

This is present in the current release of `soroban-token-sdk` (27.0.1) and prior published versions, and remains on `main`, for example at commit 19a2d480fffa003e739db7cbee0249111dbfd05c.

### What did you do?

Reviewed the `Approve` event defined by `soroban-token-sdk`, whose data field carrying the ledger through which the approval remains valid is named `expiration_ledger`: https://github.com/stellar/rs-soroban-sdk/blob/19a2d480fffa003e739db7cbee0249111dbfd05c/soroban-token-sdk/src/events.rs#L3-L11

### What did you expect to see?

The field named `live_until_ledger`, matching the terminology used in [SEP-41](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0041.md), which names this parameter on the `approve` function `live_until_ledger`. OpenZeppelin's Stellar contracts library already emits this field under that name, so the two libraries currently disagree on the field name for the same event: https://github.com/OpenZeppelin/stellar-contracts/blob/v0.7.2/packages/tokens/src/fungible/mod.rs#L400-L410

### What did you see instead?

The field is named `expiration_ledger`.

### Additional considerations

Renaming this field is a breaking change, so it could not land before v28.

Ideally this would happen as close as practical to the SEP-41 change discussed in https://github.com/orgs/stellar/discussions/1948. When that change merges, the `Approve` event is expected to become a map, at which point this field name would find its way into the event data itself, so aligning the two would avoid renaming twice and let the map carry the correct key from the outset.

Alternatively, this could perhaps be treated as an urgent bug fix and released in a patch. That is usually avoided in Rust, since dependents assume crates compile across patch versions, but it might be acceptable here given that `soroban-token-sdk` sits outside the main SDK and toward the edge of the dependency tree.

### Related

- https://github.com/stellar/rs-soroban-sdk/pull/1932 cc @fnando 
- https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1784138451204279

PR #1984, opened 2026-08-03
Fix StellarAssetInterface shared-function event docs

### LINKED ISSUE #1979: StellarAssetInterface transfer/burn/approve docs still describe pre-CAP-67 events, and the spec-subset invariant blocks the doc-comment fix
## Finding

Five `StellarAssetInterface` doc-comments in `soroban-sdk/src/token.rs` still describe pre-CAP-67 Stellar Asset Contract events. Unlike the three that #1955 covered, **these cannot be fixed by editing the doc-comments**, and I think that is worth recording before someone tries.

`soroban-env-host` `builtin_contracts/stellar_asset_contract/event.rs` builds every SAC topic vec ending in `read_name(e)?`. All seven constructors do it, without exception — so every SAC event carries the trailing SEP-11 asset topic. Verified against `soroban-env-host` 27.0.1, the version this workspace pins.

| fn | documented today | actually emitted |
| --- | --- | --- |
| `approve` | `["approve", from, spender]`, `data = [amount, expiration_ledger]` | `["approve", from, to, sep0011_asset]`, `data = [amount, live_until_ledger]` |
| `transfer` | `["transfer", from, to]`, `data = { to_muxed_id: Option<u64>, amount }: Map` | `["transfer", from, to, sep0011_asset]`, `data = amount: i128` |
| `transfer_from` | `["transfer", from, to]`, `data = amount` | `["transfer", from, to, sep0011_asset]`, `data = amount` |
| `burn` | `["burn", from]`, `data = amount` | `["burn", from, sep0011_asset]`, `data = amount` |
| `burn_from` | `["burn", from]`, `data = amount` | `["burn", from, sep0011_asset]`, `data = amount` |

Two further details on `transfer`, both from `get_amount_data_maybe_muxed`:

```rust
let data: Val = if to_mux_id.is_none() {
    amount.try_into_val(e)?
} else {
    let mut map = e.map_new()?;
    map = e.map_put(map, Symbol::try_from_small_str("amount")?.into(), ...)?;
    if let Some(to_mux_id) = to_mux_id {
        map = e.map_put(map, Symbol::try_from_val(e, &"to_muxed_id")?.into(), ...)?;
```

- The doc has the **default and the exception inverted**. The bare `i128` is the ordinary case; the `Map` appears only when `to` carries a mux id. The current comment presents the Map as the norm and the scalar as a "legacy implementations may emit" fallback.
- `to_muxed_id: Option<u64>` is never emitted. The key is either present or the data is not a Map at all. The map is also built `amount` first, then `to_muxed_id`.

Also undocumented anywhere in the trait: `transfer_maybe_with_issuer` dispatches away from `transfer` entirely — `from == to` emits `transfer`, `from` being the issuer emits **`mint`**, `to` being the issuer emits **`burn`** (dropping any mux id). A reader following the current comment expects a `transfer` event that is never emitted.

## Why the obvious fix does not work

`StellarAssetInterface` is a superset of `TokenInterface`:

```
TokenInterface        = { allowance, approve, balance, transfer, transfer_from,
                          burn, burn_from, decimals, name, symbol }
StellarAssetInterface = TokenInterface ∪ { set_admin, admin, set_authorized,
                                           authorized, mint, clawback }
```

and `stellar-asset-spec/src/tests/spec.rs` enforces:

```rust
assert!(
    token_entries.is_subset(&stellar_asset_entries),
    "StellarAssetSpec is missing entries from TokenSpec"
);
```

`ScSpecEntry` carries the doc string, so the ten shared functions must stay **byte-identical** across the two traits. Appending `sep0011_asset` to the SAC copies breaks that assertion; I confirmed this by making the edit and watching the test fail. Making the same edit in `TokenInterface` is not an option either — a generic SEP-41 token has no SEP-11 asset to append, so the asset-free shapes there are correct.

That also explains #1955's scope, which I initially misread as an oversight: `set_authorized`, `mint`, and `clawback` are precisely the SAC-only functions among the incorrect ones. The narrow scope was structurally forced, not an omission.

`set_admin` is the one remaining SAC-only function still carrying a wrong shape, and it is fixable the same way — PR to follow.

## Possible directions

I do not have a recommendation I would defend; the trade-offs are yours. Recording the options I could see:

1. **Relax the invariant** to compare `ScSpecEntry` identity ignoring the doc string, so the two traits can document the same function differently. Changes what the test guarantees.
2. **Point the shared SAC docs at the SAC page** rather than restating shapes — a sentence like "see the Stellar Asset Contract docs; SAC events append a trailing `sep0011_asset` topic" would be equally byte-identical in both traits and stop teaching the wrong shape, at the cost of not stating it inline.
3. **Accept it in the SDK and fix it downstream.** `docs/tokens/stellar-asset-contract.mdx` on `stellar/stellar-docs` embeds a copy of this block, and that copy is not bound by the subset invariant.
4. **Split the traits' doc surfaces** so the SAC entries are generated separately. Largest change; probably not worth it for doc text.

Option 3 is the one that leaves the SDK inaccurate, so it is worth being explicit rather than defaulting into it.

## What I did not verify

- Whether the subset invariant is load-bearing for a consumer, or is a convenience assertion. I read only the test, not its history or dependents.
- Whether the docs copy in `stellar/stellar-docs` is hand-synced or generated. It matters for option 3 and I could not establish it from this repo alone.

## References

- CAP-0067 — SAC event shapes, admin-topic removal, muxed data map
- CAP-0046-06 — `transfer`/`burn` topic tuples with `sep0011_asset`
- #1955 / #1956 — the SAC-only half, merged 2026-07-27
- `soroban-env-host` 27.0.1 `builtin_contracts/stellar_asset_contract/event.rs`
- Host test `test_transfer_with_issuer` asserts the bare `i128` data for an ordinary transfer, and the issuer dispatch above

leighmcculloch:
### What

Correct the event doc-comments on the five `StellarAssetInterface` functions shared with `TokenInterface` (`approve`, `transfer`, `transfer_from`, `burn`, `burn_from`) to match what the Stellar Asset Contract actually emits, and relax the spec-subset invariant so the two traits can document events differently.

| fn | env `event.rs` | what changed |
| --- | --- | --- |
| `approve` | [L28](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L28) | Added `sep0011_asset: String` topic; renamed data field `expiration_ledger` → `live_until_ledger` |
| `transfer` | [L47](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L47), [L94](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L94) | Added `sep0011_asset: String` topic; corrected default/exception framing (bare `i128` is the norm, `Map` only when `to` carries a mux id); fixed map key order (`amount` first, then `to_muxed_id`); noted issuer-routing dispatch to `mint`/`burn` |
| `transfer_from` | [L47](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L47), [L94](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L94) | Added `sep0011_asset: String` topic |
| `burn` | [L163](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L163) | Added `sep0011_asset: String` topic |
| `burn_from` | [L163](https://github.com/stellar/rs-soroban-env/blob/88cadc320cd2030bae2110036b361ff474776db1/soroban-env-host/src/builtin_contracts/stellar_asset_contract/event.rs#L163) | Added `sep0011_asset: String` topic |

### Why

The SAC always appends a trailing `sep0011_asset` topic to every event; the old docs did not mention that for these five functions.

Fixing their doc-comments directly broke `test_stellar_asset_spec_includes_token_spec` because the test expects the stellar asset contract's spec to be a true superset, but now that the docs and event structures are difference, required rewriting the test to achieve the same goal without requiring the docs to match and requiring the events to be a superset of data.

Close #1979

### Known limitations

`soroban-token-sdk::events::Approve` still uses the field name `expiration_ledger` (not `live_until_ledger`); renaming it is tracked in #1944 as a separate potentially-breaking change. It should be changed in the upcoming v28 release.

### SemVer Change

### DIFF
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -146,7 +146,7 @@ pub trait TokenInterface {
     ///
     /// Emits an event with:
     /// * topics `["transfer", from: Address, to: Address]`
-    /// * data `{ amount: i128, to_muxed_id: u64 }: Map`
+    /// * data `{ amount: i128, to_muxed_id: Option<u64> }: Map`
     ///
     /// Legacy implementations may emit an event with:
     /// * topics `["transfer", from: Address, to: Address]`
@@ -309,7 +309,7 @@ pub trait StellarAssetInterface {
     ///
     /// When `to` is a muxed address, emits an event with:
     /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
-    /// * data `{ amount: i128, to_muxed_id: u64 }: Map`
+    /// * data `{ amount: i128, to_muxed_id: Option<u64> }: Map`
     ///
     /// Otherwise, emits an event with:
     /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -315,8 +315,8 @@ pub trait StellarAssetInterface {
     /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
     /// * data `amount: i128`
     ///
-    /// When `from` is the asset issuer, a `mint` event is emitted instead.
-    /// When `to` is the asset issuer, a `burn` event is emitted instead.
+    /// When `from != to` and `from` is the asset issuer, a `mint` event is emitted instead.
+    /// When `from != to` and `to` is the asset issuer, a `burn` event is emitted instead.
     fn transfer(env: Env, from: Address, to: MuxedAddress, amount: i128);
 
     /// Transfer `amount` from `from` to `to`, consuming the allowance that
--- a/soroban-sdk/src/token.rs
+++ b/soroban-sdk/src/token.rs
@@ -122,7 +122,7 @@ pub trait TokenInterface {
     /// # Events
     ///
     /// Emits an event with topics `["approve", from: Address,
-    /// spender: Address], data = [amount: i128, expiration_ledger: u32]`
+    /// spender: Address], data = [amount: i128, live_until_ledger: u32]`
     fn approve(env: Env, from: Address, spender: Address, amount: i128, live_until_ledger: u32);
 
     /// Returns the balance of `id`.
@@ -146,7 +146,7 @@ pub trait TokenInterface {
     ///
     /// Emits an event with:
     /// * topics `["transfer", from: Address, to: Address]`
-    /// * data `{ to_muxed_id: Option<u64>, amount: i128 }: Map`
+    /// * data `{ amount: i128, to_muxed_id: u64 }: Map`
     ///
     /// Legacy implementations may emit an event with:
     /// * topics `["transfer", from: Address, to: Address]`
@@ -284,7 +284,8 @@ pub trait StellarAssetInterface {
     /// # Events
     ///
     /// Emits an event with topics `["approve", from: Address,
-    /// spender: Address], data = [amount: i128, expiration_ledger: u32]`
+    /// spender: Address, sep0011_asset: String], data = [amount: i128,
+    /// live_until_ledger: u32]`
     fn approve(env: Env, from: Address, spender: Address, amount: i128, live_until_ledger: u32);
 
     /// Returns the balance of `id`.
@@ -306,13 +307,16 @@ pub trait StellarAssetInterface {
     ///
     /// # Events
     ///
-    /// Emits an event with:
-    /// * topics `["transfer", from: Address, to: Address]`
-    /// * data `{ to_muxed_id: Option<u64>, amount: i128 }: Map`
+    /// When `to` is a muxed address, emits an event with:
+    /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
+    /// * data `{ amount: i128, to_muxed_id: u64 }: Map`
     ///
-    /// Legacy implementations may emit an event with:
-    /// * topics `["transfer", from: Address, to: Address]`
+    /// Otherwise, emits an event with:
+    /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
     /// * data `amount: i128`
+    ///
+    /// When `from` is the asset issuer, a `mint` event is emitted instead.
+    /// When `to` is the asset issuer, a `burn` event is emitted instead.
     fn transfer(env: Env, from: Address, to: MuxedAddress, amount: i128);
 
     /// Transfer `amount` from `from` to `to`, consuming the allowance that
@@ -335,8 +339,8 @@ pub trait StellarAssetInterface {
     ///
     /// # Events
     ///
-    /// Emits an event with topics `["transfer", from: Address, to: Address],
-    /// data = amount: i128`
+    /// Emits an event with topics `["transfer", from: Address, to: Address,
+    /// sep0011_asset: String], data = amount: i128`
     fn transfer_from(env: Env, spender: Address, from: Address, to: Address, amount: i128);
 
     /// Burn `amount` from `from`.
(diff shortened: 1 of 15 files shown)


soroban-sdk/src/token.rs
```diff
@@ -306,13 +307,16 @@ pub trait StellarAssetInterface {
     ///
     /// # Events
     ///
-    /// Emits an event with:
-    /// * topics `["transfer", from: Address, to: Address]`
-    /// * data `{ to_muxed_id: Option<u64>, amount: i128 }: Map`
+    /// When `to` is a muxed address, emits an event with:
+    /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
```
mootz12, line 311: I don't think this complies with SEP-0041. It doesn't document that there can be flexibility in topics, and it's not unexpected that consumers do `topic.length == 3` before trying to parse

soroban-sdk/src/token.rs
```diff
@@ -306,13 +307,16 @@ pub trait StellarAssetInterface {
...
+    /// * data `{ amount: i128, to_muxed_id: Option<u64> }: Map`
     ///
-    /// Legacy implementations may emit an event with:
-    /// * topics `["transfer", from: Address, to: Address]`
+    /// Otherwise, emits an event with:
+    /// * topics `["transfer", from: Address, to: Address, sep0011_asset: String]`
     /// * data `amount: i128`
+    ///
+    /// When `from` is the asset issuer, a `mint` event is emitted instead.
+    /// When `to` is the asset issuer, a `burn` event is emitted instead.
```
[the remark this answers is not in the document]
mootz12: +1, technically an issuer self transfer emits transfer still

stellar-asset-spec/src/tests/events.rs
mootz12: Do we test the spec event structures through the SAC somewhere? If not, I think I'd prefer if we invoke the contract here. It would enforce the spec matches the SAC.

leighmcculloch:
> `stellar-asset-spec` adds new event type definitions to its contract spec, which is an additive public API change

I disagree with this assessment. The change adds pub(crate) events that are not externally accessible, and they are used to update the `xdr()` not adding to it, but modifying the existing entries that were being included.

So I am manually updating this to be `semver:patch`.

mootz12 (approved):
LG2M, and good catch. Left some comments inline.

soroban-sdk/src/token.rs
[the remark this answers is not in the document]
leighmcculloch: I don't think this is a detail worth including in this part of the interface.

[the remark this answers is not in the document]
leighmcculloch: 👍🏻 I pushed the suggested fix.

stellar-asset-spec/src/tests/events.rs
leighmcculloch: Good idea, I'll add it.

soroban-sdk/src/token.rs
leighmcculloch: > I don't think this complies with SEP-0041.

I don't disagree, but this isn't new behaviour, so out of scope of this PR.

stellar-asset-spec/src/tests/events.rs
leighmcculloch: Turns out we do test, just the next block down. But not all events, so I've added tests for all events.

mootz12 (approved):

PR #2000, opened 2026-08-11
Remove lib arg on contracttype/error/event

leighmcculloch:
### What

Remove the `lib` arg from `contracttype`/`contracterror`/`contractevent`, and ignore the spec `lib` field when generating code, so `contractimport!` always generates full type definitions instead of aliasing into an external crate.

### Why

Follow-up to #1951, which deprecated the arg because it was undocumented, never shipped in generated bindings, and has [zero usage on mainnet](https://github.com/stellar-experimental/contract-wasms/blob/main/analysis/spec-contracttype-lib-usage.csv). Setting `lib` now fails to compile with an unknown-field error, and the field remains in the spec XDR but is always empty and ignored.

### Known limitations

N/A

### SemVer Change


leighmcculloch:
> Targeting v29.

This change is targeting v29, because the `lib` feature was only deprecated in the upcoming v28 via:
- https://github.com/stellar/rs-soroban-sdk/pull/1951

However, it's occurred to me that the `lib` values never rendered into a accessible item anyway because of a bug at the following line where the type alias was code generated without a `pub` on the front:
https://github.com/stellar/rs-soroban-sdk/blob/c49bf9c37361148eff36cd05167b560b5035bb3f/soroban-spec-rust/src/types.rs#L71

This discovery, along with the fact there is [zero usage on mainnet](https://github.com/stellar-experimental/contract-wasms/blob/main/analysis/spec-contracttype-lib-usage.csv) leaves me to think we can simply remove it in v28 without the deprecated-in-v28-removed-in-v29 dance.

@mootz12 @dmkozh anyone opposed for this change targeting v28?

soroban-spec-rust/src/types.rs
```diff
@@ -65,12 +65,7 @@ pub fn generate_struct_with_options(
 ) -> Result<TokenStream, GenerateError> {
     let ident = str_to_ident(&spec.name)?;
 
-    if spec.lib.len() > 0 {
-        let lib_ident = str_to_ident(&spec.lib)?;
-        Ok(quote! {
-            type #ident = ::#lib_ident::#ident;
-        })
-    } else if spec
+    if spec
```
leighmcculloch: Re Copilot's comment on this line: added in f28fa90.

`types.rs` now has `test_generate_{struct,union,enum,error_enum,event}_ignores_lib`, each building a spec with `lib: "libname"` and asserting the generator emits the full definition rather than a `::libname::X` alias.

Confirmed they catch the regression: temporarily restoring the alias branch in `generate_struct_with_options` makes `test_generate_struct_ignores_lib` fail.

---

soroban-sdk-macros/src/derive_error_enum_int.rs
```diff
@@ -66,7 +65,7 @@ pub fn derive_type_error_enum_int(
     // Compute spec XDR once.
     let spec_entry = ScSpecEntry::UdtErrorEnumV0(ScSpecUdtErrorEnumV0 {
         doc: docs_from_attrs(attrs),
-        lib: lib.as_deref().unwrap_or_default().try_into().unwrap(),
+        lib: Default::default(),
```
mootz12: (picky) a comment as to why this field is `Default::default()` might be nice for future reviewers, (ditto for other derive_* locations)

soroban-sdk-macros/src/lib.rs
mootz12, line 1: The PR description noted this earned a migration guide note. I'm not sure the removal of a deprecated field needs a migration guide doc, but leaving this comment to confirm the intention here.

mootz12 (approved):
LG2M

mootz12:
> However, it's occurred to me that the lib values never rendered into a accessible item anyway because of a bug at the following line where the type alias was code generated without a pub on the front

Does this concern matter for `error`, `union` and `enum`?

They seem to use pub, e.g:
https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-spec-rust/src/types.rs#L133

soroban-sdk-macros/src/lib.rs
leighmcculloch: It had a migration guide, but I removed it today without updating the pr description. I removed it for the same reason, there's evidence no one used it, so it seems unnecessary noise to include.

leighmcculloch:
> Does this concern matter for `error`, `union` and `enum`?

All the other items render as pub as they should. It was only the lib type alias that wasn't, for no particular reason.

soroban-sdk-macros/src/derive_error_enum_int.rs
leighmcculloch: 9234336

PR #2004, opened 2026-08-13
Add first-class proptest support

### LINKED ISSUE #1930: Make `Arbitrary` support for Soroban types usable and discoverable outside `cargo-fuzz` (e.g. proptest)

Contract developers writing property-based tests with [proptest](https://proptest-rs.github.io/proptest/) can't easily generate Soroban values like `Address`, and generally conclude the SDK doesn't support it at all.

The SDK *does* have generation support: `soroban_sdk::testutils::arbitrary` implements `SorobanArbitrary` for `Address` (and all contract types) via the `Prototype` pattern — `ArbitraryAddress` implements `arbitrary::Arbitrary` and converts to `Address` with `IntoVal`/`FromVal` given an `Env`. But in practice it's a gap for proptest users:

1. **Discoverability.** The entire `testutils::arbitrary` module — docs, examples, and framing — is written exclusively around `cargo-fuzz` and `fuzz_target!`. A developer reaching for proptest (arguably the more common property-testing tool for contract logic, since it runs under plain `cargo test`) finds nothing, and reasonably concludes `Address` "is not `Arbitrary`".
2. **No documented proptest bridge.** Proptest consumes `Strategy` values, not `arbitrary::Arbitrary`. The prototype types can be bridged (e.g. with [`proptest-arbitrary-interop`](https://crates.io/crates/proptest-arbitrary-interop), or a hand-written `Strategy` over `ArbitraryAddress` + `into_val(&env)`), but none of this is documented or tested in the SDK, so every team rediscovers it — or misses it entirely.
3. **Generation coverage.** `ArbitraryAddress` only ever produces contract addresses (`ScAddress::Contract`), never account addresses (`ScAddress::Account`), so even fuzz users don't exercise the account-address branch of contract logic. (`ArbitraryMuxedAddress` does cover muxed ed25519 accounts, added in CAP-71 support.)

1. **Decide on first-class `Strategy` support.** An optional feature (e.g. `testutils-proptest`) providing `Strategy` implementations or an `arb::<T>()` helper for `SorobanArbitrary` types, so proptest users get the same ergonomics `cargo-fuzz` users already have.
2. **Document the proptest path.** Extend the `testutils::arbitrary` module docs with a section showing how to generate Soroban values from proptest — the new helper if #1 is accepted, otherwise the interop-crate recipe (generate `<T as SorobanArbitrary>::Prototype`, convert with `into_val(&env)` inside the test). A doctest-backed example for `Address` and a `#[contracttype]` struct would prevent the "not `Arbitrary`" misconception.
3. **Broaden `ArbitraryAddress` coverage** to also generate `ScAddress::Account` variants, so generated inputs exercise both address kinds.

- **Status quo:** each contract repo hand-rolls its own workaround, e.g. pools of `Address::generate(&env)` values indexed by generated `usize`s. It works but hides the SDK's existing tooling and skips shrinking/coverage over the address values themselves.
- **Docs-only fix:** skip item 1 and only document the interop recipe. Cheaper, and may be enough — the core issue is discoverability.
- **Implement `arbitrary::Arbitrary` directly on `Address`:** not possible as-is, since constructing host objects requires an `Env`; that constraint is exactly why `SorobanArbitrary`'s prototype pattern exists.

leighmcculloch:
### What

Add a `testutils::proptest` module, behind a new `testutils-proptest` feature, implementing `proptest`'s `Arbitrary` trait for the prototypes of every `SorobanArbitrary` type, so a prototype can be named in the parameter list of a `proptest!` test.

### Why

The `Arbitrary` support for Soroban types was only reachable through `cargo-fuzz`, so proptest users had to discover and wire up `proptest-arbitrary-interop` themselves against an undocumented prototype pattern.

Close #1930
### SemVer Change

### TODO

- [x] Consider alternatives UXs.
- [x] Look into if we can set more reliable entry hints for collections.


soroban-sdk/src/testutils/proptest.rs
```diff
@@ -0,0 +1,313 @@
...
+/// `Bls12381G2Affine` is 192 — and a budget that fits the small ones cuts
+/// collections of the large ones short.
+fn collection_entropy_budget<T>() -> usize
+where
+    T: for<'a> ::arbitrary::Arbitrary<'a>,
+{
+    let (lower, upper) = <T as ::arbitrary::Arbitrary>::size_hint(0);
+    let element = upper.unwrap_or(UNBOUNDED_ENTROPY_BUDGET).max(lower);
+    COLLECTION_ELEMENTS * (element + 1)
+}
```
marcelosalloum, line 148: `collection_entropy_budget` only runs when the test parameter's outermost type is `ArbitraryVec`/`ArbitraryMap`. A collection nested inside a `contracttype` struct or tuple would size the buffer with `arb_from_size_hint` (`max(UNBOUNDED_ENTROPY_BUDGET, lower hint)`), so the default budget can be insufficient.

Consider letting the caller override the budget, proptest's `Parameters` is designed for this.

marcelosalloum (approved):
Thanks a ton! This looks mostly good as a first iteration. I'e added a suggestion that touches on tested collections entropy, but other than that it looks great.

As a disclaimer, I haven't used this tool extensively yet but will circle back if I identify any improvements down the line.

PR #2005, opened 2026-08-13
Implement support for externally managed contract executables (CAP-0085)

mootz12:
### What

Adds support for contracts to create and manage executable reference entries with `ExternalTag`. Blocks the use of `ExternalTag` for any UDT, event, or contract interface, and instead recommends `String` to be used.

Updates `Deployer` to allow contracts to update their executable to an executable reference entry, and deploy new contracts with an executable reference entry.

### Why

Adds support in SDK for [CAP-0085](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0085.md)

### Known limitations

None

### SemVer Change

[TODO: Check one. Merge only when confirmed that the main branch is accepting changes for the matching type of release. See the [semver-rules.md](./semver-rules.md).


soroban-sdk-macros/src/map_type.rs
```diff
@@ -108,6 +108,12 @@ pub fn map_type(t: &Type, allow_ref: bool, allow_hash: bool) -> Result<ScSpecTyp
                     "MuxedAddress" => Ok(ScSpecTypeDef::MuxedAddress),
                     "Timepoint" => Ok(ScSpecTypeDef::Timepoint),
                     "Duration" => Ok(ScSpecTypeDef::Duration),
+                    // ExecutableTag has no spec type and must be created with the `create_executable_tag`
+                    // host function for it to be used as a contract reference entry key.
+                    "ExecutableTag" => Err(Error::new(
```
dmkozh, line 113: Is this because we forgot to add the spec type to the XDR? I believe that's not the first time we miss the spec in the XDR, maybe we should decouple spec XDR from the core XDR.

soroban-sdk/src/deploy.rs
```diff
@@ -86,6 +86,61 @@
 //! # fn main() { }
 //! ```
 //!
+//! #### Deploy a contract that uses an executable reference (CAP-0085)
+//!
+//! A contract can be deployed with an executable that references an
```
dmkozh: I have trouble parsing the 'executable that reference an executable reference'. Can we say something like 'executable that references a contract data entry containing Wasm hash, instead of a direct Wasm hash'?

soroban-sdk/src/deploy.rs
```diff
@@ -86,6 +86,61 @@
...
+//! executable reference entry owned by an external contract, instead of a
+//! fixed Wasm hash. The external contract can update the executable reference
```
dmkozh: nit: I'd rather say 'The contract that owns the data entry can update...'. The caveat is that the contract doesn't technically need to be external (you can refer to your own entry), and just 'external' may imply *arbitrary* external contract.

soroban-sdk/src/deploy.rs
```diff
@@ -225,7 +280,31 @@ impl Deployer {
...
+        &self,
+        executable_owner: &Address,
+        tag: &ExecutableTag,
+    ) {
+        self.env
+            .update_current_contract_executable_ref(executable_owner.to_object(), tag.to_object())
+            .unwrap_infallible();
+    }
+
+    /// Extend the TTL of the contract instance, code, and any executable
```
dmkozh: This added comment duplicates the next sentence for the most part,  we should update the old line instead.

soroban-sdk/src/deploy.rs
```diff
@@ -274,7 +354,8 @@ impl Deployer {
             .unwrap_infallible();
     }
 
-    /// Extend the TTL of the contract instance and code with limits on the extension.
+    /// Extend the TTL of the contract instance, code and any executable reference entry
```
dmkozh: Same as line 306 comment, please deduplicate this (I see that here duplication has existed before)

soroban-sdk/src/executable_tag.rs
```diff
@@ -0,0 +1,230 @@
...
+use crate::unwrap::UnwrapInfallible;
+#[cfg(doc)]
+use crate::{deploy::Deployer, deploy::DeployerWithAddress, storage::Persistent};
+
+#[cfg(not(target_family = "wasm"))]
+use super::xdr::{ScString, ScVal};
+
+/// ExecutableTag is the key of an executable reference entry.
+///
+/// An executable reference entry is a persistent contract data entry keyed by
```
dmkozh: I read this as 'keyed by ExecutableTag  that has Wasm hash payload' (i.e. 'whose' may be read as referring to the tag, and not the entry), which isn't the case. Maybe rephrase this like 'contract data entry with ExecutableTag and Wasm hash value'

dmkozh (approved):

soroban-sdk-macros/src/map_type.rs
```diff
@@ -108,6 +108,12 @@ pub fn map_type(t: &Type, allow_ref: bool, allow_hash: bool) -> Result<ScSpecTyp
...
+                        ident.span(),
+                        "ExecutableTag has no contract spec type; use String and construct it with ExecutableTag::new instead",
+                    )),
```
leighmcculloch, line 116: So that we don't miss this again:
- https://github.com/stellar/stellar-xdr/pull/314

Can we track in a new issue resolving this in v29? I'd even prepare a PR ahead, and we should maybe start using milestones to track and not lose PRs targeting future releases.

soroban-sdk-macros/src/map_type.rs
```diff
@@ -108,6 +108,12 @@ pub fn map_type(t: &Type, allow_ref: bool, allow_hash: bool) -> Result<ScSpecTyp
                     "MuxedAddress" => Ok(ScSpecTypeDef::MuxedAddress),
                     "Timepoint" => Ok(ScSpecTypeDef::Timepoint),
                     "Duration" => Ok(ScSpecTypeDef::Duration),
+                    // ExecutableTag has no spec type and must be created with the `create_executable_tag`
+                    // host function for it to be used as a contract reference entry key.
+                    "ExecutableTag" => Err(Error::new(
+                        ident.span(),
+                        "ExecutableTag has no contract spec type; use String and construct it with ExecutableTag::new instead",
```
leighmcculloch, line 115: > has no contract spec type

Contract specs are a lower level detail about contracts, and so I don't think we should expect every contract dev will be aware of what they are or what this means. I suggest instead talking about the feature:

```suggestion
error: ExecutableTag is not supported as a parameter of contract functions; use String
```

soroban-sdk/src/tests/contract_executable_ref.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Owns executable reference entries that other contracts' executables
+/// reference.
+#[contract]
+pub struct OwnerContract;
+
+#[contractimpl]
+impl OwnerContract {
+    pub fn set_executable(env: Env, name: String, wasm_hash: BytesN<32>) {
+        let tag = ExecutableTag::new(&env, &name);
+        env.storage().persistent().set(&tag, &wasm_hash);
```
leighmcculloch, line 31: Are executable tags acceptable in other forms of storage like instance and temporary? If no, maybe we should introduce an assertion, at least a dbg assertion, that checks they aren't used in other storage forms. Although maybe that is being too parental, there might be reasons to store them there even if lookup isn't supported there. 🤔

soroban-sdk/src/auth.rs
```diff
@@ -60,6 +60,17 @@ pub struct CreateContractWithConstructorHostFnContext {
...
     Wasm(BytesN<32>),
+    /// Executable read from an entry owned by an external contract.
+    ExternalRef(ContractExecutableRef),
+}
+
+/// Reference to an executable reference entry owned by `owner` whose storage
+/// key is an `ExecutableTag` with the given `tag` value.
+#[derive(Clone, Debug)]
+#[contracttype(crate_path = "crate")]
+pub struct ContractExecutableRef {
```
leighmcculloch, line 71: I think we should remove the prefix. We don't prefix ExecutableTag.
```suggestion
pub struct ExecutableRef {
```

soroban-sdk/src/auth.rs
```diff
@@ -60,6 +60,17 @@ pub struct CreateContractWithConstructorHostFnContext {
 #[contracttype(crate_path = "crate")]
 pub enum ContractExecutable {
     Wasm(BytesN<32>),
+    /// Executable read from an entry owned by an external contract.
+    ExternalRef(ContractExecutableRef),
```
leighmcculloch, line 64:
```suggestion
    /// Executable specified by the contract instance as a specific Wasm contract code entry identified by its Wasm sha256 hash.
    Wasm(BytesN<32>),
    /// Executable reference via an persistent storage entry owned by this contract or another contract.
    ExternalRef(ContractExecutableRef),
```

soroban-sdk/src/auth.rs
```diff
@@ -60,6 +60,17 @@ pub struct CreateContractWithConstructorHostFnContext {
...
+}
+
+/// Reference to an executable reference entry owned by `owner` whose storage
+/// key is an `ExecutableTag` with the given `tag` value.
```
leighmcculloch, line 68:
```suggestion
/// Executable referenced via a persistent storage entry owned by a contract, either this contract or another contract.
///
/// The persistent storage entry owned by the owner has the tag as its key.
```

soroban-sdk/src/deploy.rs
```diff
@@ -86,6 +86,60 @@
 //! # fn main() { }
 //! ```
 //!
+//! #### Deploy a contract that uses an executable reference (CAP-0085)
```
leighmcculloch, line 89: We don't typically reference the CAPs in the docs. CAPs are temporal.
```suggestion
//! #### Deploy a contract that uses an executable reference
```

soroban-sdk/src/deploy.rs
```diff
@@ -86,6 +86,60 @@
...
+//!
+//! A contract can be deployed with an executable that references a contract data entry
+//! containing a Wasm hash, instead of a direct Wasm hash. The contract that owns the data
+//! entry can update it, causing all contracts that reference it to use the new Wasm hash
+//! value. See [ExecutableTag] for details on the entry.
+//!
+//! ```
+//! use soroban_sdk::{contract, contractimpl, Address, BytesN, Env, ExecutableTag, String};
+//!
+//! const DEPLOYED_WASM: &[u8] = include_bytes!("../doctest_fixtures/contract.wasm");
```
leighmcculloch, line 99: This is only used for the test, can you move it into the test fn so it doesn't appear as part of the contract that would need deploying? That part was initially confusing to me.

soroban-sdk/src/deploy.rs
```diff
@@ -225,14 +279,37 @@ impl Deployer {
...
     ///
-    /// The TTL is the number of ledgers between the current ledger and the final ledger the data can still be accessed.
+    /// **Important**: `executable_owner` now controls and can update the Wasm
+    /// of the current contract.
+    ///
+    /// The function won't do anything immediately. The contract executable
+    /// will only be updated after the invocation has successfully finished.
+    pub fn update_current_contract_executable_ref(
+        &self,
+        executable_owner: &Address,
```
leighmcculloch, line 297: I think we should name this owner for symmetry with the ExecutableRef type and consistency with the next parameter.
```suggestion
        owner: &Address,
```

soroban-sdk/src/auth.rs
```diff
@@ -60,6 +60,17 @@ pub struct CreateContractWithConstructorHostFnContext {
...
+#[derive(Clone, Debug)]
+#[contracttype(crate_path = "crate")]
+pub struct ContractExecutableRef {
+    pub owner: Address,
+    pub tag: String,
```
leighmcculloch, line 73: So that we don't have to break the API in v29, can you make this ExecutableTag, and automatically coerce String's to ExecutableTag in Val conversion?

soroban-sdk/src/deploy.rs
```diff
@@ -414,6 +491,46 @@ impl DeployerWithAddress {
...
+    /// `()` for contracts with no constructor or a constructor with zero
+    /// arguments.
+    ///
+    /// The address of the deployed contract is defined by the deployer address
+    /// and provided salt, the same as for [DeployerWithAddress::deploy_v2].
+    ///
+    /// Returns the deployed contract's address.
+    pub fn deploy_executable_ref<A>(
+        &self,
+        executable_owner: &Address,
```
leighmcculloch, line 515: Same reason as above.
```suggestion
        owner: &Address,
```

soroban-sdk/src/deploy.rs
```diff
@@ -414,6 +491,46 @@ impl DeployerWithAddress {
...
+        tag: &ExecutableTag,
```
leighmcculloch, line 516: Pondering if we should use the ExecutableRef contract type here. It tells a clearer narrative. And any cost lost to building another object is super minimal because these types are u64 integers, and so I expect that will all or almost all be completely optimised away. Yeah I think so. Any particular reason we don't use an ExecutableRef here and the other function that accepts both?

soroban-sdk/src/executable_tag.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+impl ExecutableTag {
+    #[inline(always)]
+    pub(crate) unsafe fn unchecked_new(env: Env, obj: ExecutableTagObject) -> Self {
+        Self { env, obj }
+    }
+
+    /// Create the tag that keys the executable reference entry named `tag`.
+    pub fn new(env: &Env, tag: &String) -> ExecutableTag {
```
leighmcculloch, line 198: For consistency with existing APIs on other types:
```suggestion
    pub fn from_string(env: &Env, tag: &String) -> ExecutableTag {
```

For example: https://docs.rs/soroban-sdk?search=from_string

soroban-sdk/src/executable_tag.rs
```diff
@@ -0,0 +1,230 @@
...
+///
+/// The protocol enforces additional checks on an entry keyed by an [ExecutableTag]:
+///
+/// - The entry must be stored with persistent durability. Any other
+///   durability panics.
+/// - The value must be the 32-byte hash of Wasm that has already been
+///   uploaded. Storing any other value panics.
+/// - The entry cannot be removed. [Persistent::remove] with a tag key panics.
+#[derive(Clone)]
+pub struct ExecutableTag {
```
leighmcculloch, line 33: We should implement SorobanArbitrary for the type for use in fuzzing.

See https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk/src/testutils/arbitrary.rs

leighmcculloch (changes requested):
Looks good. Few requests inline.

soroban-sdk/src/tests/contract_executable_ref.rs
leighmcculloch: I remember the host checks this, so if that error is clear enough no need to do anything here.

dmkozh: Host does the check, so I don't think we should need to check this additionally on the guest side. FWIW unlike muxed addresses, we allow executable tags to be nested in the arbitrary ledger keys. The check is there only for the keys that are specifically of ExecutableTag type, just in order to avoid ambiguity and confusion.

soroban-sdk/src/auth.rs
dmkozh: This is going to be annoying to do though, as the host supplies String here. I'm passing String here intentionally - it mirrors the underlying XDR and also provides more functionality than ExectuableTag (which is really meant to only be used as the storage tag). I still can see an argument for this being ExectuableTag, but it's too late to change at this point, and I'm not sure if it's worth the hassle to re-create the struct in the SDK.

leighmcculloch: Ok if that's the case then I think we should limit the executable tag storage type as a storage detail, and discourage / not use it elsewhere:
- not expose executable tag in specs
- not export the type at all in the sdk
- add a specific function for storing executable refs that take in specifically a String and a BytesN<32> and internally turns that into the right object and stores it, and a getter.

This will have a few advantages:
- The sdk won't offer a way to delete the entries
- The sdk won't offer a way to write the wrong value
- There won't be a dual String vs ExecutableTag dance

leighmcculloch: > * not expose executable tag in specs

- https://github.com/stellar/stellar-xdr/pull/315

dmkozh: This is tempting, but the main purpose of the tag type is really to avoid an additional bespoke set of storage access functions, e.g. it's possible to check if tag exists, get its TTL in tests, or extend the TTL of specifically the tag entry (e.g. you might want that as the manager contract). All these are achieved easily with the existing host (and SDK) functions given a tag key. If we're fine with restricting the possible functionality partially, or introducing a parallel set of SDK functions that we think are useful, then we could indeed restrict the interface.

leighmcculloch: > * add a specific function for storing executable refs that take in specifically a String and a BytesN<32> and internally turns that into the right object and stores it, and a getter.

A separate API like this would be really nice at the SDK and tell a clearer narrative.

There's maybe a small footgun that doing so would hide that it shares the storage namespace. But as long as we don't make an ExecutableTag value available to the contract developer that should cover most cases. The only case not covered is if someone is storing raw Val's, but there's very few situations you could do that without a contract being easily exploited/mutated by anyone.

leighmcculloch: > avoid an additional bespoke set of storage access functions

I think we should experiment with this to evaluate UX impact rather than have avoiding bespoke APIs as a goal at this layer. If the overall UX is better without a bespoke API then 👍🏻. We should consider for example: What's the cognitive load of the new API UX as a whole. Does combining them create more or less confusion for both regular storage use and external ref setup.

I suspect a bespoke API for this in the SDK will be less cognitive load on understanding how to use it correctly with less disruption of regular storage docs and easier to tell the full narrative in a single vertical of the SDK rather than scattered. But an experiment might prove that wrong for sure.

dmkozh: > I think we should experiment with this to evaluate UX impact rather than have avoiding bespoke APIs as a goal at this layer. 

To be clear, my comment refers to the host interface in the first place. I don't have a strong opinion about the SDK interface, and it's also much easier to add more functions to the SDK if we miss any (unlike host fns). In any case, I think we should consider the implementation for a few realistic use cases beyond reference update, such as checking the reference value (useful at least in tests), or extending only the reference entry TTL.

soroban-sdk-macros/src/map_type.rs
mootz12: It's not just contract function parameters, its events and UDT's as well, anything that ends up in the spec.

I'll consider if there is some better wording here.

soroban-sdk/src/executable_tag.rs
mootz12: This was omitted since the type cannot be used as an input to anything, and `String` can already be fuzzed.

soroban-sdk/src/auth.rs
mootz12: I originally kept it in line with the host so it's clearer to the dev what is actually going on. Though, I do agree the SDK impl of `ExecutableTag` and how it's exposed / used only by certain contracts could be cleaned up.

There is some merit to the pattern as we have bespoke SDK functions for writing the contracts executable value in the instance already. Let me think on what the SDK would look like, and push up an alternative PR if there is a realistic option.

soroban-sdk/src/auth.rs
mootz12: This comes from here: https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/builtin_contracts/common_types.rs#L39-L45

mootz12:
Closing in favor of https://github.com/stellar/rs-soroban-sdk/pull/2017

PR #2006, opened 2026-08-13
Unpack contracttype structs sparsely (CAP-86)

leighmcculloch:
### What

Unpack `#[contracttype]` structs with the sparse map host function, so a field absent from the map unpacks as void and therefore as `None` for an `Option` field, and a map key that is not a field of the struct is ignored, both from a `Val` and from the `ScVal` and `ScMap` XDR types.

### Why

A map had to match the struct exactly, so a contract trapped rather than reading data stored by a different version of itself, making it impossible to add or remove a field without migrating every stored value first, as described in [CAP-86](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0086.md) and [discussion 1877](https://github.com/orgs/stellar/discussions/1877).

### Known limitations

Packing is left strict and still writes every field, since omitting void fields on write breaks cross-contract APIs when only one side upgrades. I am experimenting with different approaches we might take leveraging the sparse writing, possibly just for events. For example:
- https://github.com/stellar/rs-soroban-sdk/pull/2008

### SemVer Change

### TODO

- [x] Look for test gaps, maybe use `cargo-mutants`.
- [ ] ~Look into event only write optimisation, maybe.~ (Am experimenting with this separately: https://github.com/stellar/rs-soroban-sdk/pull/2008)
- [x] Deeper look at OZ failing test.


soroban-sdk/src/_migrating/v28_contracttype_unpacking.rs
```diff
@@ -0,0 +1,129 @@
...
+//! ```
+//!
+//! ## Migrating
+//!
+//! No code changes are required for most contracts, although the subtle change in behaviour could
+//! be observable on contracts that expect type unpacking to fail in the cases that are now
+//! tolerated. Contracts that only round trip their own data see no difference, because a map
+//! written by the same struct always contains exactly its fields.
+//!
+//! Review any code that relied on unpacking failing to detect a mismatch. A struct where every
```
leighmcculloch, line 118: Investigating the CI failure on `test-crate (examples/multisig-smart-account/webauthn-verifier)`:

A cursory look suggests OpenZeppelin's `verify_sig_data_invalid_fails` feeds a malformed `WebAuthnSigData` map and asserts the error contains `Error(Object, UnexpectedSize)`, which is what strict unpack raised when the map's length didn't match the field count. Sparse unpack no longer length-checks, so the missing field unpacks as void and the contract's `.expect("WebAuthnSigData with correct format")` panics with `ConversionError` instead. That different shouldn't be observable to a contract. So this suggests this isn't a big issue, but interesting to see we quickly discovered a case where an observable change in behaviour is seen by a test.

leighmcculloch (commented):
.

dmkozh (approved):

PR #2008, opened 2026-08-14
Pack contractevent (data format = map) sparsely (CAP-86)

leighmcculloch:
### What

Build the data map of a `#[contractevent(data_format = "map")]` event with the sparse map host function, so a field whose value is void, which an `Option` field that is `None` is, is omitted from the map instead of being written with a void value.

### Why

Every event with an unset optional field pays for a key and a void value that carries no information.

It will be easier to define a single event that supports a variety of shapes for contracts that surface different information in different situations. A good example of this is the transfer event in SEP-41.

Discussion:
- https://github.com/stellar/stellar-protocol/discussions/1877

Follows:
- https://github.com/stellar/rs-soroban-sdk/pull/2006

### SemVer Change


dmkozh (approved):

soroban-sdk/src/_migrating/v28_contractevent_packing.rs
mootz12, line 1: I'm not sure sparse packing only `contractevent` structs makes much sense. It's an inconsistency with how all other types are packed, with no real benefit.

I don't think `Option<T>` types are common enough in events to warrant the change/difference. I think I prefer that the way types are packed remains consistent across the SDK.

Secondly, I'm not sure this claim is true:
> sparse unpacking in the base branch means a consumer decoding the event into a struct reads the omitted field back as None either way

Consumers of events are often off-chain, and are manually reading keys from event data.

soroban-sdk-macros/src/derive_event.rs
```diff
@@ -283,7 +283,7 @@ fn derive_impls(args: &ContractEventArgs, input: &DeriveInput) -> Result<TokenSt
             ).into_val(env)
         },
         DataFormat::Map => {
-            // Must be sorted for map_new_from_slices. Sort by the spec name (the
+            // Must be sorted for sparse_map_new_from_slices. Sort by the spec name (the
```
mootz12: This doc on `lib.rs` should be updated based on the behavior of sparse map -> https://github.com/stellar/rs-soroban-sdk/blob/contractevent-sparse-map-new/soroban-sdk/src/lib.rs#L818-L821

mootz12 (commented):
Implementation looks good, but have some concerns about the deviation this introduces.

soroban-sdk/src/_migrating/v28_contractevent_packing.rs
leighmcculloch: > > sparse unpacking in the base branch means a consumer decoding the event into a struct reads the omitted field back as None either way
> 
> Consumers of events are often off-chain, and are manually reading keys from event data.

Where do you see this? This was a statement in an early iteration I asked an agent to do and I removed it myself when I took over from the agent.

leighmcculloch: Ah it's in the PR description still, removed.

soroban-sdk-macros/src/derive_event.rs
leighmcculloch: 1afe86e

leighmcculloch:
@mootz12 discussed this offline and pushed back on it for similar reasons I originally pushed back on it when we decided we wouldn't make the contracttype change. So we concluded to defer making this change right now to such time as we can do both contracttype and contractevent together.

Introducing an inconsistency into how maps encode across events and types is surprising. Whilst we don't expect it to create an issue, it is still surprising.

dmkozh:
> Introducing an inconsistency into how maps encode across events and types is surprising. Whilst we don't expect it to create an issue, it is still surprising.

If being surprising is the main issue, then how about an opt-in approach to both? Either a new macro, or a macro parameter would work. I don't really see a way forward to ever adopting this without having some contracts switch to this.

leighmcculloch:
> how about an opt-in approach to both? Either a new macro, or a macro parameter would work.

I've updated the PR to be opt-in in the form `#[contractevent(sparse = true)]`. It works, but the spec won't be able to communicate about if an event is sparse or not, so any downstream will need to assume that any event could be sparse anyway.

leighmcculloch:
> how about an opt-in approach to both? Either a new macro, or a macro parameter would work.

I've updated the PR to be opt-in in the form `#[contractevent(sparse = true)]`.

It works, but the spec won't be able to communicate about if an event is sparse or not, so any downstream will need to assume that any event could be sparse anyway.

Because of this, I don't think the opt-in makes much sense. While it will make sense to the developer writing the contract, the developer the surprise will occur for is the consumer of the events who is someone else.

We could defer this change to v29, introduce a sparse field to the type and event spec entries. But a change to the spec should come with some weight in benefit and the field to address the inconsistency feels more like a formality rather than something that will have a meaningful impact.

So I think we should either:
1️⃣ Defer, and continue to consider options.
2️⃣ Move ahead with contractevent only sparse packing. The inconsistency is unfortunate, and we wouldn't create it today if we were designing events from scratch, but in the context of a legacy system it seems okay.

I'm okay with 2️⃣, but if anyone wants to do 1️⃣ we can hold it off to another release.

dmkozh:
> While it will make sense to the developer writing the contract, the developer the surprise will occur for is the consumer of the events who is someone else.

That's kind of debatable. Event consumers don't exist in a vacuum (unless we're talking about blanket ingestion of all events that doesn't care about the event schema anyways). For the cases where event schema matters, there is a standard that every party should respect. For example, in case of SAC transfer to_muxed_id is allowed to be absent, so there is really no surprise for anyone. I realize that there might be cases where the event spec is not very well tested and thus there is a chance that the SDK change slips through unnoticed, but I'm not sure if that's necessarily the case worth optimizing for (as that's not the only possible failure mode for poor test coverage or poor specifications).

With that in mind, I'm feel fine about 2, though I don't mind 1 either.

soroban-sdk/src/_migrating.rs
```diff
@@ -71,8 +71,15 @@
 //!    # fn main() { }
 //!    ```
 //!
+//! 5. [Events with a map data format omit void fields when publishing][v28_contractevent_packing].
+//!    A data field of a [`contractevent`] whose value is void, as an [`Option`] field that is
```
dmkozh: there is a typo ('is, is'), but I also find this sentence hard to parse. Can we say something like 'whose value is void (`None` [`Option`], or unit type `()`) is omitted from the published map'.

We also have the same confusing sentence in lib.rs

dmkozh (approved):

PR #2010, opened 2026-08-17
Panic on test state use inside a contract

### LINKED ISSUE #2009: Env test state behaves oddly inside contract calls

latest, currently `27.0.6`

### What did you do?

Called `Address::generate()` inside a contract function.

```rust
use crate::{self as soroban_sdk};
use soroban_sdk::{contract, contractimpl, testutils::Address as _, Address, Env};

#[contract]
pub struct Contract;

#[contractimpl]
impl Contract {
    pub fn gen_address(env: Env) -> Address {
        Address::generate(&env)
    }
}

#[test]
fn test_generators_restart_on_every_invocation() {
    let env = Env::default();

    let a = Address::generate(&env);
    let b = Address::generate(&env);
    assert_ne!(a, b); // success unique addresses

    let c = client.gen_address();
    let d = client.gen_address();

    assert_eq!(c, a); // fail previous address supplied
    assert_eq!(d, a); // fail that same previous address again
}
```

A unique address to be supplied.

### What did you see instead?

A previously generated address to be supplied.

The Env in the test environment has a test state that tracks generated addresses to make generated addresses predictable. The generate function is not intended to be callable from within contracts because it is not available when a contract is built to wasm. But none the less it still exists and can be called in tests. For a developer who confusingly uses the test only generated fn inside a contract they get surprising behaviour (a repeated Address) which doesn't help them discover what they are doing incorrect.

leighmcculloch:
### What

Panic during tests when calling functions inside a contract that cannot be called reliably inside the contract instead of allowing for odd and surprising behaviour to take place.

### Why

Some test utilities don't work properly inside contract functions. This limitation doesn't usually matter because test utilities aren't usable inside contract functions when building to wasm. But none-the-less they can still be called inside contract functions during a test run.

The reason they don't work properly inside a contract is because the Env inside a contract function is not the same Env that is outside of the contract function. Outside the contract function an Env carries test state, but the Env passed into the contract function via the Host doesn't contain any of that test state.

See the tests in this PR for examples.

This has always been a limitation, but panicking on the use of these test utilities inside contract functions prevents a developer experiencing odd behaviour using the test utilities there when they haven't been designed for that.

Close #2009

### SemVer Change

Close #2009

### DIFF
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -129,8 +129,7 @@ use internal::{
 #[doc(hidden)]
 #[derive(Clone)]
 pub struct MaybeEnv {
-    // On wasm an Env is always available and costs nothing to store, so it is
-    // never absent.
+    // On wasm an Env is always available.
     #[cfg(target_family = "wasm")]
     env: Env,
     #[cfg(not(target_family = "wasm"))]
--- a/soroban-sdk/src/env.rs
+++ b/soroban-sdk/src/env.rs
@@ -240,9 +240,7 @@ thread_local! {
 struct EnvTestState {
     test_name: Option<String>,
     number: usize,
-    // Shared between clones of the Env, so that a config change is seen by
-    // every clone, including the clones held inside values such as Symbol.
-    config: Rc<RefCell<EnvTestConfig>>,
+    config: EnvTestConfig,
     generators: Rc<RefCell<Generators>>,
     auth_snapshot: Rc<RefCell<AuthSnapshot>>,
     snapshot: Option<Rc<LedgerSnapshot>>,
@@ -658,7 +656,7 @@ impl Env {
 
     /// Change the test config of an Env.
     pub fn set_config(&mut self, config: EnvTestConfig) {
-        *(*self.test_state.config).borrow_mut() = config;
+        self.test_state.config = config;
     }
 
     /// Used by multiple constructors to configure test environments consistently.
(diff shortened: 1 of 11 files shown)


soroban-sdk/src/env.rs
```diff
@@ -235,15 +235,60 @@ thread_local! {
...
+        match self {
+            Self::Test { config, .. } => config,
+            Self::Contract => panic!("the test config is unavailable inside a contract function and must be accessed only from the test code outside the contract function"),
+        }
+    }
+
+    fn generators(&self) -> &Rc<RefCell<Generators>> {
+        match self {
+            Self::Test { generators, .. } => generators,
+            Self::Contract => panic!("generating values like addresses is unavailable inside a contract function and must be accessed only from the test code outside the contract function"),
```
mootz12: (picky)

```suggestion
            Self::Contract => panic!("generating values like addresses is unavailable inside a contract function and must be done only from the test code outside the contract function"),
```

mootz12 (approved):
Do we care about blocking all test state mutators, or just specific ones that are problematic?

Looks like things like `register_at` and `mock_auths` would still work.

One minor grammar error otherwise. We can skip if no changes are needed, it's clear enough.

soroban-sdk/src/env.rs
leighmcculloch: 252522e

leighmcculloch:
> Do we care about blocking all test state mutators, or just specific ones that are problematic?
> 
> Looks like things like `register_at` and `mock_auths` would still work.

Opened the following issue to consider doing this on other fns that do not depend on the test state:
- https://github.com/stellar/rs-soroban-sdk/issues/2013

PR #2017, opened 2026-08-18
Implement support for contract executable references (CAP-85)

mootz12:
### What

Add support for contracts to manage contract executable reference entries via `env.executable_refs()`.

Updates `Deployer` to allow contracts to update their executable to an executable reference entry, and deploy new contracts with an executable reference entry.

### Why

Adds support in SDK for [CAP-0085](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0085.md)

Alternative to https://github.com/stellar/rs-soroban-sdk/pull/2005, that exposes internal details like the `ExecutableTag` type.

### Known limitations

None

### SemVer Change

- [x] Major (`vX._._`) - Breaking change to the public API.
- [ ] Minor (`v_.Y._`) - Additive change to the public API.
- [ ] Patch (`v_._.Z`) - No change to the public API.

PR #2020, opened 2026-08-20
Implement support for contract executable references and improve deploy interface (CAP-85)

mootz12:
### What

Add support for contracts to manage contract executable reference entries via `env.executable_refs()`.

Updates `Deployer` to just one `deploy_contract` and one `update_current_contract` function, deprecating the rest. This utilizes the `ContractExecutable` type as the input argument to prevent the interface expansion seen in #2017.

### Why

Adds support in SDK for [CAP-0085](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0085.md)

Alternative to https://github.com/stellar/rs-soroban-sdk/pull/2005, that exposes internal details like the `ExecutableTag` type.

### Known limitations

The auth type `ContractExecutable` is re-exported at the root to prevent breaking contracts that use this type in custom contract auth.

### SemVer Change


soroban-sdk/src/auth.rs
```diff
@@ -59,7 +59,21 @@ pub struct CreateContractWithConstructorHostFnContext {
 #[derive(Clone, Debug)]
 #[contracttype(crate_path = "crate")]
 pub enum ContractExecutable {
+    /// Executable specified by the contract instance as a specific Wasm contract code entry identified by its Wasm sha256 hash.
     Wasm(BytesN<32>),
+    /// Executable reference via a persistent storage entry owned by this contract or another contract.
+    ExternalRef(ContractExecutableRef),
+}
```
leighmcculloch: It's unfortunate that we have two types to refer to an executable. For example, `ContractExecutable` above, and `address::Executable`:

https://github.com/stellar/rs-soroban-sdk/blob/d0cfe54ea5ed7a8fe86207eff4ebe9eadb86a710/soroban-sdk/src/address.rs#L219-L223

Do we ever need to support comparison between the types? We probably can't add comparison because it's a bit unclear to me how it would compare refs and wasm-hashes.

soroban-sdk/src/lib.rs
```diff
@@ -1253,6 +1254,7 @@ pub mod storage;
 pub mod token;
 mod vec;
 pub use address::{Address, Executable};
+pub use auth::{ContractExecutable, ContractExecutableRef};
```
leighmcculloch: Can we move ContractExecutable into lib.rs instead, and reexport it from the auth mod for backwards compatibility? It will I think make more sense in the docs if it's rooted and not a reexport in the root.

leighmcculloch (approved):
One ask to move ContractExecutable into lib.rs instead of aliasing it, and then alias it the other way for backwards compatibility, otherwise lgtm.

soroban-sdk/src/auth.rs
mootz12: Not that I am aware of, this type is more for determining what type of executable an Address is, and terminates refs through to their Wasm if they exist.

Due to that, not sure if it's possible to resolve without doing this lookup.

leighmcculloch (approved):

PR #2023, opened 2026-08-22
Reject empty delegate list in account example

leighmcculloch:
### What

Make the modular custom account example reject an authorization entry that attaches no delegated signers.

### Why

The example delegated authentication entirely and had no lower bound on the delegate list, so an entry with `delegates` empty skipped both loops and fell through to `Ok(())` — authorizing the account with no authentication at all. The standalone `modular_account` example in soroban-examples already guards this.

### Known limitations

N/A

### SemVer Change


leighmcculloch:
This example was introduced in v27.0.0. Because it's a pretty bad ommission in the example, this change once approved and merged should be backported to a patch release v27.0.x.

mootz12 (approved):

PR #2024, opened 2026-08-24
Add an account arbitrary address prototype

### LINKED ISSUE #2003: Change Address arbitrary implementation to support accounts (G) too
> [!NOTE]
> Requested by @marcelosalloum in https://github.com/stellar/rs-soroban-sdk/issues/1930 and separated into its own issue here.

The arbitrary implementation of Address only generates contract addresses, not accounts.

Broaden ArbitraryAddress coverage to also generate ScAddress::Account variants, so generated inputs exercise both address kinds.

-

leighmcculloch:
### What

Add `ArbitraryAddressAccount`, a prototype of `Address` that generates account (`G...`) addresses, alongside `ArbitraryAddressContract`, which generates contract (`C...`) addresses and is the default that `ArbitraryAddress` now aliases; the `Vec`, `Map`, and `Option` prototypes accept either.

### Why

There was no way to fuzz a contract with an account address — `ArbitraryAddress` only ever built contract addresses, so a fuzz test or property test taking an `Address` never exercised the account branch of a contract's logic. This gives people options, and if they need to generate both they can do so by creating their own enum to wrap them in. The default continues contracts.

Close #2003

### Known limitations

Using `ArbitraryAddressAccount` in anyway that requires the actual ledger entry to exist requires non-trivial test setup that the soroban-sdk doesn't provide any helpful utilities for. If use picks up this is something that can be explored separately.

### SemVer Change


leighmcculloch:
> I debated making `ArbitraryAddress` contracts still and adding another arbitrary type like `ArbitraryAddressAccountOrContract`, but it seems pretty unintuitive to go that route when the `ArbitraryMuxedAddress` just generates all possible values. I'm open to feedback though. 

@mootz12 @marcelosalloum keen to get your experience. I'm on the fence keeping `ArbitraryAddress` as it is, contract only, and adding a new arbitrary type for other, mostly to avoid surprises with SAC and accounts. But I prefer the consistency of having ArbitraryAddress support all address types.

soroban-sdk/src/_migrating/v28_arbitrary_address.rs
mootz12, line 1: Looks good overall.

A few thoughts:
* We should note that some env testutils like `env::mock_auths` won't work, as they assume the address is a contract (this calls `env::register_at`)
* We should add rustdocs here for `Address::generate` that clarify it only generates contract addresses

If the testutils note makes this too breaking, we could also name the new generator `ArbitraryAddressWithAccounts`.

mootz12 (approved):

soroban-sdk/src/testutils/arbitrary.rs
```diff
@@ -1627,6 +1729,179 @@ mod tests {
...
+
+    fn add_ledger_entry(env: &Env, key: xdr::LedgerKey, data: xdr::LedgerEntryData) {
+        let entry = xdr::LedgerEntry {
+            data,
+            last_modified_ledger_seq: 0,
+            ext: xdr::LedgerEntryExt::V0,
+        };
+        env.host()
+            .add_ledger_entry(&Rc::new(key), &Rc::new(entry), None)
+            .unwrap();
```
mootz12, line 1791: Does the SDK provide a way users can setup an account? This is private, so interacting with G address and stellar asset contracts will be unreachable still.

leighmcculloch: The .host() is accessible actually, just generally discouraged. And yeah there's no public utilities for doing this, you just have to construct the entries yourself.

This is in a test to just prove what would be required, but isn't in the docs.

soroban-sdk/src/_migrating/v28_arbitrary_address.rs
leighmcculloch: DMd offline and resolved to reduce this change to the non-breaking portions, and we'll defer the breaking change to see how much use the individual components get.

Still adding `ArbitraryAddressContract` and `ArbitraryAddressAccount`, and `ArbitraryAddress` will be type aliased to `ArbitraryAddressContract`.

If the breaking change gets introduced later it is just a matter of replacing `ArbitraryAddress` with an enum supporting all address types.

PR #2027, opened 2026-08-24
Make Map::as_object and Map::to_object public

leighmcculloch:
### What

Change `Map::as_object` and `Map::to_object` from `pub(crate)` to `pub`.

### Why

Just a minor tidy up for API consistency, as I noticed that `Map` was the only object-backed type in the SDK keeping these accessors pub(crate) where as other types expose both as `pub`.

### SemVer Change

### DIFF
--- a/soroban-sdk/src/map.rs
+++ b/soroban-sdk/src/map.rs
@@ -308,7 +308,12 @@ impl<K, V> Map<K, V> {
     }
 
     #[inline(always)]
-    pub(crate) fn to_object(&self) -> MapObject {
+    pub fn as_object(&self) -> &MapObject {
+        &self.obj
+    }
+
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


dmkozh (approved):
I think exposing those was a mistake (or at least unintentional), but I guess at this point we can at least make everything consistent.

leighmcculloch:
> I think exposing those was a mistake (or at least unintentional), but I guess at this point we can at least make everything consistent.

This was my thinking too. I almost opened a PR making them all pub crate. But it would be an unnecessary breakage. I might open a PR doc hiding them to keep the API thinner, but just wanted to fix the inconsistency because I think this is the third time I've gone down the rabbit hole of "why is this one pub crate and the others aren't".

leighmcculloch:
Opened:
- https://github.com/stellar/rs-soroban-sdk/pull/2029

PR #2036, opened 2026-09-03
Bump gh-aw setup action to v0.87.10

leighmcculloch:
### What

Bump `github/gh-aw-actions/setup` from v0.83.4 to v0.87.10 in the semver workflow and regenerate `semver.lock.yml` with the matching compiler.

### Why

Keeping current with the latest release. v0.85.4 and later are also the patched releases for [GHSA-9p8f-vm49-2fh4](https://github.com/github/gh-aw-actions/security/advisories/GHSA-9p8f-vm49-2fh4), which this workflow is not affected by: the advisory requires `GH_AW_CI_TRIGGER_TOKEN` together with the `create-pull-request` or `push-to-pull-request-branch` safe outputs, and this workflow uses neither.

### SemVer Change


.gitattributes
leighmcculloch: Second CI note (also not about this file — the diff has only two files to anchor to).

`test-example (token, latest)` is also red, in the **Test with Soroban Examples** workflow. Separate root cause from the OpenZeppelin failures, and likewise not this PR's.

`soroban-examples`' `soroban-token-contract` fails to compile with rustc `E0560` — it initialises a struct field that no longer exists, with the compiler noting `available fields are: live_until_ledger`. So the example is pinned to an older SDK shape.

This is red on `main` too, and the strongest possible confirmation: [run 33729045122](https://github.com/stellar/rs-soroban-sdk/actions/runs/33729045122) is at commit `d1000064`, which is *exactly this PR's base commit*, and it fails identically. The workflow has been failing on `main` continuously since at least 2026-08-26.

Like the OpenZeppelin workflow, this one compiles an external repository (`stellar/soroban-examples`) against this SDK as a downstream-breakage canary, so there is nothing to port here — the fix belongs in `stellar/soroban-examples`.

Net: all red checks on this PR come from the two external-repo canaries, both already broken on the base commit. Nothing in this diff is implicated.

---

dmkozh (approved):

PR #2043, opened 2026-09-05
Shake specs by reachability and markers

leighmcculloch:
> [!NOTE]
>
> Discussion is here:
>
> A first group of PRs deliver const-encoded contract specs, so that contract specs are produced at compile time instead of at proc-macro execution time. This provides the foundation for the capability to construct the specs from information that is not known at proc-macro execution and only known at compile time, like the fully qualified name of a type:
>
> A second group of PRs deliver fully qualified type names in contract specs. Instead of a type having the name `Context` it will have the name `soroban_sdk::auth::Context`. Qualified type names make it possible to uniquely identify types in the spec, even when they have the same name. This resolves several problems with contract specs the type identify problem (https://github.com/stellar/rs-soroban-sdk/issues/1570), type aliases limitations (https://github.com/stellar/rs-soroban-sdk/issues/1857 https://github.com/stellar/rs-soroban-sdk/issues/1063), and optimise spec shaking data section size (https://github.com/stellar/rs-soroban-sdk/issues/1978):
> 1. stellar/stellar-xdr#312
> 1. stellar/stellar-cli#2674
>
> A third group of PRs use those qualified names to shake specs by reachability, so that the markers a contract carries shrink to the few entries a spec never names:
> 1. stellar/rs-soroban-sdk#2043 ← this PR
> 1. stellar/stellar-cli#2720

### What

Stop emitting a spec shaking marker for every user-defined type, and keep spec entries by following the references between them instead, leaving markers only on the events a contract publishes and the errors it panics with.

### Why

A marker per type stood in for a reference graph the spec could not express while a type was named by a bare name, and now that a qualified name identifies exactly one definition, reachability answers which entries are used more precisely than a marker did, since it also drops a type that is marked but that nothing names. The test contract covering shaking end to end shows what that costs a contract today: the entries surviving the shake are unchanged while its markers fall from 49 to 8, its 6 published events and the 2 errors it only ever panics with.

### Negotiating the rules with the build system

Two separate questions have to be answered, and they are answered in two different places.

**Can this build system apply the rules this contract needs?** That is a build-time question, so the env var answers it, and this change moves it to `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V3`. It cannot keep the v2 name: a build system announcing v2 shakes by markers alone, and on a v3 contract, where types deliberately carry no marker, that strips every type instead of keeping the used ones. Trusting a v2 announcement would ship a spec that is wrong, not merely unshaken, so the SDK refuses to build and asks for a stellar-cli carrying stellar/stellar-cli#2720.

**Which rules is a given contract shaken by?** That is a read-time question about a wasm that may have been built long ago, so the contract answers it itself, with the `rssdk_spec_shaking` meta entry it records — moved here from `2` to `3`. `soroban_spec::shaking::filter` takes that version and applies the matching rules: `V2` keeps an entry only if it carries a marker, `V3` also follows the references from functions, events, and types, and a contract recording no version at all is left unshaken. A build system therefore announces every rule set it can apply and lets each contract select among them.

### SemVer Change


soroban-sdk-macros/src/derive_event.rs
leighmcculloch: CI note on head `3064f93` — every red check here is inherited from the base branch, none of it is this PR's.

**The Rust workflow is green:** `build` (msrv + latest), `test` (msrv + latest), `fmt`, `docs`, `readme`, `expand-test-wasms` (both no-diff checks), `cargo-deny`, `migration-docs`, `check-git-rev-deps`.

**`build-fuzz` / `fuzz-corpus`** fail compiling `soroban-sdk-macros`:

```
error[E0308]: mismatched types
  --> soroban-sdk-macros/src/derive_event.rs:206:15
206 |         name: event_name,
    |               ^^^^^^^^^^ expected `ScSymbol`, found `StringM<1024>`
```

`tests/fuzz/fuzz` is a separate workspace and pins `stellar-xdr` to rev `e73df1f6` ("Pending stellar/rs-stellar-xdr#562"), where `ScSpecEventV0.name` is still `ScSymbol`. The main workspace pins `82677e1d`, where it is `StringM<1024>`. The offending line comes from `50ee0c1` ("name user-defined types and events by qualified path") on `fully-qualified-udt-names`, and `git diff 788b536..HEAD -- tests/fuzz` is empty — this PR's only change to this file removes one argument from a `generate_marker_impl` call, ~80 lines away.

**`test-crate` (~96) and `test-example` (~40)** all fail at resolution, before any build:

```
failed to select a version for `stellar-xdr` which could resolve this conflict
```

Same root cause: the external checkouts resolve `stellar-xdr` against the interim rev rather than the one this stack needs.

**Proposed fix, on the base PR rather than here:** bump the `tests/fuzz/fuzz/Cargo.toml` pin from `e73df1f6` to `82677e1d` (and refresh its lockfile) so the fuzz workspace matches the main one. The `test-crate`/`test-example` matrices clear when stellar/rs-stellar-xdr#566 releases. I have not ported either into this PR, since fixing them here would mean carrying a change to the base branch's own pin and conflicting when #1970 fixes it itself — but say the word and I will.


---

leighmcculloch: CI note on head `e22cfbb` — every red check here is inherited from the base branch, none of it is this PR's. The failure set is unchanged from the previous head.

**The Rust workflow is green** apart from the two fuzz jobs below: `build` (msrv + latest), `test` (msrv + latest), `fmt`, `docs`, `readme`, `expand-test-wasms` (both no-diff checks), `cargo-deny`, `migration-docs`, `check-git-rev-deps` — 14 of 17 jobs, the third failure being `complete`, which only aggregates the other two.

**`build-fuzz` / `fuzz-corpus`** fail compiling `soroban-sdk-macros`:

```
error[E0308]: mismatched types
  --> soroban-sdk-macros/src/derive_event.rs:206:15
206 |         name: event_name,
    |               ^^^^^^^^^^ expected `ScSymbol`, found `StringM<1024>`
```

`tests/fuzz/fuzz` is a separate workspace and pins `stellar-xdr` to rev `e73df1f6` ("Pending stellar/rs-stellar-xdr#562"), where `ScSpecEventV0.name` is still `ScSymbol`. The main workspace pins `82677e1d`, where it is `StringM<1024>`. The offending line comes from `50ee0c1` ("name user-defined types and events by qualified path") on `fully-qualified-udt-names`, and `git diff 788b536..HEAD -- tests/fuzz` is empty — this PR's only change to this file removes one argument from a `generate_marker_impl` call, ~80 lines away.

**`test-crate` (~96) and `test-example` (~40)** all fail at resolution, before any build:

```
failed to select a version for `stellar-xdr` which could resolve this conflict
```

Same root cause: the external checkouts resolve `stellar-xdr` against the interim rev rather than the one this stack needs.

**Proposed fix, on the base PR rather than here:** bump the `tests/fuzz/fuzz/Cargo.toml` pin from `e73df1f6` to `82677e1d` (and refresh its lockfile) so the fuzz workspace matches the main one. The `test-crate`/`test-example` matrices clear when stellar/rs-stellar-xdr#566 releases. I have not ported either into this PR, since fixing them here would mean carrying a change to the base branch's own pin and conflicting when #1970 fixes it itself — but say the word and I will.

Separately, on the v3 handshake this head restores: I confirmed the gate bites, by building `tests/spec_shaking` with `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2=1` and V3 unset, which fails with `soroban-sdk requires stellar-cli v28.1.0+`. The full suite passes with both set (61 test binaries, 0 failures), and `expand-test-wasms` being green confirms the workflows set V3 everywhere a contract is built.


---

PR #2049, opened 2026-09-12
fix(bytes): safely convert N to u32 in TryFrom<Bytes> for BytesN<N> (closes #2048)

### LINKED ISSUE #2048: Undersized Bytes converted into BytesN<N> with oversized N values can violate their fixed-length invariant

`soroban-sdk` v28.0.0-rc.1 on the current `main` branch.

### What did you do?

On a 64-bit native host, create an empty (or any sized) `Bytes` and convert it to `BytesN<N>` where `N` is greater than `u32::MAX`:

```rust
let env = Env::default();
let bytes = Bytes::new(&env);
const N: usize = (u32::MAX as usize) + 1;
let value: BytesN<N> = bytes.try_into().unwrap();
assert_eq!(value.len(), 0);
assert!(!value.is_empty());
```

The `try_from` to fail because the length of the bytes being converted is not equal to `N`.

### What did you see instead?

The conversion succeeds because `N` is truncated to `u32` and the truncated N-length matches the length of the bytes being converted.

### Impact

This has no impact on contracts because this only comes into play as a problem when `usize` is sized greater than 32-bits and on `wasm32v1-none` target the size of usize is 32-bits. This bug can only be observed when building for tests on platforms where usize is greater than 32-bits.

rushikeshgarad2024-dev:
### Summary of Fix

Resolves **#2048**: When converting `Bytes` to `BytesN<N>` on 64-bit platforms where `N > u32::MAX`, using `N as u32` truncated `N` to a 32-bit integer, which could match an undersized `Bytes::len()` (e.g., `(u32::MAX as usize) + 1` truncates to `0`, matching an empty `Bytes`).

#### Changes
1. Replaced `if bin.len() == { N as u32 }` with `u32::try_from(N)`:
   ```rust
   let Ok(n) = u32::try_from(N) else {
       return Err(ConversionError {});
   };
   if bin.len() == n {
       Ok(Self(bin))
   } else {
       Err(ConversionError {})
   }
   ```
2. Added a 64-bit target unit test verifying that `BytesN<(u32::MAX as usize) + 1>` conversions fail as expected.

Closes #2048.

### DIFF
--- a/soroban-sdk/src/bytes.rs
+++ b/soroban-sdk/src/bytes.rs
@@ -1090,7 +1090,10 @@ impl<const N: usize> TryFrom<Bytes> for BytesN<N> {
 
     #[inline(always)]
     fn try_from(bin: Bytes) -> Result<Self, Self::Error> {
-        if bin.len() == { N as u32 } {
+        let Ok(n) = u32::try_from(N) else {
+            return Err(ConversionError {});
+        };
+        if bin.len() == n {
             Ok(Self(bin))
         } else {
             Err(ConversionError {})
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Hi @rushikeshgarad2024-dev, thanks for opening the fix, however I was already working on this issue and am resolving it a slightly different way that reduces the runtime code introduced into contracts. See:
- https://github.com/stellar/rs-soroban-sdk/pull/2050

PR #2056, opened 2026-09-15
Fix contractimport! doc example and generated-client name in docs

### LINKED ISSUE #2060: contractimport doc examples are `ignore` and are no longer correct

`soroban-sdk` 28.0.0-rc.2, on `main`.

### What did you do?

Tried to follow the two examples in the rustdoc for `soroban_sdk::contractimport` (`soroban-sdk/src/lib.rs`).

Examples that compile and run.

### What did you see instead?

1. The example doesn't compile as is:
  a. The generated client is named `Client` not `ContractClient`.
  b. The struct for `ContractB` is missing `#[contract]`.
  c. The fn `BytesN` parameter should be an `Address`.
  d. The contract registration uses the deprecated `register_contract_wasm` fn.

2. It also appears the examples are being `ignore`d so that has allowed them to drift over time.

3. The doc block says the `file` path is "relative to the workspace root", but `contractimport!` resolves it relative to the crate's manifest directory.

### Thanks

Some of these issues were reported by @midasbal in https://github.com/stellar/rs-soroban-sdk/pull/2056.

midasbal:
### What

Turn the `contractimport!` rustdoc examples into compiled doctests against the existing `doctest_fixtures/contract.wasm` fixture, correcting the generated client name to `Client` and other compile errors.

### Why

Both examples were marked `ignore`, so nothing caught that they referenced a nonexistent wasm file, and the compile errors.

Close #2060 

### SemVer Change


midasbal:
Added #[contract] on ContractB, needed for ContractBClient and for register to work, matching tests/import_contract.

leighmcculloch (commented):
Hi @midasbal,

Thanks for reporting the issue with the example. We use issues to report bugs, and then will consider the possible fixes. I've opened an issue for this problem here:
- https://github.com/stellar/rs-soroban-sdk/issues/2060

I believe we'd want to fix this by fixing the root cause, that the tests are marked with `ignore` which means there is nothing stopping them from drifting over time.

leighmcculloch (approved):

leighmcculloch:
I've pushed changes to fix the root cause of how we got here.

PR #2062, opened 2026-09-16
Auto weekly kick off release process for minor/patch versions

leighmcculloch:
### What

Add an agentic workflow that runs every Wednesday at noon San Francisco time, classifies every commit on main since the last release by the `semver:` label on the pull request it came from, and dispatches the Bump Version workflow with the version that follows.

### Why

So that new patch/minor releases go out predictably on a schedule without requiring a maintainer to do the assessment and schedule it part of the workflow. Instead a maintainer can respond to the new PR and either approve it and complete the release, or intercept it and reject it.

### SemVer Change


.github/workflows/next-version.md
```diff
@@ -0,0 +1,156 @@
...
+
+      # The most recent release from main is the most recently published
+      # non-draft release whose tag is an ancestor of main. The ancestry check
+      # is what excludes releases cut from branches other than main.
+      gh api "repos/{owner}/{repo}/releases?per_page=100" \
+        --jq '[.[] | select(.draft | not)] | sort_by(.published_at) | reverse | .[].tag_name' \
+        > .release/releases.txt
+      last_tag=
+      while read -r tag; do
+        if git merge-base --is-ancestor "refs/tags/$tag" origin/main 2>/dev/null; then
```
leighmcculloch: The "last release from main" is found by walking releases newest-first and taking the first one whose tag is an ancestor of `origin/main`, but this repo's normal release process cuts releases from `release/vX.Y.Z` branches whose version-bump/fuzz-lock commits are never merged back to main, so most real release tags fail this ancestor check.

Verified against the live repo: `git merge-base --is-ancestor refs/tags/v27.0.6 origin/main` is false (v27.0.6 was cut on branch `release/v27.0.6`, still unmerged into main), as is v27.0.4 and v27.0.3, while the older v27.0.5 passes. Given a release history like this, the loop skips the real last release (v27.0.6) and wrongly settles on v27.0.5, so the workflow re-classifies commits already shipped in v27.0.6 and can propose/dispatch a bump for a version range that's already released. Today's run happens to work only because the current newest release (v28.0.0-rc.2) is exceptionally an ancestor of main; the very next ordinary patch/minor release cut on a `release/*` branch will reproduce the mismatch.

---

.scripts/commits-pull-requests
```diff
@@ -0,0 +1,66 @@
...
+    if not prs:
+        return "", "", ""
+
+    pr = prs[0]
+    labels = [l["name"] for l in pr["labels"] if l["name"].startswith(LABEL_PREFIX)]
+    return str(pr["number"]), pr["html_url"], " ".join(labels)
+
+def run(args):
+    """Run a command and return its stdout, exiting if it fails."""
+    return subprocess.run(args, capture_output=True, text=True, check=True).stdout
```
leighmcculloch: `run()` calls `subprocess.run(..., check=True)` with no retry/backoff, and `pull_request()` issues one serial `gh api` call per commit in the release window, so a single transient GitHub API failure (secondary rate limit, 5xx, network blip) anywhere in the loop raises an uncaught `CalledProcessError` and kills the whole pre-agent step.

A week with a normal batch of commits since the last release triggers N sequential `gh api repos/{owner}/{repo}/commits/{sha}/pulls` calls; if GitHub throttles or hiccups on any single call, the script exits non-zero, `set -euo pipefail` aborts the pre-agent-steps block, and the scheduled Wednesday run fails outright with no commits/versions data collected — defeating the PR's stated goal of releases going out predictably on a schedule without requiring a maintainer to notice and re-run it by hand.

---

.github/workflows/next-version.md
```diff
@@ -0,0 +1,156 @@
...
+      GH_TOKEN: ${{ github.token }}
+    run: |
+      set -euo pipefail
+      mkdir -p .release
+      git fetch --quiet --force --tags origin +refs/heads/main:refs/remotes/origin/main
+
+      # The most recent release from main is the most recently published
+      # non-draft release whose tag is an ancestor of main. The ancestry check
+      # is what excludes releases cut from branches other than main.
+      gh api "repos/{owner}/{repo}/releases?per_page=100" \
```
leighmcculloch: The release listing used to find the last release is capped at `per_page=100` with no pagination, so once a repo accumulates >100 releases, any recent release older than that window (compounded by the ancestor-check bug flagged above, which skips entries) is invisible to the search and `last_tag` can end up empty even though releases exist.

If the 100 most recent releases all fail the ancestor-of-main check (plausible given the release-branch workflow described above) and the true matching ancestor tag lies beyond the 100th most recent release, `last_tag` stays empty, `.release/last-release-tag.txt` is blank, and per the agent instructions the workflow reports "no release found" and dispatches nothing, even though main clearly has releases behind it.

---

Issue #2064, opened 2026-09-17
Consider documenting or eliminating the cases that TryFromVal panics instead of erroring on some vals


leighmcculloch:
### What version are you using?

soroban-sdk 28.0.0-rc.2, on the current main branch, but this has been a long existing behaviour in all versions of the soroban-sdk, and was originally an intentional design choice to fail in cases where conversions failed host side or during host functions.

### What did you do?

Converted a `Val` into each of the SDK types and each kind of contract type, for a `Val` holding a value of every type the host supports along with contract generated types.

### What did you expect to see?

I expected to see some conversions to return errors while others would panic, because this was a design decision early on with Soroban to fail host-side when the host could see a conversion would fail. The decision was based on the basis that contracts are likely to just fail in this case as well.

But the fact that `TryFromVal` panics in some cases appears to be surprising.

At the very least the cases that panic must be well documented, which they aren't today.

But possibly we should consider closing the behaviour difference so all `TryFromVal`s error instead of panic.

### What did you see instead?

The vast majority of type conversions fail with an error that the contract can handle.

However a few cases with generated contract types fail host side:

- Struct from a map with non-`Symbol` keys — panics with `Value, UnexpectedType` in `map_unpack_to_slice`. 
- Tuple struct from a vec of the wrong length — `Object, UnexpectedSize` in `vec_unpack_to_slice`.
- Enum from a vec whose first element is a symbol that isn't a variant name — `Value, InvalidInput`, "symbol not found in slice of strs".

### Possible approaches to explore

1. Change nothing, add docs so the cases are clearly documented
2. Change logic to return an error, requires a protocol change. (Unlikely)

leighmcculloch:
Capturing existing behaviour to make sure we clearly understand which cases error and panic from an SDK user perspective:
- https://github.com/stellar/rs-soroban-sdk/pull/2065

PR #2065, opened 2026-09-17
Add tests for conversions from vals of other types

leighmcculloch:
### What

Add tests converting a `Val` into every SDK type and every kind of contract type.

### Why

There's always been a variety of how conversion errors are handled, with some erroring and some panicing host side. This test is being added to capture the status quo as a first step in getting really clear on what documentation is missing regarding #2064.

### SemVer Change


soroban-sdk/src/tests/try_from_val_type_mismatch.rs
```diff
@@ -0,0 +1,455 @@
...
+//! callers that are free to send a value of any type.
+//!
+//! Converting is an error in every case except three, which trap in the host
+//! and so panic: a contract type struct converting from a map that has keys
+//! that are not symbols, a contract type tuple struct converting from a vec of
+//! a different length, and a contract type enum converting from a vec whose
+//! first element is a symbol that is not one of the variant names.
+//!
+//! Each line that converts is marked with the result it has, so that the
+//! results are visible when scanning the tests: 🟢 converts, 🟡 returns a
```
dmkozh: I'm not a fan of using colors in code, not to mention that this is not colorblind-friendly.
Moreover, I have no idea what green+yellow combination is supposed to mean. Can we use some iconographic emojis or just short words instead?

leighmcculloch: Updated and simply removed them.

dmkozh (approved):

PR #2057, opened 2026-09-15
Add migration docs for check auth contract executable changes

mootz12:
### What

Adds migration docs for the `__check_auth` impact due to the new `ExecutableRef` variant in `ContractExecutable`.

### Why

To inform contract developers of the impact on contract accounts in p28.

### Known limitations

None

### SemVer Change

- [ ] Major (`vX._._`) - Breaking change to the public API.
- [ ] Minor (`v_.Y._`) - Additive change to the public API.
- [x] Patch (`v_._.Z`) - No change to the public API.

PR #2071, opened 2026-09-18
Add soroban-lld to shake contract spec at link time

leighmcculloch:
> [!NOTE]
>
> This PR is an experiment and is not targeting being merged, and is unlikely to be merged.

### What

Add `soroban-lld`, a linker shim that performs the post-build actions that the stellar-cli performs today, removing spec entries for unused types, deduping specs and meta entries, etc.

### Why

Today the post-build steps occur in the stellar-cli. This requires the stellar-cli to be used when building, which means IDEs that may build using cargo or other tooling that would trigger a cargo build will error. It also means Rust devs who are already familiar with Rust need to use a tool unfamiliar to them.

The soroban-sdk historically has been designed to fit into standard rust build workflows. Sometime ago we moved away from that first with using it to trigger `cargo rustc` instead of `cargo build`. More recently it became required to do spec shaking and spec and meta deduping.

This change would still require installing a tool, such as `soroban-lld` and a `.cargo/config.toml` entry. And realistically it couldn't fully replace `stellar contract build`, which also triggers rustc instead of build to ensure the correct crate-type is built.

### SemVer Change


soroban-lld/Cargo.toml
leighmcculloch: CI note: the red checks on this PR are not from this change.

`test-example (token, latest)` and the `test-crate (examples/multisig-smart-account/*, packages/accounts, ...)` jobs come from the **Test with soroban-examples** and **Test with OpenZeppelin Contracts** workflows, which check out external repos and build them against this SDK.

Both are red on `main` too, including on this PR's base commit `48d5067`, and have been failing for several days (last six runs of each on `main`: all `failure`).

The error is a non-exhaustive match (E0004) on `soroban_sdk::auth::Context` in the external `stellar-accounts` crate, missing arms for the p28 `CreateContractHostFn` / `CreateContractWithCtorHostFn` variants carrying `ContractExecutable::ExternalRef` — the migration documented in #2057.

This PR adds the `soroban-lld` crate, adds it to workspace members, and edits a build-script error string, so it cannot affect `auth::Context`. There is no fix in this repo to port: the fixes belong in `OpenZeppelin/stellar-contracts` and `stellar/soroban-examples`. I have not spent a re-run, since the base-branch evidence is conclusive.

This repo's own **Rust** workflow is the check that reflects this change; it was still in progress when this was written.

---

PR #2074, opened 2026-09-20
Fix event data_format in contractimport! output

### LINKED ISSUE #2073: `contractimport!` does not emit an event's `data_format`, making imported event types inconsistent

soroban-sdk 28.0.0, but bug present since 23.0.0.

### What did you do?

Defines some events like the below, build the contract into wasm, and imported the wasm into another contract using the `contractimport!` macro.

```rust
pub struct SingleValue {
    #[topic]
    from: Address,
    amount: i128,
}

pub struct SingleValueVoid {
    #[topic]
    from: Address,
}

#[contractevent(data_format = "vec")]
pub struct VecValues {
    #[topic]
    from: Address,
    a: u32,
    b: u32,
}
```

The `contractimport!` macro to generate events identical to their original source as above.

### What did you see instead?

`contractimport!` generates them as follows, missing the `data_format` parameter, meaning that they all encode as the default data format, which is `"map"`.

```rust
#[soroban_sdk::contractevent(topics = ["single_value"])]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct SingleValue {
    #[topic]
    pub from: soroban_sdk::Address,
    pub amount: i128,
}

#[soroban_sdk::contractevent(topics = ["single_value_void"])]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct SingleValueVoid {
    #[topic]
    pub from: soroban_sdk::Address,
}

#[soroban_sdk::contractevent(topics = ["vec_values"])]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct VecValues {
    #[topic]
    pub from: soroban_sdk::Address,
    pub a: u32,
    pub b: u32,
}
```

leighmcculloch:
### What

Emit an event's `data_format` in the code `contractimport!` generates.

### Why

The code generator did not inject the `data_format` into generated contract event types, so every imported event fell back to the attribute's `map` default and events declared `single-value` or `vec` encoded when reused by other contracts.

Close #2073

### SemVer Change


soroban-spec-rust/src/types.rs
mootz12, line 1: It looks like we also don't pass `sparse` along either. This isn't on the XDR so it would be a bigger lift.

Thoughts on if it's needed? Given it uses a different host function it feels like it should be, or at least documented as not supported via Wasm imports.

mootz12 (approved):
LG2M. One somewhat related comment.

soroban-spec-rust/src/types.rs
leighmcculloch: I believe this was discussed here: https://github.com/stellar/stellar-protocol/discussions/1877#discussioncomment-17651313, and the outcome was that it won't be included and downstream should assume for any event anywhere it could be sparse. If there are cases where sparseness creates ambiguity it'll need to be communicated some other way that is undefined.

PR #2078, opened 2026-09-21
Snapshot contract spec sections as XDR-JSON

leighmcculloch:
> [!NOTE]
>
> Discussion is here:
>
> 1. stellar/rs-soroban-sdk#2078 ← this PR
>
> 1. stellar/stellar-xdr#312
> 1. stellar/stellar-cli#2674
>
>
> 1. stellar/stellar-cli#2720

### What

Record the `contractspecv0` section of every test contract wasm in `tests-specs/`, as XDR-JSON.

### Why

To identify and track changes to contract specifications generated by the test vectors.

Today the `tests-expanded/` directory tracks the code that macros generate which has helped us understand the impact of macro changes.

Upcoming are changes to how contract specifications get built, and snapshots of the generated specs will give us a way to confirm that the refactor doesn't change the generated specs at all.

The specs are written as XDR-JSON so that changes to the specs are identifiable. That is a tradeoff. Under some conditions it would be better to store the raw binary, such as if the JSON format itself changes, but being able to inspect by eye in diffs changes to the specs will be more valuable.

### SemVer Change


sisuresh:
@leighmcculloch I'm not too familiar with this code, but here were some claude findings. Please let me know if this isn't helpful - 

- Swallowed errors: the Makefile loop at Makefile:92-96 ignores spec-json failures, so make exits 0 and writes an empty snapshot. Adding || exit 1 fixes it.
- Stale snapshots: it globs test_*.wasm in target/ instead of looping over $(TEST_CRATES), so old wasm files get snapshotted when regenerating locally.
- Wording: it only reads the first contractspecv0 section and captures the spec before shaking. The Makefile comment claims more than that.

Makefile
```diff
@@ -80,6 +80,23 @@ expand-tests: build-test-wasms
...
+# directory, as built by build-test-wasms and so before any spec shaking, as a
+# stream of XDR-JSON values. Serves to surface changes to the spec the SDK
+# embeds, which the expanded code does not show because the spec is encoded
+# from it rather than written out by it.
+spec-snapshots: build-test-wasms spec-snapshots-from-built-wasms
+
+spec-snapshots-from-built-wasms:
+	rm -fr tests-specs
+	mkdir -p tests-specs
+	for name in $(TEST_CRATES); do \
```
leighmcculloch: @sisuresh thanks, all three were right. Fixed in b975df24:
- The loop now goes over `$(TEST_CRATES)` and skips crates with no wasm, so stale files in `target/` are no longer snapshotted.
- A `spec-json` failure now stops make, with `|| exit 1`.
- The comment now says it dumps the `contractspecv0` section of the wasms as `build-test-wasms` builds them, before any spec shaking.

---

Makefile
```diff
@@ -80,6 +80,24 @@ expand-tests: build-test-wasms
 			cargo expand --package $$package --release --target wasm32v1-none | rustfmt > tests-expanded/$${package}_wasm32v1-none.rs; \
 	done
 
+# Dumps the contractspecv0 section of each test vector contract in the tests/
+# directory, as built by build-test-wasms and so before any spec shaking, as a
+# stream of XDR-JSON values. Serves to surface changes to the spec the SDK
+# embeds, which the expanded code does not show because the spec is encoded
+# from it rather than written out by it.
+spec-snapshots: build-test-wasms
```
leighmcculloch: Re Copilot's comment on this line: fixed in ea401b47. `spec-snapshots` now depends only on `build-test-wasms` and runs `$(MAKE) spec-snapshots-from-built-wasms` in its recipe, so under `make -j` the snapshots can't be taken before the wasms are built.

---

sisuresh (approved):

.scripts/spec-json/src/main.rs
mootz12: (nit) it might make sense to print this as a pretty printed JSON array. Potentially with some amount of ordering enforced to keep diffs readable.

mootz12 (approved):

.scripts/spec-json/src/main.rs
leighmcculloch: Great ideas. Implemented in:
- https://github.com/stellar/rs-soroban-sdk/pull/2086

PR #2080, opened 2026-09-22
Hash spec markers with a const SHA-256

leighmcculloch:
> [!NOTE]
>
> Discussion is here:
>
> 1. stellar/rs-soroban-sdk#2080 ← this PR
>
> 1. stellar/stellar-xdr#312
> 1. stellar/stellar-cli#2674
>
>
> 1. stellar/stellar-cli#2720

### What

Make the function `soroban_spec::shaking::generate_marker_for_xdr` callable in const contexts and in nostd crates.

> [!CAUTION]
> Replaces the use of the `sha2` crate for hashing the spec entries with an in-repo const-compatible sha256 implementation.

### Why

So that the function can be called from within Soroban contracts, in the definition of a static variable. In the stack of PRs above the PR after this shifts to building XDR into a const, and the XDR will no longer be known at proc-macro execution time. The function `generate_marker_for_xdr` is currently called from within the proc-macros, but because the XDR will not be known until const evaluation, the marker generation will also need to move to const evaluation.

### SemVer Change


soroban-spec/Cargo.toml
```diff
@@ -10,12 +10,19 @@ version.workspace = true
 edition = "2021"
 rust-version.workspace = true
 
+[features]
```
leighmcculloch: Nine checks are red here and none of them are this PR's: `test-example (token)` and eight `test-crate` jobs (`multisig-smart-account/account`, `ed25519-verifier`, `factory`, `spending-limit-policy`, `threshold-policy`, `webauthn-verifier`, `weighted-threshold-policy`, and `packages/accounts`). The identical nine are red on #2078 and #2061 below this one in the stack, so they predate this change.

The cause is upstream of this diff: the external openzeppelin crates fail to compile with `error[E0004]: non-exhaustive patterns` on `soroban_sdk::auth::Context::CreateContractHostFn` and `CreateContractWithCtorHostFn`, because `ContractExecutable` gained an `ExternalRef` variant in the env dependency update #2061 brings in, and their matches have not been updated for it. There is no fix to port — it needs a release of those crates or a change in #2061 at the bottom of the stack.

This PR's own jobs are green: the `fuzz-corpus` differential target ran its 47-file corpus with no crash, once 566af140 fixed the stale `tests/fuzz/fuzz/Cargo.lock`.

---

soroban-spec/Cargo.toml
```diff
@@ -10,12 +10,19 @@ version.workspace = true
...
+default = ["std"]
+# The marker functions are usable without std so that soroban-sdk can call
```
leighmcculloch: `expand-test-wasms` is red for a tenth reason that is also not this PR's. Its diff is confined to the `eventscontract` WASM bytes embedded in `tests-expanded/test_import_contract_tests.rs` and `test_import_contract_wasm32v1-none.rs`.

That is the stack not having absorbed `main`. `main` is one commit ahead of #2061 — `2aaed2fa` "Fix event data_format in contractimport! output" (#2074) — which changes `tests/events/src/lib.rs` and regenerates those same four expanded files. CI builds a merge ref that includes `main`, so `test_import_contract` compiles against `main`'s `tests/events` while the checked-in `tests-expanded/` in this stack predates it.

The fix belongs at the bottom of the stack: #2061 takes `main` in, and it flows up from there. Merging `main` into this PR instead would put that absorption in the wrong place and widen the diff, and regenerating `tests-expanded/` here would only hide a signal that is doing its job.

---

soroban-spec/Cargo.toml
```diff
@@ -10,12 +10,19 @@ version.workspace = true
...
+# them from const contexts in contracts built for wasm. Everything that reads
```
leighmcculloch: `spec-snapshots` is red for the same reason as `expand-test-wasms` above, and is likewise not this PR's. Its diff is confined to `tests-specs/test_events.json`, which is missing the `MapValues` and `VecValues` entries — the two events `2aaed2fa` (#2074) adds to `tests/events/src/lib.rs` on `main`. Same fix: #2061 takes `main` in at the bottom of the stack.

---

soroban-spec/src/shaking.rs
```diff
@@ -520,3 +530,60 @@ mod tests {
         assert_eq!(spec_shaking_version_for_meta(&meta), 1);
     }
 }
+
+#[cfg(test)]
+mod sha256_tests {
+    use super::{generate_marker_for_xdr, sha256};
```
leighmcculloch: Re Copilot's comment on this line: the finding doesn't hold, and no change is needed.

`mod sha256;` (line 43) puts the name in the **type** namespace; `use sha256::sha256;` (line 44) puts the function in the **value** namespace. `use super::{generate_marker_for_xdr, sha256};` imports the name in both namespaces, and the call `sha256(&input)` resolves in the value namespace, so it reaches the function. Rust reports an ambiguity only when two items collide within a single namespace, which isn't the case here.

Verified rather than assumed: `cargo test --package soroban-spec sha256_tests` compiles with no ambiguity warning or error and all three tests pass, as does `cargo test --package soroban-spec --no-default-features`.

---

Makefile
```diff
@@ -49,9 +49,14 @@ build-test-wasms: fmt
 
 # Builds the fuzz tests. Requires cargo-fuzz and cargo-afl.
 build-fuzz:
+	cd soroban-spec/fuzz && cargo +nightly fuzz check
 	cd tests/fuzz/fuzz && cargo +nightly fuzz check
 	cd tests/fuzz_afl/fuzz && cargo afl build
 
+fuzz-corpus:
+	cd soroban-spec/fuzz && for t in $$(cargo +nightly fuzz list) ; do cargo +nightly fuzz run $$t -- -runs=0 || exit 1 ; done
+	cd tests/fuzz/fuzz && for t in $$(cargo +nightly fuzz list) ; do cargo +nightly fuzz run $$t -- -runs=0 || exit 1 ; done
```
leighmcculloch: Re the review comment on this line (`-runs=0` is unlimited, never reaches the diff check, may grow the corpus): all three claims are the wrong way round, and the job's own log disproves them.

libFuzzer's `-runs=N` is the number of individual test runs, and the default is `-1` — that is the infinite setting. `0` means zero mutation runs, so libFuzzer reads and executes the seed corpus once and exits. That *is* the dedicated corpus-validation invocation being asked for; it's the standard idiom for replaying a checked-in corpus.

The `fuzz-corpus` job on `8b2aac37` ran green start to finish in four minutes (12:48:01Z → 12:52:05Z, job `106750845326`). Its log for the new target:

```
INFO: seed corpus: files: 44 min: 1b max: 10041b total: 82373b rss: 42Mb
#45	INITED cov: 59 ft: 149 corp: 40/69Kb exec/s: 0 rss: 45Mb
#45	DONE   cov: 59 ft: 149 corp: 40/69Kb lim: 10041 exec/s: 0 rss: 45Mb
Done 45 runs in 0 second(s)
```

It ran the 44 seed inputs and terminated immediately; the run count tracks the corpus size rather than growing. The `git add -N . && git diff HEAD --exit-code` step then ran and passed — which is what the green conclusion covers — so the target both reaches the diff check and leaves the corpus unmodified.

No change needed.

---

leighmcculloch:
> ****
> * High severity [Use finite fuzz runs for corpus validation](#discussion_r4077688934)
> * Medium severity [Disambiguate sha256 function import from module](#discussion_r4071796756)

Both of these issues that Copilot keeps reporting are incorrect.

soroban-spec/src/shaking/sha256.rs
```diff
@@ -0,0 +1,134 @@
...
+                b += 1;
+            }
+            w[t] = word;
+            t += 1;
+        }
+
+        let mut t = 16;
+        while t < 64 {
+            let s0 = w[t - 15].rotate_right(7) ^ w[t - 15].rotate_right(18) ^ (w[t - 15] >> 3);
+            let s1 = w[t - 2].rotate_right(17) ^ w[t - 2].rotate_right(19) ^ (w[t - 2] >> 10);
```
mootz12: An agent detected that this function is fairly expensive in const time, especially in `1.95`.

It found that the cost sha256 function can only hash a 26kB entry on 1.95 before we would get hit with `long_running_const_eval`.

It recommended replacing all `rotate_right` calls with the following function. Left agents comments for additional context:
```rust
/// Rotate `x` right by `n` bits.
///
/// Spelled out with shifts rather than `u32::rotate_right` because rustc's
/// const evaluator steps through the rotate intrinsic's fallback body, which
/// makes it several times slower and trips the `long_running_const_eval` lint
/// on inputs above a few tens of kilobytes.
const fn rotr(x: u32, n: u32) -> u32 {
    (x >> n) | (x << (32 - n))
}
```

E.g. these lines would become:
```rust
let s0 = rotr(w[t - 15], 7) ^ rotr(w[t - 15], 18) ^ (w[t - 15] >> 3);
let s1 = rotr(w[t - 2], 17) ^ rotr(w[t - 2], 19) ^ (w[t - 2] >> 10);
```

soroban-spec/src/shaking.rs
```diff
@@ -520,3 +530,129 @@ mod tests {
         assert_eq!(spec_shaking_version_for_meta(&meta), 1);
     }
 }
+
+#[cfg(test)]
+mod sha256_tests {
```
mootz12: (nit) not a strong opinion, but I think I prefer these tests within the `sha256` file, excluding `generate_marker_for_xdr`

Makefile
```diff
@@ -51,9 +51,14 @@ build-test-wasms: fmt
...
```
mootz12: I think the corpus folder here is gitignored, so not sure if this does much.

Makefile
```diff
@@ -51,9 +51,14 @@ build-test-wasms: fmt
 
 # Builds the fuzz tests. Requires cargo-fuzz and cargo-afl.
 build-fuzz:
+	cd soroban-spec/fuzz && cargo +nightly fuzz check
 	cd tests/fuzz/fuzz && cargo +nightly fuzz check
 	cd tests/fuzz_afl/fuzz && cargo afl build
 
+fuzz-corpus:
+	cd soroban-spec/fuzz && for t in $$(cargo +nightly fuzz list) ; do cargo +nightly fuzz run $$t -- -runs=0 || exit 1 ; done
```
mootz12:
```suggestion
	cd soroban-spec/fuzz && targets=$$(cargo +nightly fuzz list) && for t in $$targets ; do cargo +nightly fuzz run $$t -- -runs=0 || exit 1 ; done
```

Agent noticed this job will not fail when `cargo +nightly fuzz list` fails, it will just have an empty list. Pulling the call into a variable should cause the failure to propagate.

soroban-spec/src/shaking.rs
```diff
@@ -520,3 +530,129 @@ mod tests {
...
+            let input: Vec<u8> = (0..len).map(|i| (i % 251) as u8).collect();
+            assert_eq!(&sha256(&input), expected, "mismatch at len {len}");
+        }
+    }
+
+    /// The two samples from NIST's SHA-256 example document, a message that
+    /// fits in one block and a message that spans two.
+    /// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/SHA256.pdf
+    #[test]
+    fn sha256_matches_nist_examples() {
```
mootz12: It might make sense to include `SHA256ShortMsg` from the Nist set. Thoughts?

mootz12 (approved):
LG2M. A few comments in-line.

